PR #2, opened 2022-08-11
Pulling in new OpenAPI /api into our new new docs site

tyvdh:


docs/glossary/claimable-balance.mdx
```diff
@@ -72,45 +72,54 @@ const sdk = require("stellar-sdk");
...
-  if (!aAccount) { return }
+    console.error(`Failed to load ${A.publicKey()}: ${err}`);
+  });
+  if (!aAccount) {
+    return;
+  }
 
   // Create a claimable balance with our two above-described conditions.
-  let soon = Math.ceil((Date.now() / 1000) + 60); // .now() is in ms
+  let soon = Math.ceil(Date.now() / 1000 + 60); // .now() is in ms
```
Shaptic, line 92: why do linters think they know better than humans

tyvdh: /shrug I'm not a linter fan but the only thing I hate worse than linting is configuring linters

PR #27, opened 2022-09-01
Updating footer

briwylde08:
Change developers-community and add community links
All tools to link to dev docs tools page 
Remove get started from resources

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -79,7 +79,7 @@ const config = {
           position: "right",
         },
         {
-          href: "https://github.com/stellar/stellar-docs",
+          href: "https://github.com/stellar",
           label: "GitHub",
           position: "right",
         },
@@ -100,16 +100,16 @@ const config = {
           title: "Resources",
           items: [
             {
-              label: "Stellar Learn",
-              href: "https://www.stellar.org/learn/intro-to-stellar",
-            },
-            {
-              label: "Case Studies",
-              href: "https://www.stellar.org/case-studies",
+              label: "Developers Blog",
+              href: "https://www.stellar.org/developers-blog",
             },
             {
               label: "Stellar Quest",
               href: "https://quest.stellar.org/",
+            },
+            {
+              label: "Stellar Community Fund",
+              href: "https://communityfund.stellar.org/"
             },
           ],
         },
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -136,23 +128,24 @@ const config = {
...
+          title: "Community",
           items: [
             {
-              label: "Developer Resources",
-              href: "https://www.stellar.org/developers-legacy",
+              label: "Stellar Developer Discord",
+              href: "https://discord.com/invite/zVYdY3ktTn",
             },
             {
-              label: "Status",
```
tyvdh, line 150: We should definitely keep this link. It's one of the most important for pro tier devs. We should also include this one https://dashboard.stellar.org/

docusaurus.config.js
```diff
@@ -136,23 +128,24 @@ const config = {
...
         },
         {
-          title: "Developers",
+          title: "Community",
           items: [
             {
-              label: "Developer Resources",
-              href: "https://www.stellar.org/developers-legacy",
+              label: "Stellar Developer Discord",
+              href: "https://discord.com/invite/zVYdY3ktTn",
```
tyvdh, line 140: This is a non-expiring invite yes?

docusaurus.config.js
tyvdh: Both of these links likely belong in tools

docusaurus.config.js
```diff
@@ -100,11 +100,7 @@ const config = {
           title: "Resources",
           items: [
             {
-              label: "Get Started",
-              href: "https://www.stellar.org/start",
-            },
-            {
-              label: "Learn",
+              label: "Stellar Learn",
```
tyvdh, line 103: Probably don't need this? We should link to the Stellar homepage somewhere but probably in the header not the footer

tyvdh (changes requested):

docusaurus.config.js
briwylde08: I was just trying to shorten the list so the footer isn't ridiculously long. Just want to keep the top 4 most used then link to the rest of the tools. Most used are? Account Viewer, Dashboard, Ledger Explorer, Status? Or something else?

PR #28, opened 2022-09-01
Incorporating a few suggestions to the new docs.

Shaptic:
This includes a few of my suggestions [from Slack](https://stellarfoundation.slack.com/archives/CGY4VS8SZ/p1662055556165489?thread_ts=1662046167.887789&cid=CGY4VS8SZ), but doesn't include these:

* It's a little off-putting to see 
> The lumen is the only token that doesn’t require an issuer or trustline, and it pays all transaction fees and covers minimum balances on the network.
on the XLM page before introducing what issuer and trustlines are :thinking_face: but I'm not sure how to do that better.

* In the List of Operations, some have errors in a tabular form and some don't. Should those be consistent?
* Can we get MathJax support?

Since these are a little more opinionated and/or involved, so I'll leave those to @briwylde08.

### DIFF
--- a/docs/fundamentals-and-concepts/stellar-data-structures/accounts.mdx
+++ b/docs/fundamentals-and-concepts/stellar-data-structures/accounts.mdx
@@ -44,6 +44,6 @@ Account data is stored in subentries, each of which increases an account’s min
 
 Trustlines are an explicit opt-in for an account to hold and trade a particular asset. To hold a specific asset, an account must establish a trustline with the issuing account using the change_trust operation. Trustlines track the balance of an asset and can also limit the amount of an asset that an account can hold.
 
-A trustline must be established for an account to receive any asset except lumens (XLM). You can create a claimable balance to send assets to an account without a trustline, but the recipient has to create a trustline to claim that balance. You can learn more about claimable balances in their [encyclopedia entry](../../encyclopedia/claimable-balances).
+A trustline must be established for an account to receive any asset except lumens (XLM). You can create a claimable balance to send assets to an account without a trustline, but the recipient has to create a trustline to claim that balance. Learn more here: [Claimable Balances Encyclopedia Entry](../../encyclopedia/claimable-balances)
 
 A trustline also tracks liabilities. Buying liabilities equal the total amount of the asset offered to buy aggregated over all offers owned by an account, and selling liabilities equal the total amount of the asset offered to sell aggregated over all offers owned by an account. A trustline must always have a balance sufficiently large to satisfy its selling liabilities and a balance sufficiently below its limit to accommodate its buying liabilities.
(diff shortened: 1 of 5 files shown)


tyvdh:
Docusaurus supports https://katex.org/ and I've added that to this branch

tyvdh:
# @briwylde08 
- [ ] It's a little off-putting to see
  > The lumen is the only token that doesn’t require an issuer or trustline, and it pays all transaction fees and covers minimum balances on the network.

  on the XLM page before introducing what issuer and trustlines are 🤔 but I'm not sure how to do that better.
  
- [ ] In the List of Operations, some have errors in a tabular form and some don't. Should those be consistent?

- [ ] On the trustlines page ([link](https://developers-pr9.previews.kube001.services.stellar-ops.com/docs/fundamentals-and-concepts/stellar-data-structures/accounts#trustlines)), it states
  > Learn more here: [Claimable Balances Encyclopedia Entry](https://developers-pr9.previews.kube001.services.stellar-ops.com/docs/encyclopedia/claimable-balances)

  but something like "You can learn more from the claimable balances encyclopedia entry." flows better imo. (Btw love the rename of glossary -> encyclopedia.)

- [ ] Quoting "burn" here might make this sentence flow better
  > To delete, or burn, an asset, you must send it back to the account that issued it.

# @tyvdh 
- [x] Can we get MathJax support? Seeing e.g. The largest amount unit possible is ((2^63)-1)/(10^7) is not very eye-pleasing.

tyvdh:
@Shaptic Nice!
Screen Shot 2022-09-01 at 3 13 30 PM

Shaptic:
@tyvdh thinking about this more, I don't see much math besides the bit I found lol. Do you think the JavaScript cost (e.g. the big file to download + execute) on every page (is it every page? or only pages where you use `$ <math> $` in the source?) is worth the handful of instances where we use a bit of math that can be improved w/ HTML (e.g. `<sup>` like I did before) or Unicode characters?

tyvdh:
I mean it'll be cached and progressively loaded so it's not a big lift

Shaptic:
Re (2), converting to OpenAPI will definitely help a lot with consistency, **but** I think adding thorough details like these per-operation error messages is a "phase 2" extension to the spec that will need more work to write and possibly integrate. I just wanted to call it out because we have error and parameter tables for **all** of the operations in the existing list of operation docs. It may be better to keep them as-is, then replace them with links like done here when the API docs are confirmed to have the same info and level of detail?

briwylde08:
1. Makes sense. I think our options here are to link to the sections on trustlines and issuers below this introduction. Or we can move the entire Lumens section below the Stellar Data Structures section. I'm okay either way- @tyvdh any opinions about where Lumens is in the overall F&C section? 

2. Ideally we'd have consistency across all operations, the reason there are some tables and some links for parameters/errors is because, yes, some of the documentation in the API docs was missing. I liked the idea of not having the tables just because it's so much cleaner. 

3. Most of the links in the new docs are not conversational so this would have to be changed in a lot of places. I dunno, I think this is probably a preference thing? My idea is that... putting too many links conversationally throughout the docs makes people click away too much. I want them to get all the way through the paragraph/section and then decide if they want more information. But, again, probably just a preference thing.  

4. Fine by me

PR #40, opened 2022-09-27
Fixing some code examples syntax languages

ElliotFriend:
Found one that needed fixing, and then did a quick search for any others that might be wrong. Here's the three I found. No big impact, but will help to make the docs more consistent.

Using `javascript` rather than `js` rendered "Example" into the code snippet headers. Weird 🤷🏻‍♂️ 

Signed-off-by: Elliot Voris <elliot@voris.me>

### DIFF
--- a/docs/fundamentals-and-concepts/stellar-data-structures/assets.mdx
+++ b/docs/fundamentals-and-concepts/stellar-data-structures/assets.mdx
@@ -34,11 +34,11 @@ In Horizon, assets are represented in a JSON object:
 
 ```json
 {
-  asset_code: "AstroDollar",
-  asset_issuer: "GC2BKLYOOYPDEFJKLKY6FNNRQMGFLVHJKQRGNSSRRGSMPGF32LHCQVGF",
+  "asset_code": "AstroDollar",
+  "asset_issuer": "GC2BKLYOOYPDEFJKLKY6FNNRQMGFLVHJKQRGNSSRRGSMPGF32LHCQVGF",
   // `asset_type` is used to determine how asset data is stored.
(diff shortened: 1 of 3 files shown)


briwylde08:
@tyvdh can you take a look at this when you get a chance?

docs/fundamentals-and-concepts/stellar-data-structures/assets.mdx
```diff
@@ -32,13 +32,13 @@ The public key of the issuing account is linked on the ledger to the asset. Resp
 
 In Horizon, assets are represented in a JSON object:
 
-```json5
```
tyvdh, line 35: Comments don't render with `json` thus `json5` which does support comments

docs/tutorials/follow-received-payments.mdx
```diff
@@ -53,7 +53,7 @@ Create a new file named `make_account.js` and paste the following text into it:
 
 <CodeExample>
 
-```javascript
```
tyvdh, line 56: Does this actually matter?

tyvdh (changes requested):

docs/fundamentals-and-concepts/stellar-data-structures/assets.mdx
tyvdh: Yeah we really only need those blocks if there's a list of different code languages otherwise it's a list of examples with one example

tyvdh:
Sure why not and if prism supports json5 somehow let's add that too so CodeExample can render that as well. Otherwise no worries. I did just merge this though so you may need to create a new PR now 😬

ElliotFriend:
On it!

PR #54, opened 2022-11-15
Updated with new content

briwylde08:
Updated with publishing information about an asset


briwylde08:
Aaannnnd one more @tyvdh

PR #72, opened 2022-12-29
Inline Code Examples using Codepen.io

ElliotFriend:
## Summary

I've created an `<InlineSandbox />` custom component that will take a stellar operation name as a `module` prop and load the corresponding `index.js` and `index.html` file from [ElliotFriend/inline-stellar-examples](https://github.com/ElliotFriend/inline-stellar-examples) and place those into an editable, working codepen.

> As an example, I've embedded a `createClaimableBalance` example in the "Create Account" tutorial, and a `claimClaimableBalance` example in the "Send and Receive Payments" tutorial. (I will update with links once the preview has been generated).

## Remaining Work

- [ ] The custom component should probably check for a `404` if the provided `module` prop isn't found, and return `null` or something like that.
- [ ] So far, I've created examples for the `createClaimableBalance` and `claimClaimableBalance` operations, as well as `feeBumpTransactions` and `channelAccounts`. Bri has asked for half-a-dozen or so total.

## Outstanding Questions

- I'm using [Alpine](https://alpinejs.dev/) for a very simple state management solution. I was trying to keep these examples as framework-neutral as possible, and essentially as close to Vanilla JS as I could. It's possible there is a better way to go about that, though. I'm more than open to suggestions and comments.
- Inside the custom component, I'm worried I've used too much React stuff for a docusaurus site. Is it bad practice to "mix-n-match" like that?
- The actual examples will likely need (at least) a once-over from someone who's more experienced than I am at JS and/or front-end development. They "work" but they look pretty ugly and could probably be tuned up a bit.
- My tailwind chops are decidedly bad! I'm sure I'm doing it poorly, but that's about my level of knowledge with it Hehe.
- I've placed the snippets in that external repository. Is that a decent place for them? Long-term, does it make more sense to couple those here with the "official" docs repo? I _kinda_ tried going that route, but couldn't figure out how to load `js` or `html` files from the directory structure into the `<InlineSandbox />` component.

<a href="https://gitpod.io/#https://github.com/stellar/stellar-docs/pull/72"></a>


ElliotFriend:
I checked on the `Format MDX` error, and I have no idea what's going on... Seems like something over my head??

Here is the preview with the `createClaimableBalance` operation: https://developers-pr72.previews.kube001.services.stellar-ops.com/docs/tutorials/create-account

Here is the preview with the `claimClaimableBalance` operation: https://developers-pr72.previews.kube001.services.stellar-ops.com/docs/tutorials/send-and-receive-payments

ElliotFriend:
Rather than keeping the sandboxes willy-nilly on various pages, for now, i've created a separate "Sandboxes" menu entry with all the examples listed there. Here it is in the preview: https://developers-pr72.previews.kube001.services.stellar-ops.com/docs/category/sandboxes

briwylde08:
Closing per conversation with @ElliotFriend

PR #87, opened 2023-03-03
Update some code examples for accuracy

mollykarcher:


docs/encyclopedia/xdr.mdx
```diff
@@ -14,6 +14,6 @@ XDR is specified in [RFC 4506](http://tools.ietf.org/html/rfc4506.html) and is s
 
 Since XDR is a binary format and not as widely known as simpler formats like JSON, the Stellar SDKs all include tools for parsing XDR and will do so automatically when retrieving data.
 
-In addition, the Horizon API server generally exposes the most important parts of the XDR data in JSON, so they are easier to parse if you are not using an SDK. The XDR data is still included (encoded as a base64 string) inside the JSON in case you need direct access to it. .X files
+In addition, the Horizon API server generally exposes the most important parts of the XDR data in JSON, so they are easier to parse if you are not using an SDK. The XDR data is still included (encoded as a base64 string) inside the JSON in case you need direct access to it.
```
ElliotFriend, line 17:
```suggestion
In addition, the Horizon API server generally exposes the most important parts of the XDR data in JSON, so they are easier to parse if you are not using an SDK. The XDR data is still included (encoded as a base64 string) inside the JSON in case you need direct access to it.

## .X files
```

I believe `.X files` is meant to be another heading on this page. The link that follows points to a directory several IDL files named something like `Stellar-something.x`.

docs/tutorials/send-and-receive-payments.mdx
```diff
@@ -151,7 +151,7 @@ func main () {
 
     // Make sure destination account exists
     destAccountRequest := horizonclient.AccountRequest{AccountID: destination}
-    destinationAccount, err := client.AccountDetail(destAccountRequest)
+    _, err := client.AccountDetail(destAccountRequest)
```
ElliotFriend, line 154: I think keeping `destinationAccount` is useful for the context it adds to the example. The variable is never used, but I think the extra clarity makes it worth keeping here.

I'm not very familiar with Go, so preferring `_` may come with some benefits I don't know about. What do you think, @mollykarcher?

docs/tutorials/send-and-receive-payments.mdx
```diff
@@ -275,7 +277,7 @@ server.accounts().account(destination.getAccountId());
 
 ```go
 destAccountRequest := horizonclient.AccountRequest{AccountID: destination}
-destinationAccount, err := client.AccountDetail(destAccountRequest)
+_, err := client.AccountDetail(destAccountRequest)
```
ElliotFriend, line 280: See note above

ElliotFriend (commented):
Thanks so much for the contribution! It's really helpful to have folks like you submit updates to our example code when the SDKs are updated!

I left just a few suggestions/questions but overall seems like a pretty solid change. Thanks!

docs/tutorials/send-and-receive-payments.mdx
mollykarcher: Whereas declaring but not using a variable is typically a linting concern in most languages, it's a compile error in Go, so this doesn't compile as-written. That said, I agree it's more clear/understandable with a name, and this _is_ just a snippet so one could maybe argue that it doesn't necessarily need to compile, but I personally think it should.

docs/encyclopedia/xdr.mdx
mollykarcher: Ah! Will update

docs/tutorials/send-and-receive-payments.mdx
ElliotFriend: Aha!! That's very good to know! I agree that if we're providing it as an example, it should compile and be a useable snippet 👍🏻 

Would it be possible to do something like this to "use up" the `destinationAccount` variable?

```go
destAccountRequest := horizonclient.AccountRequest{AccountID: destination}
destinationAccount, err := client.AccountDetail(destAccountRequest)
_ = destinationAccount
```

After writing that, it looks pretty stupid... Also feels like it introduces a bigger "why" altogether...

Is there a way you can think to write this chunk that might more clearly reflect what the request is doing? Maybe just keeping the old line as a comment with a brief explanation?? 🤔 Don't know if I really like that, either...

@mollykarcher Sorry for the ramblings lol. You know this stuff better than I do, so I'll defer to you: What's the best and clearest approach here?

mollykarcher: Yeah I'm with you on that feeling weird. Alternatively, we could just add some print/log statements utilizing those accounts. Feels less superfluous and will actually use those variables

briwylde08:
Hello! @mollykarcher and @ElliotFriend, what's the status of this one? Still in progress?

docs/tutorials/send-and-receive-payments.mdx
ElliotFriend: > Alternatively, we could just add some print/log statements utilizing those accounts. Feels less superfluous and will actually use those variables

How did I miss this comment for 3 weeks!? I'm so sorry!

I like the idea of adding some print statements to the example. Would that be easy to do? Everything else with the PR looks great, and I'd love to get these fixes merged

ElliotFriend: Apparently, embarrassing myself is gonna be my thing today... 🤦🏻‍♂️ Since you've already made those changes, I'll just see myself out lol

Thank you!

ElliotFriend:
@briwylde08 I think this one is good to go!

ElliotFriend (approved):

briwylde08:
Woot thanks!!

PR #93, opened 2023-03-15
[ANCHOR-199] Anchor Platform guide: PATCH transaction examples

Ifropc:


docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+      "id": "<stellar_transaction_id>",
+      "payments": [
+        {
+          "id": "<payment_operation_id>",
+          "amount": {
+            "amount": "10",
+            "asset": "USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5"
+          }
+        }
+      ]
```
JakeUrban, line 694: This should not be the schema for `stellar_transactions`, although I just checked the OpenAPI spec and it matches.

We have a list of Stellar transactions, each with a list of payments. I think we need to flatten this to just a list of payments that each have a transaction and operation ID. We can update the docs after we update the platform.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+    -d '{"records":['"$transaction"']}'
+```
+
+</CodeExample>
+
+:::caution
+
+It's possible to send multiple transactions to fulfill user request, but due to current protocol restrictions, only the latest one will be visible for user via `stellar_transaction_id` field.
+
+:::
```
JakeUrban, line 712: I don't think its worth informing the business about this, or at least not as a "caution", since the business can't control this.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+
+#### Handling Errors
+
+When unrecoverable error happens, it's required to set transaction status to `error`. Message field can be used to describe error details.
```
JakeUrban, line 716:
```suggestion
If you encounter an unrecoverable error when processing the transaction, it's required to set transaction status to `error`. Message field can be used to describe error details.
```

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+    -d '{"records":['"$transaction"']}'
+```
+
+</CodeExample>
+
+:::tip
+
+This status should not be used after user has made a transfer (if there are some external factors that prevent completing transaction). Instead, refund should be initiated. It can be done either by user's request (in your web view), or automatically.
+
+:::
```
JakeUrban, line 741: Lets add an example for updating the platform of a refund payment.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+    -d '{"records":['"$transaction"']}'
+```
+
+</CodeExample>
+
+:::tip
+
+Setting `memo` and `memo_type` is optional. When status is changed to `pending_user_transfer_start`, Anchor Platform sets `memo` and `memo_type` automatically (only if it's not included in the request). Note that the memo must be unique, this is what helps to associate stellar transactions with SEP transaction.
+
+:::
```
JakeUrban, line 793: I would be a bit more explicit. If the business uses a third-party custodian, they need to generate a memo with their custodian that can be used to receive the user's payment. These memos should be generated per-transaction. If the business custodies their own assets, the anchor platform can generate memos for them.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+
+:::note
+
+The Stellar account that will be used to receive funds should be configured in [assets](#specify-your-services-assets)
+
+:::
+
+#### Tracking Stellar Transaction
```
JakeUrban, line 801: Instead of suggesting it in a tip, I would definitely guide businesses to use the observer unless they already have a system for monitoring payments made to their on-chain accounts.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+    -X PATCH \
+    -H 'Content-Type: application/json' \
+    -d '{"records":['"$transaction"']}'
+```
+
+</CodeExample>
+
+#### Sending External Transfer
+
+The next step is to send an external transfer, such as a bank transfer, to the user. When done, change status to notify user transfer has been sent, but has not been confirmed yet.
```
JakeUrban, line 854: Lets add that if the business' off-chain rails don't provide enough information on whether the payment is pending instead of delivered, the business should skip this step and mark payments as `completed` after initiating the payment.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -429,7 +926,8 @@ The decoded contents of the `token` will look something like this:
 
 <CodeExample>
 
-```json
```
JakeUrban, line 432: I don't think we should do this, since this isn't a transaction and is in a different section.

JakeUrban (commented):
Great start! I left lots of comments, but the content structure is there 👍

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -293,7 +307,7 @@ SECRET_SEP24_MORE_INFO_URL_JWT_SECRET="your encryption key shared with your busi
 
 To test this out, go to the [Stellar Demo Wallet][stellar-demo-wallet].
 
-![demo wallet connected to the anchor platform](/assets/anchor-platform-sep24-demo-wallet.png)
+[![demo wallet connected to the anchor platform](/assets/anchor-platform-sep24-demo-wallet.png)](/assets/anchor-platform-sep24-demo-wallet.png)
```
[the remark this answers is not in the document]
Ifropc: I added it to make image clickable (to open it in full size). 
It's not used anywhere else in docs, but I think it's a nice to have (images you added are high-resolution, so it only looks good on big resolution screens. When I had my window split in half, image is too small to see anything)

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
 
 You should see the logs reporting a successful connection to the postgres database.
 
+## Providing Updates To the Anchor Platform
```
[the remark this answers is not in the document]
Ifropc: I placed this before `Integration` section so that reader has a background to understand updates in the `Integration` section.
If we place it into the Integration, I think that nested headers will become be too complicated. (For example, menu on the right won't show sub-section of this section anymore and we will have level-5 headers)
Level 4 should be the limit, if there is a level-5 header it means sub-section should be separated into an independent section

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: TBD in another PR

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: Good idea

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: I went overboard replacing json examples content 🤦🏼

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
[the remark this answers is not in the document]
JakeUrban: Oh interesting, thats fine then.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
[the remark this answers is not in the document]
JakeUrban: I think having the `Providing Updates to the Anchor Platform` section as a top-level section is a bit out of place, since we have `Installation`, `Configuration`, and `Integration` currently. I don't think its a problem if our 5th level headers don't show up in the navigation on the right side.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+
+When the user has provided you the information necessary to facilitate the transaction, you need to update the status of the transaction and provide some of the information you collected to the Anchor Platform.
```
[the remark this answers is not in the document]
JakeUrban: If we want to keep it here we should generalize it a bit, since this sentence sounds like we're specifically talking about KYC collection, but the intro to this section should talk about the need to update the platform generally.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
```diff
@@ -401,6 +415,489 @@ docker compose up
...
+    "message": "KYC is being verified"
+  }
+}
+'
+
+curl localhost:8080/transactions \
+    -X PATCH \
+    -H 'Content-Type: application/json' \
+    -d '{"records":['"$transaction"']}'
+```
```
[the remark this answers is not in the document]
Ifropc: So, rollback to how it was done before?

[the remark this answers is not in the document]
JakeUrban: Mostly, yes, but we don't need to provide the curl command for each example.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: Yes, it's the current schema. I don't think we would be able to flatten it, because it will mess up all structure in Java code

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: It's something that business controls (it's Stellar transaction sent by business server)

JakeUrban: Oh I see you're talking about the deposit case. I don't think anchors have any reason to split the deposit payment into multiple transactions. Typically multiple payments occur because the user underpays in a withdrawal.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
JakeUrban: I'm not sure I understand, but we can discuss it later, no need to block this PR.

docs/anchoring-assets/anchor-platform/sep24-user-guide.mdx
Ifropc: In some rare cases there could be a reason to do so (for example, if there are multiple hot wallets and there is a system that prefers to empty old small wallets, that can result in multiple transactions)

JakeUrban: Interesting, I hadn't considered that. We can keep it.

JakeUrban (approved):

PR #94, opened 2023-03-16
Update openapi schema to include missing fields in asset schema

tamirms:
I have updated the asset stats schema to resemble the definition in horizon:

https://github.com/stellar/go/blob/soroban-xdr-next/protocols/horizon/main.go#L163-L184

Note that the `num_contracts` and `contracts_amount` fields will only be available once protocol 20 (soroban smart contracts) are released.

Part of https://github.com/stellar/go/issues/4727


Ifropc:
FYI: plugin uses bundled file. Modifying any other file won't change anything in generated .mdx
You need to re-generate bundled file first. (I added README for that)

ElliotFriend:
> (I added README for that)

Beat me to it by about 30 minutes lol. Glad to know I was on the right track 👍🏻

package.json
```diff
@@ -53,6 +56,7 @@
   },
   "devDependencies": {
     "@docusaurus/eslint-plugin": "^2.1.0",
+    "@redocly/cli": "^1.0.0-beta.124",
```
ElliotFriend, line 59: Is it a bad idea to add `redocly` here as a dependency? Would it just be better to use the `npx` method of running the bundle command?

Ifropc: I think it's ok to include as dependency

ElliotFriend: Sounds good to me. Thanks!

openapi/main.yml
```diff
@@ -1,38 +1,38 @@
 openapi: 3.0.3
 info:
-  title: Horizon 
+  title: Horizon
   version: 0.0.1
```
Shaptic, line 4: Should we update this to match the Horizon version string?

openapi/README.md
```diff
@@ -0,0 +1,90 @@
...
+   trouble over-writing files that already exist (or figuring out if those files
+   should be overwritten, perhaps?). So, you can use this script to clean up the
+   existing generated files.
+
+```bash
+yarn api:clean
+```
+
+2. Bundle your changes into the `bundled.yml` file. We've gone over this a bit,
+   so I wont' repeat myself. Bring any changes you've made in the other
```
Shaptic, line 61: nit: typo:

```suggestion
   so I won't repeat myself. Bring any changes you've made in the other
```

ElliotFriend: Nice catch!

openapi/main.yml
ElliotFriend: This is an excellent question! I had the same thought somewhere along the way. I suppose it would open up a-whole-nother conversation about if we want to _keep_ versioned docs for horizon, and how that should be done? [Here's the plugin docs on that](https://docusaurus-openapi.tryingpan.dev/versioning), seems way above my paygrade 🤣

ElliotFriend (approved):
Looks good to me! Thanks for keeping horizon up-to-date @tamirms!! 👍🏻

Issue #97, opened 2023-03-16
TeX doesn't render


Shaptic:
On pages like [this one](https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/assets#amount-precision), there is use of LaTeX syntax for math, but it does not render correctly. It's probably missing a MathJax script?

ElliotFriend:
Great catch! Thanks for pointing it out. I'll take a look

PR #99, opened 2023-03-17
Update config to support equation rendering with KaTeX

ElliotFriend:
Looks like all the packages and CSS were installed, but the `require` statements were just missing from the config file.

Issue #97

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -73,7 +73,8 @@ const config = {
           showLastUpdateTime: true,
           breadcrumbs: true,
           routeBasePath: "/docs",
-          remarkPlugins: [require("mdx-mermaid")],
+          remarkPlugins: [require("mdx-mermaid"), require('remark-math')],
+          rehypePlugins: [require('rehype-katex')],
           sidebarPath: require.resolve("./sidebars.js"),
           editUrl: "https://github.com/stellar/stellar-docs/tree/main",
         },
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -85,10 +86,10 @@ const config = {
   ],
   stylesheets: [
     {
-      href: "https://cdn.jsdelivr.net/npm/katex@0.16.2/dist/katex.min.css",
-      type: "text/css",
-      integrity: "sha256-oWCabCfPd4Oi21wqZezBSz/anto4VYcJqc9sM9IzQTk=",
-      crossorigin: "anonymous",
+      href: 'https://cdn.jsdelivr.net/npm/katex@0.13.24/dist/katex.min.css',
```
Shaptic, line 89: why is the version going down? :confused:

ElliotFriend: The docusaurus docs have a note about version compatibility: https://docusaurus.io/docs/markdown-features/math-equations#configuration

It says something like "use the exact same versions of these packages". It could mean only in the package installation steps, but I just copy/pasted all the relevant bits from that page.

Shaptic: ah gotcha. good context, thanks!

Shaptic (approved):
Confirmed that [the page](https://developers-pr99.previews.kube001.services.stellar-ops.com/docs/fundamentals-and-concepts/stellar-data-structures/assets#amount-precision) looks right in the preview, thanks!

PR #104, opened 2023-03-27
update base fee wording

briwylde08:


docs/encyclopedia/fees-surge-pricing-fee-strategies.mdx
```diff
@@ -8,13 +8,13 @@ Stellar requires a small fee for all transactions to prevent ledger spam and pri
 
 ## Network fees on Stellar
 
-The minimum fee for a given transaction equals the number of operations in the transaction multiplied by the base fee for the given ledger. (transaction fee = # of operations \* base fee)
+The fee for a given transaction equals the number of operations in the transaction multiplied by the base fee for the given ledger. (transaction fee = # of operations \* base fee)
 
-**Base fee:** the fee the network charges per operation in a transaction. The base fee is determined using a version of a [VCG auction](https://en.wikipedia.org/wiki/Vickrey%E2%80%93Clarke%E2%80%93Groves_auction). The base fee is currently 100 stroops per operation.
+**Base fee:** the fee you’re willing to pay per operation in a transaction. The base fee cannot be set to lower than 100 stroops per operation.
```
ElliotFriend, line 13: Ooooohh, this is a really interesting and tricky one!

There's almost two ways to think of it: from the transaction point of view, and the overall network point of view.

- From the transaction, the changes toward "that you're willing to pay" make perfect sense.
- But, for the network, the base fee is imposed on transactions, where it would make more sense to use wording like "the network charges," and those transactions can offer more/less accordingly. I think the `minimum` on Line11 helps convey idea, so I might want to see it kept.

Maybe there could be two descriptions here? "Ledger Base Fee" and "Transaction Base Fee"? Not sure.

ElliotFriend (approved):
Looks great! One tiny quibble that you are free to ignore 😁

PR #122, opened 2023-04-20
Anchor Platform: SEP-31 Documentation

JakeUrban:
- Removes "Enabled" from "Hosted Deposits & Withdrawals" section title
- Fixes docker image name to use `release-` prefix
- Move sections shared between SEP-24 & SEP-31 into top-level "Getting Started" section
  - includes assets config, env vars vs. yaml config, and data persistence
- Adds overview and config sections for SEP-31
- Integration section is TODO


docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
```diff
@@ -0,0 +1,368 @@
...
+}
+```
+
+</CodeExample>
+
+## Enable the Customer KYC API
+
+Businesses need to collect and validate KYC information on the customers they're facilitating transactions for. Clients determine what KYC information needs to be collected and send that information via a SEP-12 KYC API hosted by the Anchor Platform, but the Anchor Platform never stores personally-identifiable information (PII). Instead, it forwards requests from clients to the business server, and returns the business' responses back to the client, acting as a proxy server.
+
+See the _Anchor Platform KYC API specification_ for details on the endpoints that must be implemented on your business' server.
```
Ifropc, line 116: Link missed?

docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
```diff
@@ -0,0 +1,368 @@
...
+SEP12_ENABLED=true
+```
+
+</CodeExample>
+
+Finally, we have to define your business' customer types. Each type of customer requires different a set of KYC information. For example, you can offer your cross-border payments service in two distinct regulatory jurisdictions, so customers in different jurisdictions have different KYC requirements and would be represented using different types.
+
+:::info
+
+Currently, customer types must be mutually exclusive, meaning a customer cannot be more than one type. For example, business' cannot model a tiered KYC model with "basic" and "pro" customer types.
```
Ifropc, line 146: They technically can, if this types are mutually exclusive. I think it's good idea to elaborate:
"Meaning, user can not be assigned to both 'basic' and 'pro' types"

docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
```diff
@@ -0,0 +1,368 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+</CodeExample>
+
+## Configure Platform API Authentication
```
Ifropc, line 329: It's essentially the same as "Securing Platform API" section in SEP-24. Should it be moved to the general section instead?

docs/anchoring-assets/anchor-platform/sep31/integration.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+- [`GET /customer`][get-customer] & [`PUT /customer`][put-customer] KYC API endpoints to request & collect customers' KYC data
+- [`GET /rate`][get-rate] RFQ API endpoint to provide FX rates between the on & off-chain assets supported
+- [`PATCH /transactions`][patch-transactions] requests to update the Anchor Platform's transactions' statuses
+
+The following may also be required depending on your use case:
+
+- [`GET /fee`][get-fee] if your business wants to provide senders the option to skip the quote creation step
+- [`GET /unique_address`][get-unique-address] if your business uses a custody service for on-chain assets
+- [`PUT /customer/callback`][put-customer-callback] if your business wants to provide webhooks for customer status change events
```
Ifropc, line 16: Can you remove it because platform doesn't support it at the moment?

docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
JakeUrban: Good point, a user could be both basic and pro as long as pro users can do everything basic can do, so its not a good example. I removed it.

PR #123, opened 2023-04-20
Basic OpenAPI support for SEP-31

Ifropc:


JakeUrban:
1. Can we capitalize? So `Delete Customer Data` instead of `Delete customer data`
2. Can we make the words consistent with Horizon? So instead of "Get Customer KYC Info" we can do "Retrieve a Customer's Info". For `PUT`, we can do "Create or Update a Customer's Info". "Get a Unique Address" too.

JakeUrban (approved):

PR #126, opened 2023-04-24
Add badges for OpenAPI docs

Ifropc:
Add SEP badges to auto-generated openAPI docs.
This should make it easier for user to understand which SEP this API endpoint belongs to.


JakeUrban (approved):
Nice addition, very helpful!

PR #130, opened 2023-04-27
Anchor Platform: SEP-31 integration guide

JakeUrban:
Adds documentation on the various endpoints and requests need to be made.


docs/anchoring-assets/anchor-platform/sep31/integration.mdx
```diff
@@ -15,6 +18,258 @@ The following may also be required depending on your use case:
...
+{
+  "id": "fb5ddc93-1d5d-490d-ba5f-2c361cea41f7"
+}
+```
+
+</CodeExample>
+
+You can use any identifer for your customers as long as it is a string.
+
+Following the registration of a customer, the sending organization can use either approach when checking the customer's status. For example, you may get a `GET /customer` request like the following:
```
Ifropc, line 101: "you" as "your (business) server", right? A bit unclear

docs/anchoring-assets/anchor-platform/sep31/integration.mdx
```diff
@@ -15,6 +18,258 @@ The following may also be required depending on your use case:
...
+  platform:
+    image: stellar/anchor-platform:2.0.0-rc.2
+    command: --sep-server
+    ports:
+      - "8080:8080"
+    env_file:
+      - ./dev.env
+    volumes:
+      - ./config:/home
+  server:
```
Ifropc, line 40: Can you add a paragraph (roughly):

```
For local development, you can start your business server on your local machine. To reference your implementation, update configuration replacing `server` to `host.docker.internal`. See [docker documentation](https://docs.docker.com/desktop/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host) for more information

:::tip

For Linux machines, it's required to use host-gateway. For each services, following configuration should be added:
services:
  service:
    extra_hosts:
      - host.docker.internal:host-gateway
      
See [github](https://github.com/docker/for-linux/issues/264#issuecomment-965465879) for more info

:::
```

JakeUrban: I'm not sure this is really in scope. They can choose to run their business server outside docker, but if they choose to go that route, they can figure out how to connect their anchor platform instance to it.

JakeUrban: I've been running my business server within docker and just have it auto-reload when the code changes.

Ifropc: If you think so. Not all languages support hot reload

docs/anchoring-assets/anchor-platform/sep31/integration.mdx
```diff
@@ -15,6 +18,258 @@ The following may also be required depending on your use case:
...
+    "message": "a deposit to the recipient's bank has been initiated"
+  }
+}
+```
+
+</CodeExample>
+
+## Fee Callback Endpoint
+
+Your business may want to offer sending organizations the option to skip the quote creation process, allowing your business to use a rate determined at the time funds are paid out to the recipient. In this case, the Anchor Platform will not make a `GET /rate` request, but you will still need to provide the fee your business will charge for these types of transactions using the [`GET /fee`][get-fee] endpoint.
```
Ifropc, line 226: We say "may want to offer", but there are no explicit way to prohibit it, right? 
If they don't implement GET /fee it means they do not offer such functionality. As the result, /transaction endpoint will simply return 500 because trying to get a callback to /fee will throw.
Is my understanding correct?

Ifropc (approved):

docs/anchoring-assets/anchor-platform/sep31/integration.mdx
JakeUrban: The Anchor Platform shouldn't even make the `GET /fee` request to the business if `quotes_supported` is not `true`. So the Anchor Platform should return 400 if they try not to pass a quote.

Ifropc: You mean quotes_required?

JakeUrban: Yes my mistake. So if `quotes_required` is true, the anchor platform won't make the `GET /fee` request.

PR #132, opened 2023-04-27
Small schema fix

Ifropc:
Fixed sep field marked as required in patch transactions


Ifropc:
Yep,
org.stellar.anchor.platform.service.TransactionService#findTransaction

PR #146, opened 2023-05-15
Remove CATCHUP_RECENT recommendation when reaping history

mollykarcher:
We don't recommend this, and several partners have reached out having been misled by this


rice2000 (approved):
Thanks!

PR #151, opened 2023-05-18
Fix assets

Ifropc:


docs/anchoring-assets/anchor-platform/sep24/configuration.mdx
```diff
@@ -56,11 +56,9 @@ assets:
       enabled: true
     withdraw:
       enabled: true
-assets:
   - schema: iso4217
     code: USD
     significant_decimals: 2
-    sep24_enabled: true
```
JakeUrban, line 63: Can you verify whether `sep24_enabled: true` is necessary for amount validation? We may not need `deposit:` or `withdraw:` objects either. It would be good to iron out the config requirements for using fiat assets

Ifropc: We have it disabled in e2e tests.
Deposit/withdrawal I'm pretty sure required before https://stellarorg.atlassian.net/browse/ANCHOR-263

JakeUrban: Oh you're right. But that will be resolved before 2.0, so we can update the docs once its released.

JakeUrban (approved):

PR #153, opened 2023-05-25
Fixing `*` parsing in Markdown on federation page

ElliotFriend:
The federation encyclopedia entry has an errant `*` that is being parsed as a Markdown italics formatter: https://developers.stellar.org/docs/encyclopedia/federation#supporting-federation

This commit fixes that bug, and also cleans up a bit of the language surrounding it.


Shaptic (approved):
I would wait on the preview to sanity check, but lgtm!

PR #157, opened 2023-06-05
Add git LSF support 

Ifropc:


Ifropc:
Doesn't seem to work with PR previews and current files are not that big to use LFS

PR #159, opened 2023-06-06
Add missed country code

Ifropc:
When country code was not specified it throws NPE

### DIFF
--- a/docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
+++ b/docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
@@ -296,6 +296,8 @@ assets:
     sep38:
       exchangeable_assets:
         - iso4217:BRL
+      country_codes:
+        - BRA
   - schema: iso4217
     code: BRL
     sep38_enabled: true


docs/anchoring-assets/anchor-platform/sep31/configuration.mdx
```diff
@@ -296,6 +296,8 @@ assets:
     sep38:
       exchangeable_assets:
         - iso4217:BRL
+      country_codes:
+        - BRA
```
JakeUrban, line 300: I don't know if we actually need to have businesses configure country codes for on-chain assets.. can't that be limited to off-chain assets, since thats where the on/off-ramping occurs?

Ifropc: Right now if not specified it will give an error. We can look into improving asset configuration separately (so the country code is not required for on-chain assets anymore), WDYT?

JakeUrban: Sounds good to me.

JakeUrban (approved):

Issue #147, opened 2023-05-15
Horizon Ingestion docs consolidation


sreuland:
### What problem does your feature solve?
[Horizon ingestion docs](https://developers.stellar.org/docs/run-api-server/ingestion) refer to many aspects of ingestion with variations of configuration, approach, and references to command line flag, environment variables some of which have become obsolete or not relevant. It may lead to confusion or un-necessary steps for readers. 

### What would you like to see?
[Horizon ingestion docs](https://developers.stellar.org/docs/run-api-server/ingestion) are updated to present smaller copy, with direct, opinionated writeup on what/why ingestion is needed, the best strategy to follow for desired outcome, and concise updated examples of using the latest(and least number of) command line flags and config variables.

Review state verification and history retention(reaper) aspects and concise configuration examples to achieve.

Mention how to use ingestion filtering. Include concise configuration examples of Asset and Account filter rules and how there outcomes affect data stored in history. 

Discuss the need to use captive core mode of ingestion only, and the config settings for ensuring on-disk usage. 

Should we mention trying to do full history or discourage that in docs with explanation and recommend using DataWarehouse/Hubble for those concerns?

sreuland:
merged on https://github.com/stellar/stellar-docs/pull/168

PR #162, opened 2023-06-12
Partner Experience: update horizon config/install copy

### LINKED ISSUE #141: Horizon config section for more opinionated best practices
Partner Experience initiative - consolidate horizon config docs section, review and update/refactor the copy to be lighter footprint but with more opinionated direction towards a single group of settings to enable user's to quickly get through it and have a decent running system afterwards, rather than documenting all settings individually and letting the user figure out how to group/associate them:

How much history should they retain?
Should they use in-memory or on-disk mode?
How often should they run state verification? Use the new config params …
What version of postgresql should they run?
How often (hopefully, never) do they need to manually vacuum their db?

Updated copy on horizon config in https://developers.stellar.org/docs/run-api-server/configuring

sreuland:
The main goal of docs updates in terms of partner experience is to make them more opinionated on what to do for configuration with less choices so it is faster for setup and less possibility of choosing wrong settings. The effort here was to remove many options and instead present as few but generally robust options that are geared towards latest settings in horizon. 

This change is focused on just runtime configuration(not ingestion yet, that is another task). `Configuring` was refactored and I pulled in reformat of `Installing` as it was closely tied into how configurations get done as well. You can view new pages here:

https://developers-pr162.previews.kube001.services.stellar-ops.com/docs/run-platform-server/installing
https://developers-pr162.previews.kube001.services.stellar-ops.com/docs/run-platform-server/configuring

To that means, have also made leap in docs to suggest features/flags in Horizon to enable these streamlined configs, but don't exist yet, and therefore would need to be reviewed to determine if we want to add them. I've highlighted any functional gaps with **TODO** inline in docs to mark where these are.

Closes #141


docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+title: Configuring
+sidebar_position: 30
+---
+
+import { Alert } from "@site/src/components/Alert";
+import { CodeExample } from "@site/src/components/CodeExample";
+
+Pre-requisites:
+
+- [Installation](./installing.mdx) is completed on host machine, it should have two executables, `stellar-horizon` and `stellar-core`. Per the installation instructions, `stellar-core` should be included in your host o/s PATH as these configurations will result in Horizon searching for `stellar-core` to execute.
```
Shaptic, line 11:
```suggestion
- [Installation](./installing.mdx) is completed on a host machine and results in two executables: `stellar-horizon` and `stellar-core`. Per the installation instructions, `stellar-core` should be included in your host OS' `$PATH`, as the configurations below result in Horizon searching for a `stellar-core` to execute.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+- [Prepare the Database if first time](#prepare-the-database-first-time)
+
+You are now ready to setup configuration parameters to perform three important, distinct roles:
+
+- **serving requests** a regular web-based HTTP API,
+- **ingesting ledgers** from core nodes of a Stellar network to keep its world-view up to date, and
+- **transaction submission** for sending transactions to the Stellar network.
+
+You choose from one of two following deployment modes to perform the roles and proceed with configuration parameters to that means. A note on configuration parameters, the same parameter can be provided either as command line argument or an environment variable.
```
Shaptic, line 20: I find the "configuration parameters to that means" wording a bit confusing. Here's a suggested reword, plus moving the note to its own "Info" block:

```suggestion
To perform these roles, you can choose from one of two of the deployment modes that follow. Each has its own configuration parameters.

<Alert>

In all configuration examples, the same parameter can be provided either as command line argument or an environment variable. If both are provided, the command line takes precedence.

</Alert>

```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+## Single Instance Deployment
+
+You run `stellar-horizon` in a single o/s process and it will perform all three roles at same time.
```
Shaptic, line 24:
```suggestion
In this simpler deployment variant, you run `stellar-horizon` in a single OS process and it will perform all three roles at same time.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+| flag               | envvar         | example                              |
+| ------------------ | -------------- | ------------------------------------ |
+| `--db-url`         | `DATABASE_URL` | postgres://localhost/horizon_testnet |
+| `--network`        | `NETWORK`      | pubnet                               |
+| `--log-level`      | `LOG_LEVEL`    | info                                 |
+| `--tx-sub-enabled` | TX_SUB_ENABLED | true                                 |
```
Shaptic, line 31: this isn't a real flag, right?

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+## Single Instance Deployment
+
+You run `stellar-horizon` in a single o/s process and it will perform all three roles at same time.
+
+| flag               | envvar         | example                              |
+| ------------------ | -------------- | ------------------------------------ |
+| `--db-url`         | `DATABASE_URL` | postgres://localhost/horizon_testnet |
+| `--network`        | `NETWORK`      | pubnet                               |
+| `--log-level`      | `LOG_LEVEL`    | info                                 |
```
Shaptic, line 30: I'm assuming this will be the default, so we can drop it to simplify the config?

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+You choose from one of two following deployment modes to perform the roles and proceed with configuration parameters to that means. A note on configuration parameters, the same parameter can be provided either as command line argument or an environment variable.
+
+## Single Instance Deployment
+
+You run `stellar-horizon` in a single o/s process and it will perform all three roles at same time.
+
+| flag               | envvar         | example                              |
+| ------------------ | -------------- | ------------------------------------ |
+| `--db-url`         | `DATABASE_URL` | postgres://localhost/horizon_testnet |
+| `--network`        | `NETWORK`      | pubnet                               |
```
Shaptic, line 29: Isn't it `--network-passphrase` and the full thing? Though this would be an awesome simplification.

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+| `--log-level`      | `LOG_LEVEL`    | info                                 |
+| `--tx-sub-enabled` | TX_SUB_ENABLED | true                                 |
+
+## Multiple Instance Deployment
+
+You run multiple instances of `stellar-horizon` each performs a subset of the roles.
```
Shaptic, line 35:
```suggestion
In this scalable deployment variant, you run multiple instances of `stellar-horizon` and each performs a subset of the roles.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+### Ingestion Role Instance
+
+You need to allocate one instance to perform ongoing ingestion to capture network activity. We recommend limiting the amount of network activity to retain in the db on a sliding window of the last 30 days, we include `--history-retention-count 518400` which is number of ledgers in 30 days. **TODO - make --history-retention-count default to 518400, don't specify here anymore. make --captive-core-use-db=true by default**
```
Shaptic, line 39:
```suggestion
You need to allocate one instance to perform ongoing ingestion to capture network activity and store historical data. We recommend limiting the amount of historical data retained in the database on a sliding window of the last 30 days. This is controlled by the `--history-retention-count` parameter below, set to `518400` which is approximately the number of ledgers in 30 days. **TODO - make --history-retention-count default to 518400, don't specify here anymore. make --captive-core-use-db=true by default**
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+| flag | envvar | example |
+| --- | --- | --- |
+| `--db-url` | `DATABASE_URL` | postgres://localhost/horizon_testnet |
+| `--history-retention-count` | `HISTORY_RETENTION_COUNT` | 518400 |
+| `--network` | `NETWORK` | pubnet |
+| `--log-level` | `LOG_LEVEL` | info |
+
+### API Role Instance
+
+You can run none or multiple instances to serve read only API requests. **TODO - create --tx-sub-enabled flag, default=false. Should return 405 on POSTs to /tx if enabled. this allows skipping need to define --stellar-core-url on api servers since instance doesn't need to be aware of core.**
```
Shaptic, line 50:
```suggestion
You can run zero or more instances to serve read-only API requests. **TODO - create --tx-sub-enabled flag, default=false. Should return 405 on POSTs to /tx if enabled. this allows skipping need to define --stellar-core-url on api servers since instance doesn't need to be aware of core.**
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+| flag               | envvar           | example                              |
+| ------------------ | ---------------- | ------------------------------------ |
+| `--db-url`         | `DATABASE_URL`   | postgres://localhost/horizon_testnet |
+| `--log-level`      | `LOG_LEVEL`      | info                                 |
+| `--tx-sub-enabled` | `TX_SUB_ENABLED` | false                                |
+
+### Transaction Submission Role Instance
+
+You can run none or multiple instances to serve transaction submission requests. If you run an instance with transaction submission enabled, yuur deployment is required to perform the ingestion role on the same database referenced here. Horizon transaction submission depends on ingestion to confirm tx submission status. **TODO - Confirm Horizon can use captive core to send tx sub requests and skip stellar-core-url?. If --tx-sub-enabled=true, horizon should check at startup that ingestion is happening on it's db or exit startup with error.**
```
Shaptic, line 60:
```suggestion
You can run zero or more instances to serve transaction submission requests. If you run an instance with transaction submission enabled, your deployment is **required** to perform the ingestion role on the same database referenced by this instance. Horizon's transaction submission depends on ingestion to confirm transaction submission status. **TODO - Confirm Horizon can use captive core to send tx sub requests and skip stellar-core-url?. If --tx-sub-enabled=true, horizon should check at startup that ingestion is happening on it's db or exit startup with error.**
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+| flag | envvar | example |
+| --- | --- | --- |
+| `--db-url` | `DATABASE_URL` | postgres://localhost/horizon_testnet |
+| `--network` | `NETWORK` | pubnet |
+| `--log-level` | `LOG_LEVEL` | info |
+| `--stellar-core-url` | STELLAR_CORE_URL | http://localhost:11626 |
+| `--tx-sub-enabled` | `TX_SUB_ENABLED` | true |
+
+## Notes
```
Shaptic, line 70: imo this section should come before the example configs

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+- `--db-url`, required, specifies the Horizon database. It's value follows this format: `dbname=<pg_dbname> user=<pg_user_name> password=<pg_user_pwd> host=<host_address>`
+
+- `--log-level`, optional, can be one of 'info', 'error', 'debug'.
+
+- Horizon will create a sub-directory under the current working directory of the o/s process to store captive core runtime data files. Expect up to 10GB usage(**TODO - verify the max storage needs of core/buckets on disk?**). You can override this location with the optional `--captive-core-storage-path`, set to a directory on file system which captive core wiull store the runtime files.
+
+### Passing configurations to Horizon
+
+Depends on how Horizon was installed:
```
Shaptic, line 86: We should have complete sentences opening a paragraph,

```suggestion
Depending on how Horizon was installed, the way to configure it differs:
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+    </Alert>
+- containerized
+  - non-Helm, pass all configuration parameters to the horizon docker image as [docker environment variables](https://docs.docker.com/engine/reference/commandline/run/#env).
+  - Helm, pass all configuration parameters in the [Helm Install command](https://helm.sh/docs/helm/helm_install/)
+
+## Prepare the Database first time
+
+Before running the Horizon server first time, you must initialize the Horizon database. This database will be used for all of the information produced by Horizon, most notably historical information about transactions that have occurred on the Stellar network.
+
+To prepare a database for Horizon's use, first ensure it is blank. It's easiest to create a new database on your PostgreSQL server specifically for Horizon's use. We recommend creating a new user(role) in postgres dedicated for horizon's db and assigning that user(role) as the owner of the new horizon db. To illustrate this as example using `psql`, first login to databse server using the `psql` cli as a superuser, and then create the new user(role) and db for horizon:
```
Shaptic, line 105:
```suggestion
To prepare a database for Horizon's use, first ensure it is blank. It's easiest to create a new database on your PostgreSQL server specifically for Horizon's use. We recommend creating a new user(role) in postgres dedicated to Horizon's database and assigning that user(role) as the owner of this database. 

To illustrate an example using `psql`, first login to database server using the `psql` command-line tool as a superuser, and then create the new user(role) and database for Horizon:
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+postgres=#
+postgres=# CREATE ROLE horizon WITH LOGIN;
+CREATE ROLE
+postgres=#
+postgres=# CREATE DATABASE horizon OWNER horizon;
+CREATE DATABASE
+postgres=#
+```
+
+Additionally set a password on your new `horizon` user using `ALTER USER`.
```
Shaptic, line 117:
```suggestion
Additionally, you can set a password on your new `horizon` user using `ALTER USER`.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+Given this example, you can now complete the `DATABASE_URL="dbname=horizon user=horizon password=<value_here> host=<address_of_pg_server>" configuration parameter.
```
Shaptic, line 119: code block isn't closed 

```suggestion
Given this example, you can now complete the `DATABASE_URL="dbname=horizon user=horizon password=<value_here> host=<address_of_pg_server>"` configuration parameter.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+Next, use the Horizon binary to install the schema onto the empty db from command line:
+
+```
+$ export DATABASE_URL="dbname=horizon user=horizon password=<value_here> host=<address_of_pg_server>"
+$ stellar-horizon db init --db-url
```
Shaptic, line 125: You don't need this since you `export`ed above:

```suggestion
$ stellar-horizon db init
```

docs/run-platform-server/index.mdx
```diff
@@ -3,16 +3,12 @@ title: "Overview"
 sidebar_position: 0
 ---
 
-Horizon is responsible for providing an HTTP API to data in the Stellar network. It ingests and re-serves the data produced by the Stellar network in a form that is easier to consume by the average application relative to the performance-oriented data representations used by Stellar Core.
+Horizon is a central component of the Stellar Platform, it provides an HTTP API to data in the Stellar network. It ingests and re-serves the data produced by the Stellar network in a form that is easier to consume by the average application relative to the performance-oriented data representations used by Stellar Core.
```
Shaptic, line 6:
```suggestion
Horizon is a central component of the Stellar Platform: it provides an HTTP API to data in the Stellar network. It ingests and re-serves the data produced by the Stellar network in a form that is easier to consume by the average application relative to the performance-oriented data representations used by Stellar Core.
```

Shaptic (commented):
I did a first-pass review! 

Lots of minor comments oriented around phrasing and "proper" documentation. Imo, as a general rule, official docs should be as easy to copy-paste as possible, so I tried to make suggestions oriented around e.g. maintaining complete sentences, expanding abbreviations, phrasing changes for clarity, etc.

There are also a handful of grammatical fixes. We should definitely get devrel (and maybe even someone from marketing) do a proper wording + grammar pass before merging this branch into `main`.

docs/run-platform-server/configuring.mdx
sreuland: not yet, I'm proposing to add it as a TODO comment on https://github.com/stellar/stellar-docs/pull/162/files/6ae0bd3a50eccb4a7f8ea3ce99bb89d389321db5#diff-a2be184a631b87a98857de0dd0abe901235c4639179be7c62719b5ab4e6502ddR50

docs/run-platform-server/configuring.mdx
sreuland: not quite, this is proposing a new flag `--network = [pubnet | testnet | other]` as a way to allow automating remainder of network specific flags and the network toml for captive core, I added a TODO with a bit more justification on using this new flag - https://github.com/stellar/stellar-docs/pull/162/files/6ae0bd3a50eccb4a7f8ea3ce99bb89d389321db5#diff-a2be184a631b87a98857de0dd0abe901235c4639179be7c62719b5ab4e6502ddR74

sreuland:
> There are also a handful of grammatical fixes. We should definitely get devrel (and maybe even someone from marketing) do a proper wording + grammar pass before merging this branch into `main`.

yes, these changes won't be landing on `main` in short term, the plan is to use feature branch `partner_experience`, to stage [related docs](https://github.com/orgs/stellar/projects/37/views/14) as a whole. The feature branch will also gate on review and implementation of any new functionality in code that docs refer to as TODO proposals before merging to `main`.

docs/run-platform-server/configuring.mdx
sreuland: what's your opinion on using second person pronouns such as 'you'? briefly looking at tech/academic guides, they suggest rephrasing to 1st/3rd, the former looks widely used in docs, and assuming best to stay consistent for overall style?

docs/run-platform-server/configuring.mdx
sreuland: nice catch, thanks.

docs/run-platform-server/installing.mdx
```diff
@@ -0,0 +1,44 @@
...
+import { CodeExample } from "@site/src/components/CodeExample";
+
+To install Horizon, you have choices, we recommend the following for target infrastructure:
+
+- bare-metal:
+  - if host is debian linux, install prebuilt binaries [from repositories](#package-manager) using package manager.
+  - for any other hosts, download [prebuilt release binaries](#prebuilt-releases) of Stellar Horizon and Core for host target architecture and operation system.
+- containerized:
+  - non-Helm Chart, if the target envrironment for container to run does not support Helm chart usage, run the prebuilt docker image of Horizon published on [dockerhub.com/stellar/horizon](https://hub.docker.com/r/stellar/stellar-horizon).
+  - Helm charts, when the target envrionment uses container orchestration such as Kubernetes and has enabled Helm Charts on cluster. The Horizon Helm chart manages installation life cycle. Use the [Helm Install command](https://helm.sh/docs/helm/helm_install/), it will accept Horizon's configuration parameters. Please review [Configuration](./configuring.mdx) first, to identify any specific configuration params needed. **TODO - create the Horizon Helm Chart.**
```
sreuland, line 15: Hello @jacekn , as part of new documentation, we are referring to the potential Helm Chart usage for Horizon. Wondering if we shouild spin up a dev ticket for Horizon Helm Chart on the mono repo, or if you are aware of Helm for Horizon already, should we coordinate with an Ops ticket for this feature? CC'ing @urvisavla

Shaptic (commented):
ok to merge w/ TODOs? if yes I can :heavy_check_mark: but @mollykarcher / Jesse (what's his GH username??) should look as well for product/partner perspective

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+Depends on how Horizon was installed:
+
+- bare-metal
+  - non-package manager, we recommend using environment variables to pass configurations. There are many tools you can use to manage them, such as [direnv](http://direnv.net/) or [dotenv](https://github.com/bkeepers/dotenv).
+  - [package manager](./installing.mdx#package-manager), the provided `stellar-horizon-cmd` wrapper will import a configuration from `/etc/default/stellar-horizon`, hence, if you want to set configurations, edit the file at `/etc/default/stellar-horizon`.
+    <Alert>
+      This script invokes Horizon with the `stellar` user, so make sure that
+      permissions for the user are set up accordingly. The current working
+      directory should be writable for this user and the user should be able to
+      execute the `horizon` and `stellar-core` binaries; etc.
```
urvisavla, line 111: Should this be "execute the `stellar-horizon`" instead since the name of the binary is stellar-horizon?

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+- If you want to connect Horizon to a different stellar network or override any of the defaults that horizon will make, the key configurations that can be set are: `--history-archive-urls`, `--captive-core-config-path`, `--network-pass-phrase`, `--captive-core-storage-path`, `--stellar-core-url`.
+
+- `--db-url`, required, specifies the Horizon database. It's value follows this format: `dbname=<pg_dbname> user=<pg_user_name> password=<pg_user_pwd> host=<host_address>`
+
+- `--log-level`, optional, can be one of 'info', 'error', 'debug'.
+
+- Horizon will create a sub-directory under the current working directory of the o/s process to store captive core runtime data files. Expect up to 10GB usage(**TODO - verify the max storage needs of core/buckets on disk?**). You can override this location with the optional `--captive-core-storage-path`, set to a directory on file system which captive core wiull store the runtime files.
```
urvisavla:
```suggestion
- Horizon will create a sub-directory under the current working directory of the o/s process to store captive core runtime data files. Expect up to 10GB usage(**TODO - verify the max storage needs of core/buckets on disk?**). You can override this location with the optional `--captive-core-storage-path`, set to a directory on file system which captive core will store the runtime files.
```

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,145 @@
...
+  - Helm, pass all configuration parameters in the [Helm Install command](https://helm.sh/docs/helm/helm_install/)
+
+## Prepare the Database first time
+
+Before running the Horizon server first time, you must initialize the Horizon database. This database will be used for all of the information produced by Horizon, most notably historical information about transactions that have occurred on the Stellar network.
+
+To prepare a database for Horizon's use, first ensure it is blank. It's easiest to create a new database on your PostgreSQL server specifically for Horizon's use. We recommend creating a new user(role) in postgres dedicated for horizon's db and assigning that user(role) as the owner of the new horizon db. To illustrate this as example using `psql`, first login to databse server using the `psql` cli as a superuser, and then create the new user(role) and db for horizon:
+
+```
+postgres=#
```
urvisavla, line 126: When I was setting up PostgreSQL locally I found the `createuser` and `createdb` utilities that come bundled with the installation of Postgres easy to use. 
1. Create a new user called 'horizon' with a password
`createuser -P horizon`  
2. Create a new database called 'horizon' with the user 'horizon' as the owner
 `createdb -O horizon horizon`

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+| flag               | envvar           | example                              |
+| ------------------ | ---------------- | ------------------------------------ |
+| `--db-url`         | `DATABASE_URL`   | postgres://localhost/horizon_testnet |
+| `--network`        | `NETWORK`        | pubnet                               |
+| `--tx-sub-enabled` | `TX_SUB_ENABLED` | true                                 |
+| `--ingest`         | `INGEST`         | false                                |
+
+## Notes
+
+- In all configuration examples, the same parameter can be provided either as command line argument or an environment variable. If both are provided, the command line takes precedence.
```
urvisavla, line 80: This clarification is great! I was wondering what happens if I specify both the command line and environment variable when setting up horizon first time locally. Also, I was curious if there are any advantages to choose one approach over the other and may be that's the only one we mention in our documentation (?).

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+You run multiple instances of `stellar-horizon` each performs a subset of the roles.
+
+### Ingestion Role Instance
+
+You need to allocate one instance to perform ongoing ingestion to capture network activity. This will default to limiting storage of ingested network activity in the database to a sliding window of the last 30 days.
+
+** TODO - make --history-retention-count default to 518400 **
+
+** TODO - make --captive-core-use-db=true by default **
```
urvisavla, line 42: This is addressed in https://github.com/stellar/go/issues/4856

urvisavla (approved):
I've added a few minor comments but other than that it looks great!

docs/run-platform-server/configuring.mdx
sreuland: we suggest to use env variables across the board in 'passing config' section:

https://github.com/stellar/stellar-docs/pull/162/files#diff-a2be184a631b87a98857de0dd0abe901235c4639179be7c62719b5ab4e6502ddR100

I should add more context there to explain why, i.e. env variables are preferred all the time as they are more secure for obfuscating sensitive info that only the internal process can see, whereas command line leaves a trail that can be observed outside the process. It also depends on the install type, as far as what the operator/user interacts with, for bare metal, they define the env variables, but for containers, they use the container runtime mechanisms to pass env variables through such as docker/helm. 

@jacekn , in the case of Helm chart for Horizon, assuming install params would likely be passed as values file to `helm install ...` correct?

jacekn: @sreuland yeah that right. We tried to make the chart default work out of the box with ability to tune them.
The chart is in early stages though so it can absolutely be improvement and defaults tuned if they are not right. They were based on what we run at SDF but those might not necessarily be the best ones for the community.

docs/run-platform-server/configuring.mdx
sreuland: I'm not going to rock the boat on this, and follow existing precedence of referring to second person.

sreuland:
> ok to merge w/ TODOs? if yes I can ✔️ but @mollykarcher / Jesse (what's his GH username??) should look as well for product/partner perspective

yes, we are spinning up dev tickets for the TODO's, and will remove the TODO's from here first before merge.

mollykarcher (commented):
Took a first pass, overall I think this is a great start! 

I started to add a couple of these, but I decided to forgo any/all small formatting/grammar/nit-picking type of comments. I think we should probably hold these kinds of things until a final pass before we merge it all back into main. There'll just be too much noise with all the movement/changes if we attempt that before then.

mollykarcher:
> yes, we are spinning up dev tickets for the TODO's, and will remove the TODO's from here first before merge.

Yep 👍 aligned with this approach

docs/run-platform-server/configuring.mdx
urvisavla: Thanks for providing the reasoning. It makes sense and I agree with you in that we should opt for the lowest common denominator.

docs/run-platform-server/configuring.mdx
urvisavla: It looks like the binary name is "horizon" when downloaded as prebuilt binaries or built locally but it is "stellar-horizon" when installed via package manager. This distinction becomes relevant when users copy and paste example commands from our docs as it may not work correctly depending on the installation method. Previously the documentation had an additional step of renaming the binary using `sudo cp horizon /usr/local/bin/stellar-horizon` I'm uncertain about the reasoning behind the difference in names and the steps required to unify 
but mentioning it here as a potential TODO item (?).

docs/run-platform-server/configuring.mdx
sreuland: ok, @urvisavla , can we capture this as dev ticket, 'deprecate command line input of horizon flags' , with an outcome of warnings printed by horizon if any cli present?

I'll do pass through here to rename ref's to any cli format of flags, to the environment variable format.

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+## Multiple Instance Deployment
+
+You run multiple instances of `stellar-horizon` each performs a subset of the roles.
+
+### Ingestion Role Instance
+
+You need to allocate one instance to perform ongoing ingestion to capture network activity. This will default to limiting storage of ingested network activity in the database to a sliding window of the last 30 days.
+
+** TODO - make --history-retention-count default to 518400 **
```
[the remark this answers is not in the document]
sreuland: yes, on second thought, will include recommendation for setting `--history-retention-count=518400` explicitly in docs, as the minor gain of not setting one parameter seems outweighed by the potential confusion that would be caused otherwise.

docs/run-platform-server/configuring.mdx
sreuland: except for `--help` , we should allow just that one cli flag

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+- `--network`, optional, can be one of Stellar's public networks, 'pubnet', or 'devnet'. Triggers horizon to automatically set configurations for remaining horizon settings and generate the correct core toml/cfg settings. If you only need Horizon to connect to one of those public Stellar networks, this will take care of all related configuration.
+
+**TODO - create the new `--network` flag and the functionality it performs to configure all aspects of network connection, including toml, passphrase, archive urls, etc for networks 'pubnet' and 'testnet'. Refer to quickstart for example of same consolidated config parameter for setting network. **
+
+- If you want to connect Horizon to a different stellar network other than pubnet or testnet or override any of the defaults that `--network` usage will initiate, the key configuration flags that can be set are: `--history-archive-urls`, `--captive-core-config-path`, `--network-pass-phrase`, `--captive-core-storage-path`, `--stellar-core-url`.
```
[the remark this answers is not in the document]
sreuland: good point, swept that up - https://github.com/stellar/stellar-docs/pull/162/commits/a0f5d125ee71ece0e21d48dc0517dae3b34f3b13

docs/run-platform-server/index.mdx
```diff
@@ -3,16 +3,12 @@ title: "Overview"
...
 
 Before we begin, it's worth reiterating the sentiment echoed in the [Run a Core Node](../run-core-node) guide: **we do not endorse running Horizon backed by a standalone Stellar Core instance**, and especially not by a _validating_ Stellar Core. These are two separate concerns, and decoupling them is important for both reliability and performance. Horizon instead manages its own, pared-down version of Stellar Core optimized for its own subset of needs (we'll refer to this as a "Captive Core" instance).
 
-## Upgrading From Horizon 1.x
-
-If you're coming from an existing deployment of Horizon, you're probably running it alongside a standalone "Watcher" Stellar Core node. As noted above, this architecture is now **deprecated**, and support for it will be dropped in Horizon 3.x. The [Migration](./migrating.mdx) guide should facilitate the process to moving to the Captive Core architecture and get you up to speed.
-
 ## Why Run Horizon?
```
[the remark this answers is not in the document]
sreuland: ok, I can include more on that with refs to Hubble/DW and blog link for use cases beyond 30 days in upcoming changes on the ingestion.mdx for https://github.com/stellar/stellar-docs/issues/147

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+| `--db-url`         | `DATABASE_URL`   | postgres://localhost/horizon_testnet |
+| `--network`        | `NETWORK`        | pubnet                               |
+| `--tx-sub-enabled` | `TX_SUB_ENABLED` | true                                 |
+| `--ingest`         | `INGEST`         | false                                |
+
+## Notes
+
+- In all configuration examples, the same parameter can be provided either as command line argument or an environment variable. If both are provided, the command line takes precedence.
+
+- If you have configured your deployment to perform the ingestion role, then it is **strongly** recommended to review [Ingestion](./ingestion.mdx) first and [Filtering](./ingestion-filtering) second and factor that into configuration parameters to achieve best performance related to your application requirements before proceeding further.
```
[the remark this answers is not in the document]
sreuland: yes, needed to add a Notes bullet for `TX_SUB_ENABLED` in general, and reference to channel accounts usage. https://github.com/stellar/stellar-docs/pull/162/commits/8ba722ce02749aa17849e3bf3a5c1f342a0ebf3a

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+  - When enabled, requires **live** ingestion process to be running on the same database because Horizon depends on new ledgers from network to confirm a transaction submission status, Horizon will report startup error if it detects no **live** ingestion. Requires `NETWORK` or `STELLAR_CORE_URL` be defined for access to network.
+  - When disabled with FALSE, Horizon will return 405 on POSTs to /tx.
+
+- `NETWORK`, optional, can be one of Stellar's public networks, 'pubnet', or 'devnet'. Triggers horizon to automatically set configurations for remaining horizon settings and generate the correct core toml/cfg settings. If you only need Horizon to connect to one of those public Stellar networks, this will take care of all related configuration.
+
+**TODO - create the new `NETWORK` env variable and the functionality it performs to configure all aspects of network connection, including toml, passphrase, archive urls, etc for networks 'pubnet' and 'testnet'. Refer to quickstart for example of same consolidated config parameter for setting network. **
```
urvisavla, line 86: I'm creating a dev ticket for this task and wanted to clarify a few points
1. If "NETWORK" is not specified, Horizon will still automatically set the values for CAPTIVE_CORE_CONFIG_PATH and CAPTIVE_CORE_STORAGE_PATH.

2. The CAPTIVE_CORE_STORAGE_PATH is not dependent on the network chosen. So the default storage path will remain the same regardless of the network.

sreuland: > If "NETWORK" is not specified, Horizon will still automatically set the values for CAPTIVE_CORE_CONFIG_PATH and CAPTIVE_CORE_STORAGE_PATH.

Do you see that behavior based on `NETWORK-PASSPHRASE`, I found an auto network config feature done in past based on that value, [go#3783](https://github.com/stellar/go/pull/3783), was not aware of this! I think the intent for its usage was more oriented to ease for dev, the design discussion struggled with prod usage, which we should learn from. 

@mollykarcher , I wonder how much this auto-config default behavior of `NETWORK-PASSPHRASE` is being relied upon in the field? It's not mentioned in the -help output, seems kind of quietly in there. Should we focus on shoring up the existing auto-config functionality of `NETWORK-PASSPHRASE` and referring to that usage here in docs or move forward with `NETWORK`? I kinda of prefer the latter as the former seems to conflate the meaning of what appears to be just a value-based setting, but need a check on this.

In either case for auto-config, I don't think Horizon should bother to configure default network or captive core settings if the key network settings are absent, as there is a valid configuration for the API role, where INGEST=false, TX_SUB_ENABLED=false and no network access is required, only the database is used. 

https://github.com/stellar/stellar-docs/pull/162/files/8ba722ce02749aa17849e3bf3a5c1f342a0ebf3a#diff-a2be184a631b87a98857de0dd0abe901235c4639179be7c62719b5ab4e6502ddR51

@urvisavla , this touches on need for new validations on config param settings that we should capture in the ticket also which we propose Horizon to detect/warn about, i.e. if INGEST=true or TX_SUB_ENABLED=true then network config settings must be present, either the key auto-config parameter or the collection of lower level params.

docs/run-platform-server/configuring.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+Pre-requisites:
+
+- [Installation](./installing.mdx) is completed on a host machine and results in two executables: `stellar-horizon` and `stellar-core`. Per the installation instructions, `stellar-core` should be included in your host OS' `$PATH`, as the configurations below result in Horizon searching for a `stellar-core` to execute.
+- [Initialize Database](#initialize-horizon-database)
+
+You are now ready to setup configuration parameters to perform three important, distinct roles:
+
+- **serving requests** via a regular web-based HTTP API, and
+- **ingesting ledgers** from core nodes of a Stellar network to keep its world-view up to date, and
```
sreuland, line 17: @mollykarcher , @urvisavla , @Shaptic , wondering about a possible fourth role..a 'current state' role, a horizon instance that runs ingestion w/captive core, but stores no history, just the current state. provides cohort of api endpoints that come with that data only such as order book, offers, trades(trade-agg?), etc, not sure if enough app uses cases fall into this.

the role would equate to lighter-weight deployment, likely not external postgres/db, maybe get by with transient sqlite db spun up internally on disk/mem, this may have been evaluated already.

urvisavla: >a horizon instance that runs ingestion w/captive core, but stores no history, just the current state.

Would this be same as running ingestion with no history retention? I just want to clarify my understanding of running ingestion with no history.

PR #167, opened 2023-06-23
Add Hubble Docs

sydneynotthecity:
Hubble does not have any public docs, and this update adds a new section under Docs called `Accessing Historical Data`. We focused on the following topics: 

* Overview - provide an overview of what Hubble is, how it differs from Horizon and when you should use it
* Connecting - details 3 different connection methods to BigQuery (UI, SDKs and viz tool)
* Viewing Metadata - shows end users how to discover information about Hubble through BigQuery Metadata

Future PRs will add more technical docs and examples, which we plan to put under `APIs`. This will include an entity relationship diagram, data dictionaries and detailed examples.

PR #169, opened 2023-06-30
Re-organizing API documentation collections

ElliotFriend:
These changes make it a bit more manageable to collect our various API specifications into a single navbar item in our docusaurus site. With the upcoming addition of the Hubble information (#167), this will ease the implementation, and make things a bit more uniform.

- I've included some nginx rewrites that I _think_ will do the correct redirections from any old URLs to the new ones.
- There is now a single `API` dropdown menu item in the site's navbar, instead of having separate navbar links for all the various APIs, we can now collect them under a single button.
- The `DocSidebar` component has been swizzled to allow us to keep the "category" of API used in the breadcrumbs, but to not clutter the sidebar with extra (and unnecessary) links.


ElliotFriend:
Good grief that's a lot of files changed!! **290!!** 😮 

The main changes are these:

- Moved horizon's _input_ `/openapi` directory into `/openapi/horizon`
- Moved horizon's _output_ `/api` directory into `/api/horizon`
- Moved AP's _output_ `/ap_api` directory into `/api/anchor-platform`
- Modified the navbar to only present a single dropdown list containing each API documentation set, rather than individual items all across the top.
- Changed the openapi plugin configuration to make the `/api/*` URL behave as a single instance of the `plugin-content-docs` plugin, instead of using a separate instance for each API
- Customized the sidebar behavior on `/api` URLs so that the "category" of API (horizon, anchor platform) doesn't appear unnecessarily in the sidebar but does still show in the breadcrumbs
- Attempted some nginx redirects from the old output URLs to the new ones.

JakeUrban:
I would recommend "APIs" instead of "API" for the drop-down label, but otherwise LGTM, thanks for taking this on!

tyvdh:
Ah I see https://github.com/stellar/stellar-docs/pull/169/commits/c6b3f831620f41cf137e46a32782dabe40bcc146 now. So maybe this has already been dealt with?

ElliotFriend:
> Ah I see `c6b3f83` now. So maybe this has already been dealt with?

@tyvdh Yes, I believe those redirects will handle everything. As a means of testing them, [I put together this script](https://gist.github.com/ElliotFriend/aa5e35d370322d9826357db17730a530). Here's what it does:

1. downloads and parses the `sitemap.xml` for the production docs site.
2. for each of the URLs, if it includes `/api` or `/ap_api`, it changes the domain to the PR preview domain
3. it requests that preview url, searches through the response and makes sure that an expected redirect took place
4. it outputs a bunch of crap that might not be useful, but can help to context-ify what's going on.

The only one that didn't respond as "expected" was the `/api` page, which is a new one that is an "APIs Overview" page.

**TL;DR** - I checked all the redirects in the sitemap. Things redirect as I was expecting 👍🏻

tyvdh:
MERGE IT BABY!!!

tyvdh (approved):

PR #177, opened 2023-07-11
add TS examples for equivalent SDK methods

piyalbasu:


docs/building-apps/wallet/sep24.mdx
```diff
@@ -143,6 +189,18 @@ suspend fun depositDifferentAccount(): InteractiveFlowResponse {
...
 ```
 
+```typescript
+const recipientAccount = "G...";
+const depositDifferentAccount = async (): Promise<InteractiveFlowResponse> => {
+  return anchor.interactive().deposit({
+    accountAddress: accountKp.publicKey(),
+    assetCode,
+    authToken,
+    fundsAccountAddress: recipientAccount
```
piyalbasu, line 199: I believe `fundsAccountAddress` is the equivalent key for `destinationAccount` above, but lmk if I'm wrong

Ifropc: I think so. It should be optional, is it like this right now? (If not, making it optional is part of WAL-889 IIRC)

docs/building-apps/wallet/intro.mdx
```diff
@@ -60,6 +66,19 @@ val walletCustom = Wallet(
 )
 ```
 
+```typescript
+const customClient: AxiosInstance = axios.create({
+  baseURL: "https://some-url.com/api",
```
Ifropc, line 71: Why do we even need to set baseUrl? I don't think it's good to show it in the example

docs/building-apps/wallet/intro.mdx
```diff
@@ -60,6 +66,19 @@ val walletCustom = Wallet(
...
+  timeout: 1000,
+  headers: { "X-Custom-Header": "foobar" },
+});
+let appConfig = new ApplicationConfiguration(DefaultSigner, customClient);
```
Ifropc, line 75: Shouldn't DefaultSigner be there by default? (I.e. you don't need to set it manually)

docs/building-apps/wallet/intro.mdx
```diff
@@ -60,6 +66,19 @@ val walletCustom = Wallet(
 )
 ```
 
+```typescript
```
Ifropc, line 69: I think all this code belongs to the "Configuring the Client" section

docs/building-apps/wallet/sep24.mdx
```diff
@@ -28,6 +28,10 @@ Let's start with getting an instance of `Interactive` class, responsible for all
 val sep24 = anchor.sep24()
 ```
 
+```typescript
+let sep24 = await anchor.getInfo();
```
Ifropc, line 32: `anchor.interactive()` ?

docs/building-apps/wallet/sep24.mdx
piyalbasu: `fundsAccountAddress` is optional. It gets set to `accountAddress` if it isn't explicitly set and then sent to the `/interactive` endpoint as `account`

docs/building-apps/wallet/sep24.mdx
```diff
@@ -143,6 +189,18 @@ suspend fun depositDifferentAccount(): InteractiveFlowResponse {
 }
 ```
 
+```typescript
+const recipientAccount = "G...";
```
Ifropc, line 193: Can you add TODO here to modify code when memo is supported?

docs/building-apps/wallet/sep24.mdx
```diff
@@ -173,6 +241,19 @@ val watcher = sep24.watcher()
 val result = watcher.watchOneTransaction(token, "transaction id")
 ```
 
+```typescript
+let watcher = anchor.watcher();
+
+let result = watcher.watchOneTransaction({
+  authToken,
+  assetCode,
```
Ifropc, line 249: Why do we need assetCode when watching one transaction?

docs/building-apps/wallet/sep24.mdx
```diff
@@ -236,6 +324,13 @@ It's also possible to fetch transaction by the asset
 val transactions = sep24.getTransactionsForAsset(asset, token)
 ```
 
+```typescript
+const transactions = await anchor.getHistory({
```
Ifropc, line 328: `getHistory` should be deprecated... It has some extra filtering logic that shouldn't be there.
`getTransactionsForAsset` is more flexible than `getHistory`

Ifropc (approved):

docs/building-apps/wallet/sep24.mdx
Ifropc: Got it. Small thing: for withdrawal it should be `originAccount`

docs/building-apps/wallet/intro.mdx
piyalbasu: You don't necessarily, it's just an example of the params `axios.create` takes. I can rm

docs/building-apps/wallet/intro.mdx
piyalbasu: You don't, I can just pass `null` here

Ifropc: Can you just omit it or it's not possible?

docs/building-apps/wallet/sep24.mdx
Ifropc: Ok figure it out later, as I don't think it should be necessary

acharb: @CassioMG is more familiar with this code than I am. But from what I understand we're using the asset code as the key for the registry to keep track of transactions: https://github.com/stellar/typescript-wallet-sdk/blob/b0a1ff12c2531c9526ebc59bfbed285f1db04d5d/src/walletSdk/Watcher/index.ts#L100-L113

PR #183, opened 2023-07-14
Add init.sql

Ifropc:


docs/anchoring-assets/anchor-platform/getting_started.mdx
```diff
@@ -237,6 +237,8 @@ services:
       - "5432:5432"
     env_file:
       - ./dev.env
+    volumes:
+      - ./init.sql:/docker-entrypoint-initdb.d/init.sql
```
JakeUrban, line 241: `CREATE DATABASE <name>` will fail if the DB already exists. I just want to confirm that the docker container won't crash if the init script isn't successful.

JakeUrban (approved):

docs/anchoring-assets/anchor-platform/getting_started.mdx
Ifropc: Only running once
https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L312-L338

PR #184, opened 2023-07-17
[ANCHOR-339] Add the /event endpoint to Callback API.yml

lijamie98:
This is for the implemenation of [ANCHOR-339](https://stellarorg.atlassian.net/browse/ANCHOR-339)

[ANCHOR-339]: https://stellarorg.atlassian.net/browse/ANCHOR-339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openapi/anchor-platform/Callbacks API.yml
+++ b/openapi/anchor-platform/Callbacks API.yml
@@ -466,7 +466,7 @@ paths:
             application/json:
               schema:
                 $ref: './schemas.yml#/components/schemas/Error'
-  /event:
+  /events:
     post:
       tags:
         - Events
--- a/openapi/anchor-platform/Callbacks API.yml
+++ b/openapi/anchor-platform/Callbacks API.yml
@@ -466,7 +466,7 @@ paths:
             application/json:
               schema:
                 $ref: './schemas.yml#/components/schemas/Error'
-  /events:
+  /event:
     post:
       tags:
         - Events
@@ -485,7 +485,7 @@ paths:
                 timestamp:
                     type: string
                     description: |
-                      `timestamp` the current Unix timestamp (number of seconds since epoch) at the time the callback is sent.
+                      `timestamp` the current Unix timestamp (number of seconds since epoch) at the time event is sent.
                 payload:
                   $ref: './schemas.yml#/components/schemas/Event'
       responses:
--- a/openapi/anchor-platform/Callbacks API.yml
+++ b/openapi/anchor-platform/Callbacks API.yml
@@ -466,3 +466,34 @@ paths:
             application/json:
               schema:
                 $ref: './schemas.yml#/components/schemas/Error'
+  /event:
+    post:
+      tags:
+        - Events
+        - SEP-24
+        - SEP-31
+      summary: Receive a event from the anchor platform.
+      operationId: postEvent
+      description: |
+        Receive a JSON object representing an event.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                timestamp:
+                    type: string
+                    description: |
+                      `timestamp` the current Unix timestamp (number of seconds since epoch) at the time the callback is sent.
+                payload:
+                  $ref: './schemas.yml#/components/schemas/Event'
+      responses:
+        '200':
+          description: The event is successfully processed and ready to receive next event. The response body is empty.
+        '400':
+          description: The event is invalid or rejected. The response body contains the error message.
+
+
+
+


openapi/anchor-platform/Callbacks API.yml
```diff
@@ -466,3 +466,34 @@ paths:
             application/json:
               schema:
                 $ref: './schemas.yml#/components/schemas/Error'
+  /events:
```
JakeUrban, line 469: I think we have singular names for the rest of our endpoints in this API right? I agree plural is the standard but we might want to stay consistent.

openapi/anchor-platform/Callbacks API.yml
```diff
@@ -466,3 +466,34 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              properties:
+                timestamp:
+                    type: string
+                    description: |
+                      `timestamp` the current Unix timestamp (number of seconds since epoch) at the time the callback is sent.
+                payload:
+                  $ref: './schemas.yml#/components/schemas/Event'
```
JakeUrban, line 490: Is now a good time to revisit the schema?

- Is the `id` attribute specific to the event or transaction?
- Why do we have `sep` as a top level attribute when it is available inside the `transaction` object?
- Why do we have a `transaction_error` type? Can't we merge this with `transaction_status_changed`?

JakeUrban: Also, do we need a timestamp on each event to ensure business can check if they've already processed an event after the one received? Is that possible?

openapi/anchor-platform/Callbacks API.yml
```diff
@@ -466,3 +466,34 @@ paths:
...
+      responses:
+        '200':
+          description: The event is successfully processed and ready to receive next event. The response body is empty.
+        '400':
+          description: The event is invalid or rejected. The response body contains the error message.
```
JakeUrban, line 495: I'm not sure we want to define any non-200 response code. All responses should have 200, and all other response codes will be interpreted as an error and retries will be made according to the AP's strategy.

openapi/anchor-platform/Callbacks API.yml
lijamie98: Changed to singular.

openapi/anchor-platform/Callbacks API.yml
lijamie98: The `Event` schema can be further improved. I think we have a couple of days to review. 

For `id`, `sep`, and the transaction status, I am open to redesign if that's also ok with you.

And we should have a `created_at` in the `Event` schema.

openapi/anchor-platform/Callbacks API.yml
lijamie98: The response code is limited `400`. If the business server returns anything other than 400, they will be considered the same way as `400`. 

The retries can be handled in the next fix version.

openapi/anchor-platform/Callbacks API.yml
JakeUrban: Sounds good to me.

openapi/anchor-platform/Callbacks API.yml
JakeUrban: But 400 itself is not a valid status, so I don't think we should define it in the spec.

JakeUrban (approved):
Approved with the small comment about 400's

openapi/anchor-platform/Callbacks API.yml
```diff
@@ -466,3 +466,34 @@ paths:
...
+      operationId: postEvent
+      description: |
+        Receive a JSON object representing an event.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                timestamp:
```
Ifropc, line 485: Why is timestamp not a part of an event?

openapi/anchor-platform/Callbacks API.yml
```diff
@@ -466,3 +466,34 @@ paths:
             application/json:
               schema:
                 $ref: './schemas.yml#/components/schemas/Error'
+  /event:
+    post:
+      tags:
+        - Events
+        - SEP-24
+        - SEP-31
+      summary: Receive a event from the anchor platform.
```
Ifropc, line 475: [an event]

openapi/anchor-platform/Callbacks API.yml
Ifropc: If we redesigning event schema in the separate PR this change LGTM

Ifropc (approved):

JakeUrban:
closing this because these changes have been pulled into https://github.com/stellar/stellar-docs/pull/232

PR #188, opened 2023-07-19
Add SDP API Docs

JakeUrban:
Adds the SDP API docs to the APIs section of the developer docs.

When reviewing changes, use the PR preview. The only file worth reviewing, and the only file that needs updates, is `openapi/sdp/main.yml`.

The OpenAPI specification was mostly sufficient for generating good docs, however there are a few remaining items:

- The specification used example requests & responses instead of defining schemas for each of the requests and responses.   Our docs do not support examples, so each endpoint is missing POST or PATCH schemas and response schemas.
- Each "resource" is missing a description
- Each endpoint is missing a description. Right now they match the status code they return like OK or Created

NOTE: DO NOT CHANGE THE SUMMARY OR OPERATION ID OF THE ENDPOINTS

These fields in the specification map to the file names of the autogenerated markdown files. If you change the summary or operation Id of an endpoint, you'll have to update the relevant endpoint table to link to the updated file name.


torisamples:
Thank you! I can take a first pass at writing the resource descriptions today and then work on the endpoint descriptions.

ElliotFriend:
This is an awesome addition @JakeUrban! One observation:

You'll need to modify one of the `rewrite` rules in `/nginx/includes/redirects.conf`. As it is now, someone can find themselves redirected to `docs.site/api/horizon/sdp/whatever` on accident. It only seems to happen when you physically type the URL into the address bar or copy/paste into the address bar. Example: go to [this URL](https://developers-pr188.previews.kube001.services.stellar-ops.com/api/sdp/resources/retrieve-a-receiver). I'm not _exactly_ sure why the redirect doesn't happen through ordinary page navigation, though. Probly some client-side wizardry with docusaurus, maybe? I can add that redirect for you, if you like.

Two (pretty inconsequential) questions:

1. The `api:bundle` command in `package.json` is getting a bit cumbersome. Do you mind if I push a change to your branch that cleans that up just a bit? I'm thinking something like breaking horizon, ap, sdp into their own commands and `api:bundle` calling each of them.
2. The URL schemes for SDP and AP are inconsistent. We already have `docs.site/api/anchor-platform`, but we're adding `docs.site/api/sdp`. We should probably be consistent across our different APIs, although I don't really have a preference for which way we might go. I think `/api/stellar-disbursement-platform` feels quite unwieldy, though lol. This is probably a pretty nit-picky thing anyway, though. What are your thoughts?

JakeUrban:
Hi @ElliotFriend, it would be great if you could add the redirect!

As for 1, I think thats a good idea, but maybe outside the scope of this PR. If you make another PR for this refactor I can merge main to update this branch.

On 2, I definitely don't want to use `api/ap`, because nobody calls the anchor platform "AP" in the same way we use "SDP". So if we want to be consistent, I would vote for `api/stellar-disbursement-platform` so we can keep `api/anchor-platform`. Personally I don't think its a big deal if the URLs don't all use full names though.

briwylde08:
I spelled disbursement wrong. I blame my concussion. But you get the idea.

marwen-abid (approved):
LGTM 🚀 🚀 🚀

PR #192, opened 2023-07-20
Breaking up the `api:bundle` yarn command for better maintainability

ElliotFriend:
Discussed in #188 

The `api:bundle` command is getting a bit cumbersome, and with the new additions of SDP and hubble, it will only become more-so. This change breaks them into unique commands `bundle-horizon` and `bundle-anchor-platform` which are then called inside of the original `api:bundle` command.


JakeUrban (approved):
LGTM!

PR #194, opened 2023-07-24
SDP: add guidance for funding distribution account

JakeUrban:
Adds a new section to the "Getting Started Guide" for funding a distribution account with testnet USDC.


torisamples (approved):
Looks good!

PR #195, opened 2023-07-25
feat(ledgers): explain this is called a "block" in other blockchains

chadoh:
I found Stellar's non-standard use of the word "ledger" pretty surprising. I think it's useful to explain the concept in terms that other blockchain users might be familiar with.


briwylde08:
Been thinking about this one. I'm interpreting your addition as... other blockchains refer to state (what Stellar calls a ledger) as a block. Is this what you're saying? I think instead that a block in other networks refers to what Stellar calls a transaction set. Thoughts?

briwylde08 (approved):

chadoh:
I'm used to using phrases like:

- "the state of the blockchain at a given block height"
- "the state of the blockchain at a given block"
- "the state of the ledger at a given block/height"

Like, "blockchain" and "ledger" are synonyms, and "block"/"height"/"block height" are all synonyms. 

I'm also used to saying things like "wait 3 blocks for finality," not "wait 3 _ledgers_". 

Stellar makes "ledger" a synonym for "block", which I find v surprising.

PR #216, opened 2023-08-24
added reference to helm chart usage on core,horizon install

sreuland:
Suggest the new helm chart option available for installing horizon and core.

### DIFF
--- a/docs/run-api-server/installing.mdx
+++ b/docs/run-api-server/installing.mdx
@@ -10,8 +10,9 @@ To install Horizon, you have a few choices. You can...
 - install prebuilt binaries [from our repositories](#package-manager) via your package manager if running a Debian-based system,
 - download a [prebuilt release](https://github.com/stellar/go/releases/latest) of Horizon for your target architecture and operation system, or
 - [build Horizon and Stellar Core yourself](#building) from scratch.
+- use [horizon helm chart](https://github.com/stellar/helm-charts/blob/main/charts/horizon/README.md) which internally uses same prebuilt binary in the runtime image.
 
-**The first method is recommended**: Not only do you ensure OS compatibility and dependency management, you'll also install some convenient wrappers that make running Horizon and Stellar Core in their respective environments much simpler.
+**prebuilt binaries method is recommended**: Not only do you ensure OS compatibility and dependency management, you'll also install some convenient wrappers that make running Horizon and Stellar Core in their respective environments much simpler.
 
 ## Installation Methods
 
@@ -24,6 +25,21 @@ SDF publishes new releases to its custom Ubuntu repositories. Follow [this guide
 ```bash
 sudo apt update
 sudo apt install stellar-horizon stellar-core
+```
+
+</CodeExample>
+
+## Helm Chart Installation
+
+If the deployment can be done on Kubernetes, there is a [horizon helm chart](https://github.com/stellar/helm-charts/blob/main/charts/horizon) available.
+
+If kubernetes is not an option, the manifests in those charts may still be good reference for showing how to configure and run horizon as docker container. Just run the helm command with `template` to print the container config to screen:
+
+<CodeExample>
+
+```bash
+$ git clone https://github.com/stellar/helm-charts; cd helm-charts
+$ helm template -f charts/horizon/values.yaml charts/horizon/
 ```
 
 </CodeExample>
--- a/docs/run-core-node/installation.mdx
+++ b/docs/run-core-node/installation.mdx
@@ -3,7 +3,7 @@ title: Installing
 sidebar_position: 30
 ---
 
-There are three ways to install Stellar Core: you can use a Docker image, use pre-built packages, or build from source. Using a Docker image is the quickest and easiest method, so it's a good choice for a lot of developers, but if you're running Stellar Core in production you may need to use packages or, if you want to sacrifice convenience for maximum control, build from source. Whichever method you use, you should make sure to install the latest [release](https://github.com/stellar/stellar-core/releases) since releases are cumulative and backwards compatible.
+There are multiple ways to install Stellar Core: you can use a Docker image, pre-built packages, helm-chart, or build from source. Using the Docker image is the quickest and easiest method, so it's a good choice for a lot of developers, but if you're running Stellar Core in production you should consider the helm chart or packages, if you want to sacrifice convenience for maximum control, build from source. Whichever method you use, you should make sure to install the latest [release](https://github.com/stellar/stellar-core/releases) since releases are cumulative and backwards compatible.
 
 ## Docker-based installation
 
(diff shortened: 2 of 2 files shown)


docs/run-api-server/installing.mdx
```diff
@@ -28,6 +29,40 @@ sudo apt install stellar-horizon stellar-core
...
 </CodeExample>
 
+## Helm Chart Installation
+
+If the deployment can be done on Kubernetes, there is a [horizon helm chart](https://github.com/stellar/helm-charts/blob/main/charts/horizon) available.
+
+<CodeExample>
+
+```bash
+$ git clone https://github.com/stellar/helm-charts; cd helm-charts
```
jacekn, line 39: Maybe we can install straight from the helm repo? Something like this might work (untested):
```
helm repo add stellar https://helm.stellar.org/charts && helm repo update stellar
helm install my-horizon stellar/horizon --namespace my-horizon-namespace-on-cluster --set global.image.horizon.tag=XXX 
```

sreuland: ok, I'll test the chart deploys for horizon/core direct from repo on sandbox. For setting values on helm, which do you think we should recommend for admins to use from helm cli, get a local copy of the `values.yml` from the charts repo, or recommend `--set`s on specific keys here? I was leaning towards values.yml, since it scales a bit better over time, as it won't become stale, it could be linked to the values.yml in charts repo, rather than documenting  multiple `--set`s here which could require maintenance to keep up to date.

sreuland: @jacekn , found an issue during helm install from repo on horizon chart - https://github.com/stellar/helm-charts/issues/64

sreuland:
keeping this in draft until helm chart behavior verified.

docs/run-api-server/installing.mdx
jacekn: > For setting values on helm, which do you think we should recommend for admins to use from helm cli, get a local copy of the values.yml from the charts repo, or recommend --sets on specific keys here? 

If all we need to set is the image tag I'd say `--set` is nicer because it's short and concise. Maybe we can use that and mention the values method for people who need to customize things?

sreuland: it might suffice to use just `--set` for most of use cases, and include one more for `global.network`, I'll mention to checkout out values.yml from repo for any non standard use case, i.e. if the user wants to use any network other than testnet/pubnet, then they have to manually set many other attributes in values.yml.

docs/run-core-node/installation.mdx
```diff
@@ -3,7 +3,7 @@ title: Installing
 sidebar_position: 30
 ---
 
-There are three ways to install Stellar Core: you can use a Docker image, use pre-built packages, or build from source. Using a Docker image is the quickest and easiest method, so it's a good choice for a lot of developers, but if you're running Stellar Core in production you may need to use packages or, if you want to sacrifice convenience for maximum control, build from source. Whichever method you use, you should make sure to install the latest [release](https://github.com/stellar/stellar-core/releases) since releases are cumulative and backwards compatible.
+There are multiple ways to install Stellar Core: you can use a Docker image, pre-built packages, helm-chart, or build from source. Using the Docker image is the quickest and easiest method, so it's a good choice for a lot of developers, but if you're running Stellar Core in production you should consider the helm chart or packages, if you want to sacrifice convenience for maximum control, build from source. Whichever method you use, you should make sure to install the latest [release](https://github.com/stellar/stellar-core/releases) since releases are cumulative and backwards compatible.
```
tsachiherman, line 6: This paragraph list few ways to install Stellar Core. The "grocery list" here would ideally match, in both name & order the subsequent topic titles. Could you align these ?

docs/run-api-server/installing.mdx
```diff
@@ -10,8 +10,9 @@ To install Horizon, you have a few choices. You can...
 - install prebuilt binaries [from our repositories](#package-manager) via your package manager if running a Debian-based system,
 - download a [prebuilt release](https://github.com/stellar/go/releases/latest) of Horizon for your target architecture and operation system, or
 - [build Horizon and Stellar Core yourself](#building) from scratch.
+- use [horizon helm chart](https://github.com/stellar/helm-charts/blob/main/charts/horizon/README.md) which internally uses same prebuilt binary in the runtime image.
 
-**The first method is recommended**: Not only do you ensure OS compatibility and dependency management, you'll also install some convenient wrappers that make running Horizon and Stellar Core in their respective environments much simpler.
+**prebuilt binaries method is recommended**: Not only do you ensure OS compatibility and dependency management, you'll also install some convenient wrappers that make running Horizon and Stellar Core in their respective environments much simpler.
```
jacekn, line 15: I'm not sure how best to phrase this but another big win with packages is cryptographic verification of the packages and updates. apt has built-in checks to ensure package were not tampered with so users can guarantee that updates come from SDF

docs/run-api-server/running.mdx
```diff
@@ -5,7 +5,68 @@ sidebar_position: 40
...
+--namespace my-namespace  \
+--set ingest.persistence.enabled=true \
+--set web.enabled=true \
+--set web.existingSecret=my-db-secret \
+--set global.image.horizon.tag=2.26.1 \
+--set global.network=testnet \
+--set ingest.existingSecret=my-db-secret \
+--set ingest.horizonConfig.captiveCoreUseDb=true \
+--set ingest.resources.limits.cpu=2 \
+--set ingest.resources.limits.memory=6Gi \
```
jacekn, line 46: It might be OK to skip the limits. Many clusters won't have any resource limits and if resources are not specified pods will assume they can burst without any restrictions

sreuland: Ok, I will add some additional statement for users to be aware of and confirm quotas on cluster/namespace and to use these if unsure in any case. As I spent a couple hours debugging failed deployments due to captive core crashes in the pod, and figured out it was due to OOM, and that was after forcing captive core to use db instead of in-memory as well.

docs/run-core-node/installation.mdx
sreuland: re-aligned the text here for drill-down into matching sections - https://github.com/stellar/stellar-docs/pull/216/commits/4cbc536dc93f26d00ee1a9871d9a19e623a11f24

docs/run-api-server/installing.mdx
sreuland: added mention of these aspects - https://github.com/stellar/stellar-docs/pull/216/files#diff-cf952d95be5052ac3eb42baa625826c3ebf71d59b3d1c85274606eba312e9d48R15

sreuland:
@jacekn , I have addressed feedback on helm usages, at this point the remaining aspect is the core helm docs have a dependency on values.yaml's for testnet/pubnet that haven't been created yet:

https://github.com/stellar/stellar-docs/pull/216/files#diff-e0a15f4578c42a6d9909a0ed0007c7ced8c399449f8acc8d87d9a6d10dcbd5a2R40-R42

Once https://github.com/stellar/helm-charts/issues/65 is done, I think this PR will be ready to merge.

briwylde08:
Hi @sreuland! Just checking on this one again. Is this still a WIP?

PR #223, opened 2023-09-05
Partner Experience: update the helm horizon working example

sreuland:
Updated the horizon helm chart example to also show how to create db credentials secret which is needed on cluster for helm chart to reference for db connectxivity.

### DIFF
--- a/docs/run-platform-server/running.mdx
+++ b/docs/run-platform-server/running.mdx
@@ -15,7 +15,7 @@ If you chose containerized installation, you don't execute the horizon binary di
 
 - To run on Kubernetes using helm chart, ensure you have followed the [pre-requisite](./installing.mdx#helm-chart-installation) of installing the Helm cli tool and adding the Stellar chart repo to helm client:
 
-  The horizon process [requires access to a Postgres 12 database](./configuring.mdx#preparing-the-database), first use the common kubernetes cli tool `kubectl` from your workstation to create a kubernetes secret on the intended namespace of the kubernetes cluster which will hold the horizon database url.
+  The horizon process [requires access to a Postgres 12 database](./configuring.mdx#preparing-the-database). First use the common kubernetes cli tool `kubectl` from your workstation to create a kubernetes secret on the intended namespace of the kubernetes cluster which will hold the horizon database url.
 
   <CodeExample>
 
--- a/docs/run-platform-server/running.mdx
+++ b/docs/run-platform-server/running.mdx
@@ -74,7 +74,7 @@ If you chose containerized installation, you don't execute the horizon binary di
 
 <br />
 
-Once horizon process starts it will emit logging to standard out. When run, you should see output similar to:
+Once the horizon process starts it will emit logging to standard out. When run, you should see output similar to:
 
 <CodeExample>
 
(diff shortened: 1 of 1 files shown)


docs/run-platform-server/running.mdx
```diff
@@ -13,36 +13,68 @@ If you chose containerized installation, you don't execute the horizon binary di
...
+    helm install myhorizon stellar/horizon \
+    --namespace my-horizon-namespace-on-cluster \
+    --values values.yaml
+    ```
+
     </CodeExample>
 
+  - specific to customizing network configuration parameters, if you want to connect to a network other than presets of `testnet` or `pubnet`, then you won't use `global.network`, instead, use local copy of [values.yaml](https://github.com/stellar/helm-charts/blob/main/charts/horizon/values.yaml) and set `ingest.coreConfig`, and refer to [\_core-config.tpl](https://github.com/stellar/helm-charts/blob/main/charts/horizon/templates/_core-config.tpl) for example of all the key/value pairs to include.
+
+  - minimum resource limits, verify whether `LimitRange` defaults are defined on the target namespace in kubernetes for deployment, if so, ensure that the defaults provide at least minimum resource limits of `6Gi` of memory and `1` cpu. Otherwise, define the limits explicitly on the helm install via the `ingest.resources.limits.*` shown in example, to ensure the deployed pods have adequate resources.
```
aditya1702, line 73: Is the "minimum resource limits" sort of a heading here? If yes, maybe we can write it as - "Minimum Resource Limits:"

docs/run-platform-server/running.mdx
```diff
@@ -13,36 +13,68 @@ If you chose containerized installation, you don't execute the horizon binary di
...
+
 <br />
-Once horizon process starts it will emit logging to standard out. When run, you should
-see output similar to:
+
+Once horizon process starts it will emit logging to standard out. When run, you should see output similar to:
```
aditya1702, line 77:
```suggestion
Once the horizon process starts it will emit logging to standard out. When run, you should see output similar to:
```

docs/run-platform-server/running.mdx
```diff
@@ -13,36 +13,68 @@ If you chose containerized installation, you don't execute the horizon binary di
 
 - To run the image as a container on docker daemon of host, use `docker run stellar/stellar-horizon:<tag_version> --env-file <env_filename>`, and specify each horizon configuration flag identified during [Configuring](./configuring.mdx) as separate line in `<env_filename>` of `HORIZON_CONFIG_PARAM=value`.
 
-- To run on Kubernetes using helm chart, ensure you have followed the [pre-requisite](./installing.mdx#helm-chart-installation) of installing the Helm cli tool and adding the Stellar chart repo to helm client, now deploy the horizon helm chart to the cluster, including `--devel` uses the latest development chart version:
+- To run on Kubernetes using helm chart, ensure you have followed the [pre-requisite](./installing.mdx#helm-chart-installation) of installing the Helm cli tool and adding the Stellar chart repo to helm client:
 
-    <CodeExample>
+  The horizon process [requires access to a Postgres 12 database](./configuring.mdx#preparing-the-database), first use the common kubernetes cli tool `kubectl` from your workstation to create a kubernetes secret on the intended namespace of the kubernetes cluster which will hold the horizon database url.
```
aditya1702, line 18:
```suggestion
  The horizon process [requires access to a Postgres 12 database](./configuring.mdx#preparing-the-database). First use the common kubernetes cli tool `kubectl` from your workstation to create a kubernetes secret on the intended namespace of the kubernetes cluster which will hold the horizon database url.
```

aditya1702 (approved):
Looks good. Just some small changes.

docs/run-platform-server/running.mdx
sreuland: was treating it as more of a detail related to parent section, following the existing pattern of a bullet list for related info, rather than pulling up to section header. the compute resources will be initially stated in prerequisites.mdx, which is active now - https://github.com/stellar/go/issues/4960

PR #226, opened 2023-09-06
Fix sep-10 auth with TS

Ifropc:


docs/building-apps/wallet/sep10.mdx
```diff
@@ -62,11 +62,10 @@ suspend fun getAuthToken(): AuthToken {
 ```
 
 ```typescript
-let authKey = Keypair.fromSecret("my secret key");
+const authKey = SigningKeypair.fromSecret("my secret key");
```
Ifropc, line 65: Previously were using horizon sdk KeyPair

piyalbasu (approved):

docs/building-apps/wallet/sep10.mdx
acharb: woops, thanks

acharb (approved):
lgtm!

PR #231, opened 2023-09-12
Partner Experience: Update Horizon HW spec

urvisavla:
Update based on [findings](https://github.com/stellar/go/issues/4960#issuecomment-1714554635) from https://github.com/stellar/go/issues/4960

### DIFF
--- a/docs/run-platform-server/prerequisites.mdx
+++ b/docs/run-platform-server/prerequisites.mdx
@@ -20,7 +20,7 @@ The minimum hardware specifications to effectively run Horizon are as follows:
 
 - CPU: 4 cores
 - RAM: 16 GB
-- Storage: SSD with a capacity of 100 GB
+- Storage: SSD with a capacity of 100 GB capable of handling at least 1.5K IOPS (I/O operations per second).
 
 **_PostgreSQL Database Server Compute Instance:_**
 
--- a/docs/run-platform-server/prerequisites.mdx
+++ b/docs/run-platform-server/prerequisites.mdx
@@ -19,7 +19,7 @@ The minimum hardware specifications to effectively run Horizon are as follows:
 **_Horizon Compute Instance:_**
 
 - CPU: 4 cores
-- RAM: 32 GB
+- RAM: 16 GB
 - Storage: SSD with a capacity of 100 GB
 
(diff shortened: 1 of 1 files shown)


docs/run-platform-server/prerequisites.mdx
```diff
@@ -19,7 +19,7 @@ The minimum hardware specifications to effectively run Horizon are as follows:
 **_Horizon Compute Instance:_**
 
 - CPU: 4 cores
-- RAM: 32 GB
+- RAM: 16 GB
 - Storage: SSD with a capacity of 100 GB
```
sreuland, line 23:
```suggestion
- Storage: SSD with a minimum capacity of 100 GB and capable of handling at least 3K IOPS (I/O operations per second) and 3MB/s throughput.
```

this is based on observations of captive core on disk performance is tightly related to the physical disk speeds - https://github.com/stellar/ops/issues/1667

sreuland (approved):
concise change! left one comment for consideration on ssd throughput.

docs/run-platform-server/prerequisites.mdx
urvisavla: I haven't noticed captive core instances requiring 3K IOPS. Typically, the IOPS remains well below 1K with occasional spikes up to 1.5K. I'm wondering if there have been any changes since https://github.com/stellar/ops/issues/1667 which might account for lower IOPs?

sreuland: yes the IOPS demand may have lessened a bit since `BUCKETLIST_DB=true` was added on horizon captive core config by default recently on https://github.com/stellar/go/pull/4733. with that enabled, it lessens the amount of db round trips that core needs to make to its on disk db.

that IOPS demand may have been more pronounced in the captive core `catchup` period when horizon first starts and not the `run` period. but if you don't see it surpassing that, then cool. It may be worthwhile to include mention of a minimum IOPS(ops per second) and throughput(bytes per second) rates as part of `SSD` in any case, based on the numbers you observed as the min baseline.

urvisavla: I have updated the IOPs, but I'm leaving the throughput requirement for the time being. I'll need to do additional data observation for that.

PR #215, opened 2023-08-23
SDFSOF-122: Updated stellar documentation with JSON-RPC API and Custody Integration details

MazurakIhor:
### What

- Replaced POST /transactions to JSON-RPC calls
- Updated SEP24/31 integration pages to include JSON-RPC details.
- Added JSON-RPC API section and descriptions of added RPC methods
- Added section about custody integration and its configuration

### Why

- Documentation should be up-to-date and contain information about custody integration and JSON-RPC API.

PR #232, opened 2023-09-12
anchor platform: add documentation for events

JakeUrban:
Adds the `POST /event` callback to the Callback API documentation. Includes a brief description of how to use event callbacks and the importance of recovering from missed events.

I think it makes sense to update the docs section once we merge SoftServe's work, since their PR (#215) changes the structure so much.


openapi/anchor-platform/bundled_callback.yml
```diff
@@ -711,6 +731,335 @@ components:
...
+            - transaction_created
+            - transaction_status_changed
+            - transaction_error
+            - quote_created
+          description: |
+            The transaction event type. Can be one of the following:
+            - `transaction_created` - a transaction was created through the SEP endpoints. The payload is in the `transaction` field.
+            - `transaction_status_changed` - the status of a transaction has changed. The payload is in the `transaction` field.
+            - `transaction_error` - error processing the transaction. The payload is in the `transaction` field.
+            - `quote_created - a quote was created via the SEP38 API. The payload is in the `quote` field.
```
Ifropc, line 1045:
```suggestion
            - `quote_created` - a quote was created via the SEP38 API. The payload is in the `quote` field.
```

openapi/anchor-platform/bundled_callback.yml
```diff
@@ -711,6 +731,335 @@ components:
...
+          enum:
+            - transaction_created
+            - transaction_status_changed
+            - transaction_error
+            - quote_created
+          description: |
+            The transaction event type. Can be one of the following:
+            - `transaction_created` - a transaction was created through the SEP endpoints. The payload is in the `transaction` field.
+            - `transaction_status_changed` - the status of a transaction has changed. The payload is in the `transaction` field.
+            - `transaction_error` - error processing the transaction. The payload is in the `transaction` field.
```
Ifropc, line 1044: If it's an error, how do we retrieve an error itself?

openapi/anchor-platform/bundled_callback.yml
```diff
@@ -711,6 +731,335 @@ components:
...
+            - `quote_created - a quote was created via the SEP38 API. The payload is in the `quote` field.
+        timestamp:
+          type: string
+          format: date-time
+        payload:
```
Ifropc, line 1049: Ideally payload should be
```
oneOf
   - PayloadTransaction
   - PayloadQuote
 
PayloadTransaction:
oneOf
   - PayloadTransactionSep24 ( object is transaction: TransactionSEP24 )
   - PayloadTransactionSep31 (object is transaction: TransactionSEP31 )
```

But types in openapi is very tedious so idk if it's worth to even bother

openapi/anchor-platform/bundled_callback.yml
JakeUrban: `transaction_error` means one of two things

1. the anchor platform experienced an unexpected error and put the transaction in the `error` status as result. If this happens, the business has to triage the issue and if possible, move it back out of the error status and potentially updating other fields so that it can continue processing. At best, the platform will have put an informative message in the `message` field as a starting point.
2. the business put the transaction into the error status itself via `PATCH /transactions` or the associated RPC method. In this case, the business knows what the error was.

I have a 3.0 jira issue for merging `transaction_error` event types into the `transaction_status_changed` type.

openapi/anchor-platform/bundled_callback.yml
JakeUrban: We don't have to decide on this now but I actually don't know if creating different permutations of the platform transaction schema is the right approach, as opposed to having a unified platform transaction schema documented. I think this makes a bit more sense, even if we have to explain that the value of `transaction.sep` has an impact on the expected values for `transaction.status` and `transaction.kind`.

openapi/anchor-platform/bundled_callback.yml
Ifropc: I think it's a good idea to put this message in the OpenAPI file spec

openapi/anchor-platform/bundled_callback.yml
Ifropc: I think having separate types between SEPs make sense to indicate mandatory/optional/not used fields. Everything else is too much work to clearly separate types.

openapi/anchor-platform/bundled_callback.yml
JakeUrban: I think it is, I see the message field on in the docs.

openapi/anchor-platform/bundled_callback.yml
JakeUrban: I don't think the fields differ between SEPs though, other than the values of `status` and `kind` right?

Ifropc: No, there are quite a few differences in platform transaction between sep-24 and sep-31

Ifropc:
Can you re-run codegen task? Backtick is fixed in yaml file, but rendered version is still incorrect

openapi/anchor-platform/bundled_callback.yml
JakeUrban: Lets merge this as-is but we can continue to discuss this. The intention behind the platform API was to unify the schema across SEPs so if there are schema differences we should resolve them.

Ifropc: image
image

JakeUrban: Seems like the diff is:

- `kind` has different values depending on the value of `sep`
- `status` has different values depending on the value of `sep`
- `quote_id` is present for SEP-31 transactions but not for SEP-24.
  - this should be present on SEP-24 though, since its supported in the protocol, no?
- `source_account` and `destination_account` is present on SEP-24, but not SEP-31
  - I don't see a reason why they shouldn't be on SEP-31
- `memo` and `memo_type` is on SEP-24 but not SEP-31
  - same here, not sure why we don't have this on SEP-31
- `customers` is on SEP-31, not SEP-24
  - we've discussed this before, I think the outcome was to add `customers` to SEP-24 but sender and receiver will be the same
- `creator` is on SEP-31, not SEP-24
  - this actually seems like a bug, how do businesses know which wallet created the transaction without this?

In every case above, there doesn't seem to be a problem making the schema's match.

Ifropc (approved):

PR #241, opened 2023-09-21
Partner Experience: horizon upgrade best practices

### LINKED ISSUE #145: Horizon upgrade best practices
No copy present in current docs https://developers.stellar.org/docs/run-api-server/ for how to upgrade a recent installed horizon to newer version. There is copy on migrating from 1.x which is obsolete now?

* Rename [Migrating from 1.x](https://developers.stellar.org/docs/run-api-server/migrating) menu to be 'Upgrade', and repurpose the right hand side page content:
  * How does user verify if state rebuilds will be triggered? 
  * How does user know if database schema upgrades will be triggered and for how long?
    * Changelog usually mentions if db schema changed and expected duration of the new migrations step added ...
    * What mechanisms or cli tricks are available to check if a schema migration will be triggered from current to new? 
      * install new horizon on a machine that has access to db, don't start it, just run `horizon db migrate status` ... what else is better to do here? 
  * Describe  how db schema migration is invoked automatically by horizon at start up, are there any command line flags that affect or related to this upgrade behavior? I.e. if ingest=false, it skips attempting schema upgrades and just starts web server.
  * What are recommended pre-upgrade steps:
    * determine if a new version of horizon will trigger a schema update
    * back up db
    *  multi-instance deployment considerations, rolling upgrades, if helm this is managed.

sreuland:
Create a guide focused on upgrading from and to Horizon 2.x versions.

Provide insight on db migrations and state rebuilds and how the effect the installation.

Give guidance on how to prepare and plan for an upgrade before doing an upgrade.

Offer tips to confirm upgrade completion, and triage steps when it hasn't.

Closes #145

### DIFF
--- a/docs/run-platform-server/scaling.mdx
+++ b/docs/run-platform-server/scaling.mdx
@@ -23,12 +23,6 @@ For low to medium load environments with up to 30-90 days of data history retent
 
 ![](/assets/horizon-scaling/Topology-2VMs.png)
 
-### Isolating Captive Core
-
-Additionally, Captive Core can be further isolated into its own VM, especially for isolating high throughput historical catch-up with parallel workers, leaving it unaffected by API request servicing load.
-
-![](/assets/horizon-scaling/Topology-3VMs.png)
-
 ## Enterprise _n_-Tier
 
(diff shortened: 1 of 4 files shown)


docs/run-platform-server/scaling.mdx
```diff
@@ -23,7 +23,7 @@ For low to medium load environments with up to 30-90 days of data history retent
 
 ![](/assets/horizon-scaling/Topology-2VMs.png)
 
-### Extension: Isolating Captive Core
+### Isolating Captive Core
```
tamirms, line 26: We have deprecated remote captive core so I don't think it's possible to isolate captive core to its own VM

sreuland: great catch, removed, thanks!

tamirms (approved):

aditya1702 (approved):
Lgtm!

PR #248, opened 2023-10-16
update tier-1-orgs

nano-o:
I would like to suggest updating the documentation on Tier 1 organizations as per this PR. Happy to discuss.

### DIFF
--- a/docs/run-core-node/tier-1-orgs.mdx
+++ b/docs/run-core-node/tier-1-orgs.mdx
@@ -5,7 +5,7 @@ sidebar_position: 120
 
 To help with Stellar’s decentralization, the most advanced teams building on Stellar run validators and strive to join the ranks of “Tier 1 organizations.”
 
-Tier 1 organizations are the organizations that are members of minimal quorums in the core of the Stellar network. This is an emergent property of the Stellar network that depends on the configurations of all the validators in the Stellar network.
+Remember that the Stellar network consists of organizations that each run validators, and each organization decides for itself, by configuring a quorum set, which and how many other organizations it requires agreement from in order to commit to a particular new ledger. Tier 1 organizations are a group of organizations that, due to the fact that most other organizations require agreement from them, bear the safety and liveness of the Stellar network on their shoulders[^1].
 
 To become a Tier 1 organization, a team running validators must convince enough other organizations in the Stellar network to trust them by including them in their quorum sets. As part of this process, they must meet some requirements that are accepted by the community of Stellar validators. For example, Tier 1 organizations generally run three validators, coordinate any changes to their quorum sets with each other, and hold themselves to a higher standard of uptime and responsiveness.
 
@@ -76,3 +76,5 @@ You should do regular check-ins on your quorum set. If nodes have bad uptime or 
 If you think you can be a Tier 1 organization, let us know on the #validators channel on the [Stellar Developers Discord](https://discord.gg/stellardev). We can help you through the process, and once you’re up and running, we’ll help you join Tier 1 so that you can take your rightful place as a pillar of the network. Once you’ve proven that you are responsive, reliable, and maintain good uptime, we will recommend that other validators adjust their quorum set to include your validators.
 
 As Stellar grows, and more and more businesses build on the network, Tier 1 organizations will be crucial to a healthy expansion of the network.
+
+[^1] The notion of Tier 1 organization can be defined precisely, but this is besides the point of this document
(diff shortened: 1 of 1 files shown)


rice2000 (changes requested):
I think we should point to Discord since Keybase is (de facto) deprecated at this point.

docs/run-core-node/tier-1-orgs.mdx
```diff
@@ -3,39 +3,45 @@ title: Tier 1 Organizations
 sidebar_position: 120
 ---
 
-To help with Stellar’s decentralization, the most reliable and advanced Stellar teams join the ranks of “Tier 1 Organizations.” These organizations run three validators, coordinate any changes to their quorumsets, and hold themselves to a higher standard of uptime and responsiveness.
+To help with Stellar’s decentralization, the most advanced teams building on Stellar run validators and strive to join the ranks of “Tier 1 organizations.”
 
-SDF works closely with Tier 1 Orgs to ensure the health of the network, maintain good quorum intersection, and build in redundancy to minimize network disruptions. This guide outlines what it takes to be a Tier 1 Org.
+Tier 1 organizations are the organizations that are members of minimal quorums in the core of the Stellar network. This is an emergent property of the Stellar network that depends on the collective QSet configurations of the validators in the Stellar network.
```
ire-and-curses, line 8: Can we link out to a definition or glossary term for "QSet"? Or otherwise define it the first time it's used.

ire-and-curses (approved):
Looks great! Just one suggestion.

docs/run-core-node/tier-1-orgs.mdx
nano-o: I was trying to not use "quorum set" because I think that term is very confusing and we should get rid of it. However, given how pervasive it is in our docs and code, that might be a lost cause. Should I just go with "quorum set"?

nano-o: I'll just go with "quorum set" for now and we can discuss this later.

docs/run-core-node/tier-1-orgs.mdx
```diff
@@ -3,39 +3,45 @@ title: Tier 1 Organizations
 sidebar_position: 120
 ---
 
-To help with Stellar’s decentralization, the most reliable and advanced Stellar teams join the ranks of “Tier 1 Organizations.” These organizations run three validators, coordinate any changes to their quorumsets, and hold themselves to a higher standard of uptime and responsiveness.
+To help with Stellar’s decentralization, the most advanced teams building on Stellar run validators and strive to join the ranks of “Tier 1 organizations.”
 
-SDF works closely with Tier 1 Orgs to ensure the health of the network, maintain good quorum intersection, and build in redundancy to minimize network disruptions. This guide outlines what it takes to be a Tier 1 Org.
+Tier 1 organizations are the organizations that are members of minimal quorums in the core of the Stellar network. This is an emergent property of the Stellar network that depends on the configurations of all the validators in the Stellar network.
```
rice2000, line 8: This is the one sentence I get stuck on.  I think I trip over "minimal quorums" (what are those? why are there multiple?  what makes them minimal?), and over the "core of the Stellar network" (not Stellar Core, but something else?) and "emergent property" (that one I can look up, but still it's a difficult concept) and "depends on the configurations of all validators" (_all_ validators?  hard to wrap my head around).  Maybe skip this?  Or move it later, after the next two paragraphs that are a bit easier to grep?

nano-o: What about the following?

Remember that the Stellar network consists of organizations that each run validators, and each organization decides for itself, by configuring a quorum set, which and how many other organizations it requires agreement from in order to commit to a particular new ledger. Tier 1 organizations are a group of organizations that, due to the fact that most other organizations require agreement from them, bear the safety and liveness of the Stellar network on their shoulders[^1].

[^1] This can be defined precisely, but is besides the point of this document

nano-o: That's a larger discussion, but I think that it might make sense to switch from talking about "quorum sets" (which is bad because a quorum set is not a set of quorums) to talking about "agreement requirements". Especially since we use "proof of agreement".

rice2000: I think that new bit is fantastic!  Very clear.  Re: "quorum sets," I'm not opposed to changing it ("trust configurations"?), but I think you're correct that it would take a fair amount of finding and replacing, and the change may not stick since people are used to saying "quorum sets," and it can be hard to change usage once a phrase is widespread.

rice2000 (approved):

PR #253, opened 2023-10-24
[ANCHOR-357] SEP-6 documentation updates

philipliu:
This adds SEP-6 support to the Anchor Platform API documentation.

Note: This will be merged when Anchor Platform 2.4 is released.


static/assets/rpc-methods/notify_refund_pending.json
```diff
@@ -94,6 +97,7 @@
           "type": "integer",
           "description": "Protocol of the transaction",
           "enum": [
+            "6",
```
Ifropc, line 100: Somehow it's broken in the preview, doesn't show SEP-31 badge for some reason, can you try re-generate mdx?

Ifropc (approved):

static/assets/rpc-methods/notify_refund_pending.json
philipliu: Does the RPC support SEP-31? It doesn't look like it according to the [handler](https://github.com/stellar/java-stellar-anchor-sdk/blob/c4e756846970136b267cd52c4891e16d1c799a4e/platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyRefundPendingHandler.java#L96-L118).

Ifropc: It's in the enum though, should it be removed?

philipliu: The transaction object schema seems to be exactly the same across all RPC actions, regardless if they apply to the SEP. I just added it here for consistency.

Ifropc: Ah nvm I was looking at the wrong place. I think tag should be `SEP - 6` not `SEP -6` in order for it to render correctly?
Also I'd prefer if SEP tags didn't have space between the dash, but it's something we can fix in another PR

philipliu: > Ah nvm I was looking at the wrong place. I think tag should be SEP - 6 not SEP -6 in order for it to render correctly?

Nice catch, let me fix that.

> Also I'd prefer if SEP tags didn't have space between the dash, but it's something we can fix in another PR

I'll submit a PR to fix that after this is merged.

Ifropc (approved):

PR #255, opened 2023-10-25
Update ledgers.mdx

rice2000:
Contract Data are the new Soroban data entry.  Could be elaborated on later, but this is a quick add to start with.


ElliotFriend (approved):
lgtm

PR #265, opened 2023-11-07
feat: add reader feedback buttons to the bottom of each page

### LINKED ISSUE #264: Provide a mechanism for readers to provide simple feedback on page quality

We don't have an easy/quick way to garner feedback from site readers.

As a follow-up to stellar/soroban-docs#586, we can provide users with a simple set of thumbsup/thumbsdown buttons. Essentially, it's exactly the same thing, just here.

### What alternatives are there?

Don't have the buttons? Use some third-party?

ElliotFriend:
Provide users a simple thumbsup/thumbsdown to signify whether they find the given page useful or not.

Resolves #264 (at least as a first effort)

Refs: #264, stellar/soroban-docs#586

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -25,13 +25,6 @@ const config = {
       },
     ],
     [
-      "@docusaurus/plugin-google-analytics",
-      {
-        trackingID: "UA-53373928-1",
-        anonymizeIP: true,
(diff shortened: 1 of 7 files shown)


briwylde08 (approved):
Yay!

PR #273, opened 2023-11-21
docs: Updating testnet reset from 9am UTC to 5pm UTC

xiv:
Updating the testnet reset timing from 9am(09:00) UTC to 5pm(17:00) UTC


rice2000 (approved):
LGTM!  Thanks!

PR #274, opened 2023-11-21
Clarity on txsub configurations

sreuland:
Added more clarity to how to use INGEST and STELLAR_CORE_URL on tx sub enabled instance.

### DIFF
--- a/docs/run-platform-server/configuring.mdx
+++ b/docs/run-platform-server/configuring.mdx
@@ -62,16 +62,16 @@ You can run none or multiple instances to serve read only API requests. Notice t
 
 ### Transaction Submission Role Instance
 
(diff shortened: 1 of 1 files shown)


urvisavla (approved):
lgtm!

I'd like to clarify that `STELLAR_CORE_URL` is *not* required for running a API-only instance (`INGEST=false` and `DISABLE_TX_SUB=true`)? The docs indicate that but I think the code still expects that `STELLAR_CORE_URL` be explicitly defined in that case.
cc @aditya1702

sreuland:
> I'd like to clarify that `STELLAR_CORE_URL` is _not_ required for running a API-only instance (`INGEST=false` and `DISABLE_TX_SUB=true`)? The docs indicate that but I think the code still expects that `STELLAR_CORE_URL` be explicitly defined in that case.

was able to replicate that on master, filed new bug to get that fixed - https://github.com/stellar/go/issues/5124

PR #294, opened 2024-01-25
Anchor Platform: Add fields to PUT /customer callback

philipliu:
This updates the PUT `/customer` callback documentation with the new binary fields.


Ifropc:
Do you mind moving all binary fields to the end of the list (so it's easier to see them all together)?

Ifropc (approved):

philipliu:
@Ifropc It follows the same ordering defined in the SEP-12 protocol. The AP code also follows the same ordering, so I would like to keep it as is unless we change it everywhere else.

PR #303, opened 2024-02-14
p20 copy updates

briwylde08:
These copy updates can be merged following a successful Protocol 20 upgrade vote


ElliotFriend (approved):
looks great to me!

PR #319, opened 2024-02-27
fix the link to soroban docs in announcement bar

ElliotFriend:
I messed it up the first time and put `href="_blank"` instead of `target="_blank"`, so it doesn't work


briwylde08 (approved):
Pure magic.

PR #325, opened 2024-02-28
[Soroban Merge] Create a "Platforms" section

ElliotFriend:
This PR adds a "Platforms" navbar item, that drops down to include Horizon, SDP, Anchor Platform. I've also moved the "Admin Guides" from the `/docs` section into these respective sections so as to keep everything together.

Still some work to do:

- [x] Soroban-RPC should be moved over here from `soroban.stellar.org`
- [x] We use a custom `ApiItem` as the `docItemComponent` for the auto-generated openapi pages, but it looks less than good for those admin guide pages. We could probably just add some logic to that `ApiItem` component to ignore them?
- [x] nginx redirects haven't been added/updated yet
- [x] make sure the scripts in `package.json` are properly written to reflect new filepaths
    - [x] test to make sure the redocly things build properly still

Refs: stellar/soroban-docs#757


kalepail (approved):
🥳

PR #349, opened 2024-03-05
[ANCHOR-603] Add fee_details documentation

Ifropc:


static/assets/rpc-methods/do_stellar_payment.json
```diff
@@ -132,6 +132,35 @@
...
+            },
+            "asset": {
+              "type": "string"
+            },
+            "details": {
+              "type": "array",
+              "description": "Details about fee charged",
+              "items": {
+                "properties": {
+                  "name": {
```
philipliu, line 150: Shouldn't `name` and `amount` be required?

static/assets/rpc-methods/notify_interactive_flow_completed.json
```diff
@@ -91,6 +91,40 @@
         ]
       }
     },
+    {
+      "name": "fee_details",
+      "description": "Description of fees charged by the anchor",
```
philipliu, line 96: Should we update the description so it's consistent with the ones in the other RPCs?

philipliu (commented):
I think we would also need to update the integration guides. Do we plan to do it in another PR?

static/assets/rpc-methods/notify_interactive_flow_completed.json
Ifropc: Done

philipliu (approved):

PR #374, opened 2024-03-19
100 soroban txs per ledger

briwylde08:
The networks section said there was a max of 30 Soroban txs per ledger when there are really 100 txs per ledger


ElliotFriend (approved):
great catch! thanks

PR #387, opened 2024-03-24
Create 2023-03-21.mpx

Danasacco:


kalepail:
The folder name is 2023, should be 2024.

Also let's include relevant links inline with the comments. There were quite a number that were linked during the event, we should use those.

kalepail (approved):

PR #398, opened 2024-03-28
Fix the fully typed examples

leighmcculloch:
### What
Update the examples in the fully typed docs learn page to use contract IDs and the latest CLI command structure.

### Why
It appears it's way out of date and these commands won't work. One of the example commands no longer exist. And the use of hashes for contract IDs is no longer supported.

### DIFF
--- a/docs/learn/smart-contract-internals/types/fully-typed-contracts.mdx
+++ b/docs/learn/smart-contract-internals/types/fully-typed-contracts.mdx
@@ -69,11 +69,17 @@ Options:
   -h, --help  Print help
 ```
 
-Like any other CLI, you can also get help for any of these subcommands using something like `native balance --help`. Soroban CLI again fetches the on-chain interface types, this time using it to generate a full list of all arguments to the function, and even generates examples. And just to be totally clear, you don't need to alias these commands. This would work, too:
+Like any other CLI, you can also get help for any of these subcommands using something like `native balance --help`. Soroban CLI again fetches the on-chain interface types, this time using it to generate a full list of all arguments to the function, and even generates examples.
+
+:::tip
+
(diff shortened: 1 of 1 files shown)


docs/learn/smart-contract-internals/types/fully-typed-contracts.mdx
```diff
@@ -35,49 +35,44 @@ So that's what Soroban CLI gives you.
...
+  -h, --help  Print help
+```
 
-  -h, --help
-          Print help (see a summary with '-h')
+Like any other CLI, you can also get help for any of these subcommands using something like `native balance --help`. Soroban CLI again fetches the on-chain interface types, this time using it to generate a full list of all arguments to the function, and even generates examples. And just to be totally clear, you don't need to alias these commands. This would work, too:
+
+```bash
+soroban contract invoke --network testnet --id CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC -- balance --help
 ```
```
ElliotFriend, line 76: The "you don't need to alias..." part might make for a good admonition box


````suggestion
Like any other CLI, you can also get help for any of these subcommands using something like `native balance --help`. Soroban CLI again fetches the on-chain interface types, this time using it to generate a full list of all arguments to the function, and even generates examples.

:::tip

And just to be totally clear, you don't need to alias these commands. This would work, too:

```bash
soroban contract invoke --network testnet --id CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC -- balance --help
 ```

:::
````

ElliotFriend (commented):
Looks really good! Thanks for making this update!

ElliotFriend (approved):
Looks great! Thanks

PR #401, opened 2024-03-28
properly redirect for the `tier-1-orgs` page

ElliotFriend:


Shaptic (approved):
LGTM, but do they need a `/?` at the end of each match in case there's a trailing slash in the link?

ElliotFriend:
Great question! I think we have that taken care of with the nginx config here: https://github.com/stellar/stellar-docs/blob/3dc16ba636d6e912db7cf8e81d06ac8fc2bb148b/nginx/nginx.conf#L20-L22

i'm certainly not an nginx expert, but i think we've managed to get the redirects/configs right (finally lol) to take care of trailing slashes

Shaptic:
@ElliotFriend sounds good! I was reviewing based on the nearby lines which had that included but it sounds like those are redundant

Issue #414, opened 2024-04-02
[Post merge cleanup] Add definiton of Soroban in the glossary


janewang:
### What problem does your feature solve?
We use the name “Soroban” throughout the documentation but never explain it. In the Soroban docs, we talked about the name and its origin. Let's move that over

janewang:
PR opened https://github.com/stellar/stellar-docs/pull/417

briwylde08:
@janewang I think we can close this one with the new Soroban glossary entry. Need to keep in mind that we will need to change the official definition of Soroban to include only the runtime environment. This is mostly relevant for our Welcome page.

janewang:
Agreed, closing

PR #417, opened 2024-04-02
Add description for soroban name

janewang:


briwylde08:
I'm not sure if we need to define Soroban as an abacus here, it kind of clogs up the intro page a bit IMO. Maybe we can add a Glossary entry for Soroban and mention it there!

src/pages/index.mdx
```diff
@@ -22,10 +22,12 @@ This section walks you through building on Stellar without smart contracts. Lear
 
 ### Smart contracts
 
-Smart contracts on Stellar launched on Mainnet following a successful validator vote on February 20th, 2024. Stellar's smart contract platform includes the smart contract environment, a Rust SDK, a CLI, and an RPC server. 
+Smart contracts on Stellar launched on Mainnet following a successful validator vote on February 20th, 2024. Stellar's smart contract platform, known as Soroban, includes the smart contract environment, a Rust SDK, a CLI, and an RPC server.
 
 This section details how to get started by writing a Hello World contract, then dives deeper into smart contracts on Stellar with various example contracts and how-to guides. Also, learn how to use Stellar assets in smart contracts with the Stellar Asset Contract or how to create your own smart contract token.
 
+The name "Soroban" comes from the Japanese abacus, which is a traditional counting tool used for mathematical calculations. The Soroban abacus is known for its efficiency and accuracy in performing arithmetic operations.
```
leighmcculloch, line 29: Maybe worth also referencing that it is a lightweight tool.

src/pages/index.mdx
```diff
@@ -22,10 +22,12 @@ This section walks you through building on Stellar without smart contracts. Lear
 
 ### Smart contracts
 
-Smart contracts on Stellar launched on Mainnet following a successful validator vote on February 20th, 2024. Stellar's smart contract platform includes the smart contract environment, a Rust SDK, a CLI, and an RPC server. 
+Smart contracts on Stellar launched on Mainnet following a successful validator vote on February 20th, 2024. Stellar's smart contract platform, known as Soroban, includes the smart contract environment, a Rust SDK, a CLI, and an RPC server.
```
leighmcculloch, line 25: I have been thinking about the way we use the term Soroban to more narrowly refer to the contract runtime, rather than a platform as a whole. I might be thinking about this incorrectly. But for example, Stellar is a platform because it provides the entire stack, and the Stellar platform uses the Soroban runtime to run contracts. In this world the only things labeled as "soroban" are the runtime itself (which we call the Soroban Environment) and the soroban Rust SDK which is used to build contracts to run on the Soroban Enviroment. Everything else surrounding these things are I think not Soroban specific but part of the Stellar platform. For example, the CLI, the RPC, while they started as Soroban specific tools, they are more broadly part of the Stellar platform and if another platform adopted Soroban they'd really need to dev their own CLI and RPC.

But I say all this not knowing if the nomenclature I'm using is accurate.

cc @tomerweller

janewang: I think Soroban historically has been referring to the smart contract platform. If we are renaming to only the contract runtime, which I don't oppose, we may need to also think about some of the marketing language on stellar.org, as I think that has wider impact than dev docs.

janewang:
@briwylde08 I love the idea of adding the definition of soroban in the glossary. Here's the ticket: https://github.com/stellar/stellar-docs/issues/414

Waiting for more clarity on the scope of soroban naming, then I will close this PR.

src/pages/index.mdx
tomerweller: I think the end goal should be what @leighmcculloch describes - Soroban is the name of the runtime (or "environment", or "virtual machine" to align with evm and svm). 

I recognize that in a lot of places the current scope is much wider and I don't think it's the end of the world for now, especially while tools like soroban-cli and soroban-rpc haven't been refactored yet.

leighmcculloch: > to align with evm

We could use this comparison explicitly as a way to help people understand what Soroban is. Soroban == EVM. Rust == Solidity. Stellar == Ethereum.

janewang: @briwylde08 FYI on `soroban` usage. 

Closing this PR as soroban definition will be in the glossary.

PR #439, opened 2024-04-05
Fix some 404s

ElliotFriend:


briwylde08 (approved):
Way to kick those 404s out of here!

PR #462, opened 2024-04-12
Fix Gitpod link

briwylde08:
fixes https://github.com/stellar/stellar-docs/issues/459

PR #464, opened 2024-04-12
Fix gitpod links.

fnando:
After seeing #462, I noticed that there was a bunch of other broken links.


briwylde08 (approved):
I should have looked at all these when I noticed the other one!! Thank you!

PR #465, opened 2024-04-15
Sync tabs and persist choice on localStorage.

fnando:
These changes will open Windows tabs on Windows by default, and the choice will be persisted on local storage.


ElliotFriend (approved):
great addition! Thanks for the hard work!

PR #466, opened 2024-04-16
Add Soroban errors for better debugging

janewang:


ElliotFriend:
This is REALLY really good! The amount of detail and information provided is really helpful and useful!

The only question I might ask is whether or not "Debugging" is the right page for this? Yes, certainly the information is useful for debugging purposes. I only ask because 1) it's _a lot_ of content (more than already exists there), and 2) the existing content is a little more focused on local testing/debugging, before it's really gets to the network/transaction stage.

Maybe this stuff would be a good fit for this [draft "Transaction Lifecycle" page](https://github.com/stellar/stellar-docs/blob/main/docs/learn/smart-contract-internals/transaction-lifecycle.mdx) that's never been tackled? If so, we should **definitely** add a link/callout on the "Debugging" page, since it's still really useful info.

janewang:
@ElliotFriend I thought the same. I think we could call it `Soroban Errors` because that's what these errors are. There is an Errors page which I think could be called `Generate Errors`: https://developers.stellar.org/docs/learn/smart-contract-internals/errors

For transaction lifecycle I think we should focus more on how a transaction moves through the stages, but less on the errors and how to fix them. I believe Naman or Dima said at one point that they were working on that. 

Wdyt?

janewang:
@ElliotFriend I pushed some changes with Debugging Modes and Debugging Errors. PTAL

docs/learn/smart-contract-internals/debugging-errors.mdx
```diff
@@ -0,0 +1,117 @@
+---
+sidebar_position: 70
```
ElliotFriend, line 2: let's change this to `75`, so it's next in the sidebar after "Debugging Modes"

docs/learn/smart-contract-internals/debugging-errors.mdx
```diff
@@ -0,0 +1,117 @@
...
+  <meta charSet="utf-8" />
+  <meta
+    property="og:title"
+    content="Debug smart contracts natively and as WASM."
+  />
+  <meta
+    property="og:description"
+    content="Compile smart contracts natively for local testing or into WASM for deployment in order to debug."
+  />
+</head>
```
ElliotFriend, line 18: This looks like it's duplicated from other page, so it may need to be changed. Or, we can add a more detailed description on line 4, and this whole thing could be removed. Docusaurus already generates these `<meta>` tags in the document head, based on the "frontmatter" present in the file.

ElliotFriend (changes requested):
This looks really great! Just a couple changes to the docusaurus/markdown stuff, but everything else looks good!

I think you're right about the "transaction lifecycle" being a better fit for the more generalized flow of transactions, instead of error/failure info. 👍🏻

ElliotFriend (approved):
looks great!

Issue #478, opened 2024-04-16
Support Horizon decentralization


janewang:
### What problem does your feature solve?
We'd like to present Horizon in much of the same way that RPC is presented in our docs. 

### What would you like to see?
- API docs use testnet Horizon endpoints
- Horizon mainnet is pointed to our partner page
- Horizon mainnet usage is replaced by testnet usage examples in the docs.

PR #502, opened 2024-04-17
Update horizon open api to testnet

janewang:
This PR attempts to remove the Horizon mainnet url from the open api files and is part of this ticket https://github.com/stellar/stellar-docs/issues/478


ElliotFriend:
this looks good to me! i don't see any problems with it, and `main` merged pretty nicely right back into it.

Probably the main thing to consider with this are the examples objects. For example, on this horizon page:

- [live dev docs](https://developers.stellar.org/network/horizon/api-reference/resources/get-transactions-by-account-id)
- [pr preview](https://developers-pr502.previews.kube001.services.stellar-ops.com/network/horizon/api-reference/resources/get-transactions-by-account-id)

On both of those pages, under the "Responses" section, click the `RetrieveAnAccountsTransactions` button/pill and it pops up with a JSON response. The links in that JSON would point to live data on `horizon.stellar.org`, but result in 404s on `horizon-testnet.stellar.org` because the account doesn't exist on tesnet, or the transaction hasn't happened, etc.

Not truly a show-stopper, but could be confusing/weird if someone is looking at those examples, expecting them to provide useable links. I don't know of an alternative way to get predictably 200 status links for testnet horizon (like after a reset, yaknow).

@briwylde08 As the true "docs master," does that bother you any, or is it a non-issue to you?

janewang:
@ElliotFriend This is the same conversation I raised in the channel https://stellarfoundation.slack.com/archives/C049E67BFSP/p1713448837768389

I have been discussing with core to not fully reset and leave some testnet data behind for better developer experience. In the future,  I believe testnet should be stable enough for us to be able to show reasonable responses but it is not currently the case. Right now, there's a reset every 3 months. Our options are to show stubbed data or to enable user to input data like from Lab. 

Possibly we can mock the response and link to Lab so they could do a proper swagger like request from Lab.

ElliotFriend:
Aaah, i hadn't seen that conversation. Thanks for pointing it out!

It appears to me that user-editable server URLs are likely not going to be implemented in the openapi plugin we use ([see this issue](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/issues/188)). We _could_ add a server in the config file to act as a stub pretty easily.

Another option might be to use accounts like the Genesis account, or Friendbot account, etc. that we know will be real URLs, that (of course) doesn't cover _every_ possible URL that horizon might return (dex offers, for example). Or, we could have a "setup script" that runs at every testnet reset to create the specific accounts/assets/offers/claimable balances/whatever that we use in our examples.

All that said, I don't think it's a blocker here for this PR.

janewang:
@ElliotFriend Oh I really like the idea of having a script that could run after testnet reset. 

I could open a ticket for that. If no blocker for this PR, we could merge. I need a approve to merge. 

I have one more section to get to before I finish removing all mainnet horizon from our docs. Looking to get to it next week.

ElliotFriend:
As I've thought a bit more this afternoon, I may have to walk back a bit on my earlier "not a blocker" statements. Maybe it's a bit rushed to merge it without at least a solid idea of what we will do to alleviate the resulting broken URLs it'll cause.

I can do some thinking this weekend, and come back on Monday with some more thoughts or ideas?

PR #448, opened 2024-04-09
Add API Reference sections to Networks

briwylde08:
Fixes: https://github.com/stellar/stellar-docs/issues/371

PR #506, opened 2024-04-18
Fix sidebars in Network pages

ElliotFriend:
when we added "API Reference" sections to the various items within the `/network` directory, we forgot to update the sidebar generator, and a couple other auto-generating things. This PR will fix that and restore the sidebars to their full extent.

Refs: #448


briwylde08 (approved):
Good catch!

PR #516, opened 2024-04-23
Update soroban examples

fnando:


ElliotFriend (approved):
These are some great updates! Thank you!!

It looks like this unix/windows set of TabItems might make for a good custom component (down the road, not in this PR). We have [this custom CodeExample](https://github.com/stellar/stellar-docs/blob/main/src/components/CodeExample.js) one (that we don't always use) that we could maybe copy from and/or repurpose to suit these needs. That might save on some of the duplicate imports and such.

Do you think your `getPlatform` helper function would still work in that way?

fnando:
Yeah, I think extracting a component would be a great idea. I didn't want to prematurely try to add a new abstraction, but I get down to it once every single article is converted (there's just one left but will require a new soroban-cli release because it doesn't work on Windows at all).

PR #520, opened 2024-04-24
Update contract getting started

BlaineHeffron:
Updates the smart contracts getting started tutorial to reflect the new ContractClient class in the js-stellar-sdk. Also fixes a typo of the hello world contract name.

### DIFF
--- a/docs/smart-contracts/getting-started/create-an-app.mdx
+++ b/docs/smart-contracts/getting-started/create-an-app.mdx
@@ -222,27 +222,27 @@ Now let's add a new component to the `src/components` directory called `ConnectF
 
 ```html title="src/components/ConnectFreighter.astro"
 <div id="freighter-wrap" class="wrap" aria-live="polite">
-    <div class="ellipsis">
-      <button data-connect aria-controls="freighter-wrap">Connect</button>
-    </div>
+  <div class="ellipsis">
+    <button data-connect aria-controls="freighter-wrap">Connect</button>
   </div>
-  
-  <style>
-    .wrap {
-      text-align: center;
-    }
-  
-    .ellipsis {
-      line-height: 2.7rem;
-      margin: auto;
-      max-width: 12rem;
-      overflow: hidden;
-      text-overflow: ellipsis;
-      text-align: center;
-      white-space: nowrap;
-    }
-  </style>
-  
+</div>
+
+<style>
+  .wrap {
+    text-align: center;
+  }
+
+  .ellipsis {
+    line-height: 2.7rem;
+    margin: auto;
+    max-width: 12rem;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    text-align: center;
+    white-space: nowrap;
+  }
+</style>
+
 <script>
   import { isAllowed, setAllowed, getUserInfo } from '@stellar/freighter-api';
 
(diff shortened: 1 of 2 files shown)


ElliotFriend (approved):
Thanks for the fixes! Looks great

docs/smart-contracts/getting-started/create-an-app.mdx
```diff
@@ -255,20 +255,21 @@ Now let's add a new component to the `src/components` directory called `ConnectF
...
-    const publicKey = await getPk();
+  async function setLoggedIn(publicKey: string) {
     ellipsis.innerHTML = `Signed in as ${publicKey}`;
     ellipsis.title = publicKey;
   }
 
   if (await isAllowed()) {
-    if (await getPk()) setLoggedIn();
+    const publicKey = await getPk();
+    if (publicKey) setLoggedIn(publicKey);
```
chadoh, line 265: Why this change?

docs/smart-contracts/getting-started/create-an-app.mdx
```diff
@@ -324,23 +325,30 @@ Current value: <strong id="current-value" aria-live="polite">???</strong><br />
...
   const currentValue = document.querySelector("#current-value");
-
+  async function getPk() {
+    const { publicKey } = await getUserInfo();
+    return publicKey;
+  }
+  if (await isAllowed()) {
+    const publicKey = await getPk();
+    if (publicKey) incrementor.options.publicKey = publicKey;
+  }
```
chadoh, line 338: This is fine for now, but I think Freighter now has a `getPublicKey` or just `publicKey` method that we can use here to simplify this logic.

docs/smart-contracts/getting-started/create-an-app.mdx
```diff
@@ -324,23 +325,30 @@ Current value: <strong id="current-value" aria-live="polite">???</strong><br />
...
   button.addEventListener("click", async () => {
     button.disabled = true;
     button.classList.add("loading");
     currentValue.innerHTML =
       currentValue.innerHTML +
       '<span class="visually-hidden"> – updating…</span>';
 
-    const tx = await incrementor.increment();
+    const tx = await incrementor.increment({timeoutInSeconds: 30});
```
chadoh, line 346: Do we need this? Should the default be increased from `10` so we don't need it anymore?

chadoh (approved):
Looks good. I have a couple questions and notes, but none are show-stoppers. I think these are good changes that we should merge sooner rather than later, and then maybe follow up with more PRs later.

docs/smart-contracts/getting-started/create-an-app.mdx
BlaineHeffron: Change was to avoid the two await getUserInfo calls but in retrospect its cost is negligible so doesn't matter.

docs/smart-contracts/getting-started/create-an-app.mdx
BlaineHeffron: It seems to be needed, which appears to be a bug - https://github.com/stellar/js-stellar-sdk/issues/950

docs/smart-contracts/getting-started/create-an-app.mdx
BlaineHeffron: You are correct, this code works:

docs/smart-contracts/getting-started/create-an-app.mdx
BlaineHeffron: I took it out because the default will be changed to 60 for release.

BlaineHeffron:
I added the `signTransaction` method to the sign and send in the `Counter` script. This is needed based on how the current soroban astro template creates the `Client`. I also updated the url to the new stellar repo for the soroban astro template. I removed the `timeoutInSeconds` passed to the increment because it shouldn't be needed with the new default being set to 60 (see https://github.com/stellar/js-stellar-sdk/pull/951). I simplified the logic for getting the public key in `Counter`.

PR #525, opened 2024-04-26
Python should be JS

briwylde08:


ElliotFriend (approved):
good catch. thank you!

PR #538, opened 2024-05-02
Update rpc-providers.mdx +Mainnet Available on NOWNodes

NOWNodes:


ElliotFriend (approved):
Congrats on bringing your Mainnet service online! Thanks for updating our docs with that info!!

PR #540, opened 2024-05-02
Add async txsub endpoint documentation

aditya1702:
This PR adds an openAPI spec to generate the documentation for the new async txsub endpoint in Horizon - https://github.com/stellar/go/pull/5188


openapi/horizon/bundled.yml
```diff
@@ -1163,6 +1163,145 @@ paths:
               examples:
                 SubmitaTransaction:
                   $ref: '#/components/examples/SubmitaTransaction'
+  /transactions_async:
+    post:
+      summary: Submit Async Transaction
```
JakeUrban, line 1168: Its longer, but maybe we should consider "Submit a Transaction Asynchronously". I think that better describes the approach, since the transaction itself isn't async, its the process for submitting it that is async.

openapi/horizon/bundled.yml
```diff
@@ -1163,6 +1163,145 @@ paths:
...
+      description: This endpoint submits a transaction in an async manner to the Stellar network. It is designed to allow users to submit transaction without blocking them while waiting for a response from Horizon. At the same time, it also provides clear response status codes from stellar-core to help understand the status of submitted transaction. You can then use Horizon's [GET transaction endpoint](./retrieve-a-transaction.api.mdx) to check whether the transaction was successfully ingested or not.
```
JakeUrban, line 1169:
```suggestion
      description: This endpoint submits a transaction to the Stellar network asynchronously. It is designed to allow users to submit transaction without blocking them while waiting for a response from Horizon. At the same time, it also provides clear response status codes from stellar-core to help understand the status of the submitted transaction. You can then use Horizon's [GET transaction endpoint](./retrieve-a-transaction.api.mdx) to wait for the transaction to be included in a ledger and ingested by Horizon.
```

openapi/horizon/bundled.yml
```diff
@@ -1163,6 +1163,145 @@ paths:
...
+                TransactionException:
+                  summary: Transaction Submission Exception
+                  value:
+                    type: transaction_submission_exception
+                    title: Transaction Submission Exception
+                    status: 500
+                    detail: 'Received exception from stellar-core. The `extras.error` field on this response contains further details. Descriptions of each code can be found at: https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-submission-async/transaction_submission_exception'
+                    extras:
+                      envelope_xdr: ''
+                      error: Exception details here
```
JakeUrban, line 1271: I'm not familiar with this response case. When would this occur?

openapi/horizon/bundled.yml
```diff
@@ -1163,6 +1163,145 @@ paths:
...
+      parameters:
+        - in: query
+          name: tx
+          schema:
+            type: string
+            description: The base64-encoded XDR of the transaction.
+          required: true
+      responses:
+        '201':
+          description: Transaction has been received by core and is in pending status.
```
JakeUrban, line 1182: Maybe we should again point people to `GET /transactions/:id` to monitor its success or failure.

openapi/horizon/components/schemas/asyncTransactionSubmissionSchema.yml
```diff
@@ -0,0 +1,35 @@
...
+          nullable: true
+          description: TransactionResult XDR string which is present only if the submission status from core is an ERROR.
+        tx_status:
+          type: string
+          enum: ["ERROR", "PENDING", "DUPLICATE", "TRY_AGAIN_LATER"]
+          description: Status of the transaction submission.
+        hash:
+          type: string
+          description: Hash of the transaction.
+    Problem:
```
JakeUrban, line 17: I would've figured the `Problem` schema was already defined for other error types. Just want to make sure we're not duplicating it here.

JakeUrban (commented):
Looks great! I added some phrasing suggestions but nothing blocking.

openapi/horizon/bundled.yml
ElliotFriend: Yeah, I think I'm with @JakeUrban on that. Something that more clearly denotes the submission is the async part, not the tx. What about something like "Submit Transaction (Async)"?

openapi/horizon/components/schemas/asyncTransactionSubmissionSchema.yml
ElliotFriend: I'm curious about this, too.

Along those lines, does anyone know when the last time the Horizon spec has been thoroughly gone through, with an eye towards reducing duplicate definitions and such? My suspicion (perhaps wrongly) is a while or never.

If it would be a benefit, I can find some room in the near-medium term to take a shot.

ElliotFriend (commented):
Thanks for making the additions! I don't think any of Jake's questions are blockers, either. But, I'd love to hear your thoughts before we merge.

openapi/horizon/bundled.yml
aditya1702: @JakeUrban @ElliotFriend Yeah I agree. I think "Submit a Transaction Asynchronously" sounds better here.

openapi/horizon/components/schemas/asyncTransactionSubmissionSchema.yml
aditya1702: @JakeUrban Yeah I checked the code and weirdly, the Problem object is not documented in any schema yamls.

@ElliotFriend I dont think we have updated this part of the Horizon docs (endpoint definitions) in a long time. We last revamped the admin docs but not this one.

openapi/horizon/bundled.yml
aditya1702: @JakeUrban I dont think we should add that info here. I would like to have this description just describe the response status. We already mention checking the GET endpoint in the endpoint summary. What do you think?

openapi/horizon/components/schemas/asyncTransactionSubmissionSchema.yml
aditya1702: Ah I see they are defined here - https://developers-pr540.previews.kube001.services.stellar-ops.com/network/horizon/api-reference/errors/http-status-codes/horizon-specific/transaction-failed

openapi/horizon/bundled.yml
JakeUrban: I think this is probably a failure in the Soroban runtime, like a server 500 error. We can resolve.

JakeUrban (approved):

PR #541, opened 2024-05-02
Move data indexers to Data Availability section

briwylde08:


network/data-indexers/README.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+### [Goldsky](https://goldsky.com/)
+
+Mirror and indexer coming soon.
+
+### [Ankr](https://www.ankr.com/)
+
+Also offers RPC and Horizon instances.
+
+### [Zettablock](https://www.zettablock.com/)
```
ElliotFriend: Is Zettablock actually available for Stellar? I see that we're included in the little scrolling banner area on their site, but in the docs, dashboard, blog, anywhere else on their site I can't seem to find any way to use Stellar data.

I see we added this link a month ago, though, so I'm not sure if something has changed since then.

ElliotFriend: If we do want to keep this here, we may want to figure out an accurate description for it

ElliotFriend:
I really like having these resources nested under the "data availability" section of the site. I pushed a couple of tiny changes (mostly formatting stuff).

There are two things I'm thinking about the change:

1. In the sidebar, the name "Ecosystem Data Indexer Providers" is a bit long and wraps to a second line, which looks a little wonky. I like that it fits with our "standard" nomenclature of "Ecosystem fill-in-the-blank Providers," though. I'm a bit torn. I suppose if it's just the one single page, we don't necessarily _need_ the sidebar there at all.
1. It feels a little weird to have a whole section devoted just to a single page. Maybe it makes sense to somehow include a page for the block explorers in this section? I don't know if "data indexers" would still be the right name for the section, though. Or, maybe it would make sense to call this section "Ecosystem Providers" and include pages for these indexers, block explorers, plus the providers pages from horizon/rpc?

Here's the page in the preview, for reference: https://developers-pr541.previews.kube001.services.stellar-ops.com/network/data-indexers

briwylde08:
1. What about "Data Indexer Providers", I agree it is too long currently.
2. I would prefer to keep it just data indexers and not move other stuff around too much. I agree it is a bit odd to have just a whole section for one page that just links out to stuff but /shrug

briwylde08:
Turning into a draft for further discussion

janewang:
+1 on "Data Indexers", it's what's called in the industry. 

Thank you for working on this @briwylde08!

briwylde08:
TL;DR
I propose we keep Data Indexers in Tools.

I think the discussion here is whether to move the Data Indexers to the Network section. I actually think I'd prefer to keep Data Indexers in the Tools section since they just link out to third parties instead of having full-fledged documentation like Horizon, Hubble, and RPC. I also talked to Alexis and she said that nobody has searched for "Data Indexer" or "Data" or "Indexers" in the last 30 days, so people aren't having a hard time finding them. I propose we keep Data Indexers in Tools.

@ElliotFriend, what do you think?

janewang:
Indexers are strictly related to data availability. It makes more sense to move under Data Availability along side with Horizon, RPC, and Hubble. Our motivation to work on indexers is closely related to issues experienced from Horizon. Just because no one is searching for them doesn't mean we have correctly organized our doc. Where data falls short, we should rely on good judgement.

ElliotFriend:
@janewang @briwylde08

Hi, all!

I'm also thinking the right move here is to keep Data Indexers in the Tools section. I've got a couple reasons I'm leaning this direction:

> **TL;DR**
> - the content is quite different from the rest of the Network section
> - it's pretty sparse to give an entire sub-section to
> - technically speaking, it could give rise to config/maintenance complications
> - other networks have their comparable data indexer content in their comparable Tools sections
> - i think our readers are expecting this _kind_ of content under Tools, rather than Network

1. The content we have for "Data Indexers" is very different from the other content within the "Network" section. We really just link to the various data indexer websites, with just a little bit of context. On the flip side, the other "Network" sub-sections (sdp, core, horizon, etc.) have a pretty comprehensive set of docs, admin guides, etc.
2. There's not much stuff here to justify having its own sub-section under "Network." Again, [here is the page](https://developers-pr541.previews.kube001.services.stellar-ops.com/network/data-indexers) in the preview. It looks a little silly (imo) to have this whole sub-section dedicated to just this single page, and its own sidebar to go along with it.
3. It's kindof a non-issue, but technically speaking the content we have under `/docs` and the content we have under `/network` are handled by two different docusaurus plugins (which are each configured quite differently). That's not to say that it _can't_ work there, but moving the page would come with some additional configuration and maintenance (especially relevant with docusaurus upgrade decisions).

I think you're right on the money, Jane, when you say:

> Where data falls short, we should rely on good judgement.

I think the relevant "good judgment" to consider **must** be rooted in the user's (i.e., the reader's) expectations and experience. From that point of view, it feels like this content is out of place as a "Network" sub-section, and that it's more "at home" in the "Tools" section. Additionally, I've done a quick, non-exhaustive survey of some other chains: [Near](https://docs.near.org/tools/indexing), [Base](https://docs.base.org/docs/tools/data-indexers), and [Polygon](https://docs.polygon.technology/tools/chain-indexer-framework/overview/) have the same kind of data indexer content as we have under their "Tools" sections (even some quite a bit more verbose content). Since we don't have data compelling us toward surfacing the data indexers into another section, it seems like we should follow suit with some of the industry patterns (that we're already doing, anyway).

I really think our readers are expecting to find these kinds of external links (even the data-centric links) under the Tools section. Hopefully that makes sense. Let me know if you have any other questions or thoughts.

janewang (approved):

janewang:
After speaking with Leigh, we think we can move this to the Data availabilities section and reference it from Tools.

briwylde08:
Going to close this and do with the docs v2 merge

PR #542, opened 2024-05-03
Update library.mdx

rice2000:
Update the .NET SDK link so it goes to the current repo (maintained by Beans)


ElliotFriend (approved):
Thanks!

PR #547, opened 2024-05-06
Add Core-side software versions for preview 1 of v21 release.

dmkozh:

PR #548, opened 2024-05-06
Update platform p21 versions

mollykarcher:
Updating more versions, following https://github.com/stellar/stellar-docs/pull/547

### DIFF
--- a/docs/reference/software-versions.mdx
+++ b/docs/reference/software-versions.mdx
@@ -25,15 +25,14 @@ Release candidates are software releases that are also released to the [Testnet]
 | Soroban Environment | `v21.0.1` |
 | Stellar Core | `v21.0.0rc1` |
 | Soroban Rust SDK | `21.0.1-preview.1` |
-| Soroban CLI | `` |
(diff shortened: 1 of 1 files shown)


docs/reference/software-versions.mdx
```diff
@@ -25,15 +25,14 @@ Release candidates are software releases that are also released to the [Testnet]
 | Soroban Environment | `v21.0.1` |
 | Stellar Core | `v21.0.0rc1` |
 | Soroban Rust SDK | `21.0.1-preview.1` |
-| Soroban CLI | `` |
-| Soroban RPC | `` |
-| Stellar Horizon | `` |
-| Stellar Friendbot | `` |
```
mollykarcher, line 31: This delete is intentional. It never gets updated so I don't think it's necessary to list here

leighmcculloch: In the quickstart image where it gets updated we also just use the Horizon version tag anyway.

anupsdf:
LGTM

PR #555, opened 2024-05-09
embed network fee stats using Canvas

ElliotFriend:
I've placed it at the top of the page, but maybe it fits somewhere else better? Also, took a stab at writing a brief description, but I'm sure it could be improved.


janewang:
HI @ElliotFriend thank you for putting together this PR so quickly. 

Perhaps we could add the inclusion current fee charts to right above this: https://developers.stellar.org/docs/learn/fundamentals/fees-resource-limits-metering#set-the-highest-fee-youre-comfortable-paying

What do you think?

janewang (approved):

PR #557, opened 2024-05-09
add href for data lineage on references inside hubble

srhgeorgia:
Add a category on sidebarsNetwork.js, with the link for the dbt docs. 

Network > Hubble > References > Data Lineage.


sidebarsNetwork.js
```diff
@@ -24,6 +24,17 @@ const sidebars = {
           type: "autogenerated",
           dirName: "hubble",
         },
+        {
+          type: 'category',
+          label: 'References',
```
ElliotFriend, line 29: Can we change `References` to `Reference` (singular)? It's a bit more consistent with some of our other menus/sidebars.

PR #568, opened 2024-05-13
Add a table of hardware requirements for various node types

ElliotFriend:
I've collected info for core, horizon, and rpc nodes. I used the following sources for this info:

- core node: https://developers.stellar.org/network/core-node/admin-guide/prerequisites#compute-requirements
- horizon: https://developers.stellar.org/network/horizon/admin-guide/prerequisites (includes the api server, and the postgresql server)
- soroban rpc: https://github.com/stellar/helm-charts/blob/main/charts/soroban-rpc/values-pubnet.yaml (I tried to use the helm chart and its comments)

Should I include details for anchor platform and sdp?

I'm not entirely sure the best way to evaluate which cloud SKU works best, I took a stab, but I could use some recommendations for how to find the right fit. For example,

- many of the cloud vm skus don't include storage, but it can be configured through whatever related service. is that acceptable for some of these workloads?
- i tended toward the "general computing" skus, since I didn't know if memory-, cpu-, storage-optimized was better for any of them. Would love some insight on that.

I also added some sidebar config that should be helpful for readers of this page (and perhaps the network overview page, too)


ElliotFriend:
Here's the relevant preview page: https://developers-pr568.previews.kube001.services.stellar-ops.com/network/requirements

briwylde08:
Love it! Question. Is "Node Hardware Requirements" the right language? I associate "Node" with Core Node but not necessarily with Horizon.

briwylde08:
Also, there are existing hardware requirements in the Admin Guides for these various services. For example: https://developers-pr568.previews.kube001.services.stellar-ops.com/network/soroban-rpc/admin-guide#hardware-requirements. Should we remove these?

network/requirements.mdx
```diff
@@ -0,0 +1,23 @@
+---
+title: Node Hardware Requirements
+---
+
+Here you can find the requirements for running the various types of nodes the operate the Stellar network. We offer some modest recommendations for each type of resource, as well as a generally applicable SKU for some popular cloud computing providers. These may require customizations or adjustments, depending on your specific use-case.
+
+| Node Type | CPU | RAM | Disk | AWS SKU | Google Cloud SKU |
+| --- | --- | --- | --- | --- | --- |
+| Core Validator Node | 8x Intel Xeon @ 3.4 GHz | 16 GB | 100 GB NVMe SSD | [c5d.2xlarge] | [n4-highcpu-8] |
```
anupsdf, line 9: AWS might have updated their site since [c5d.2xlarge](https://aws.amazon.com/ec2/instance-types/c5/) is showing as 200 GB NVMe SSD. 
With BucketlistDB turned on, we had seen about 30 GB drop in storage usage. But we can continue to recommend c5d.2xlarge.

rice2000 (approved):

PR #571, opened 2024-05-14
Update ingestion filtering page to document new logic

aditya1702:
We recently made a change to the ingestion filtering logic: https://github.com/stellar/go/pull/5303. This PR updates the docs to reflect that.


network/horizon/admin-guide/ingestion-filtering.mdx
```diff
@@ -19,6 +19,7 @@ Filtering feature operates during ingestion in **live** and **historical range**
 
 Some key aspects to note about filtering behavior:
 
+- If both asset and account filters are configured, transactions are stored if it matches atleast one filter.
```
sreuland, line 22:
```suggestion
- If both asset and account filters are configured as enabled and each filter is provisioned at least one white list rule, then transactions are stored on database when any single white list rule from either filter matches on the given transaction.
```

JakeUrban (commented):
nit: minor edits

JakeUrban (approved):

Issue #509, opened 2024-04-18
Document consensus expectations around closeTime


marta-lokhova:
Add more detail about accuracy of [timestamps](https://developers.stellar.org/docs/learn/encyclopedia/ledger-headers#close-time) produced by consensus

marta-lokhova:
Hi @janewang, thanks for the PR! I think it'd be great to clarify expectations on accuracy of close time (I should have linked the original discussion that prompted this issue: https://stellarfoundation.slack.com/archives/C9H5JTW1J/p1713392041720139). Close time is selected by individual validators during voting, so it might be _slightly_ inaccurate depending if the clock on the validator has drifted. There is also a specific "validity window" that validators use to validate close time from other nodes, that I think is worth documenting.

janewang:
@marta-lokhova Would you be able to take over this ticket and add the information described? It was not clear from the ticket that this information is what is looked for.

marta-lokhova:
Of course, I wasn't expecting you guys to deal with it (hence I assigned it to myself)

janewang:
@marta-lokhova Are you able to work on this for us? Thanks in advance!

janewang:
@marta-lokhova Are you able to assist with closing this one?

marta-lokhova:
Hi @janewang, I'm planning to work on this in this sprint.

PR #572, opened 2024-05-15
Add a note about accuracy of close time

marta-lokhova:
Resolves https://github.com/stellar/stellar-docs/issues/509

### DIFF
--- a/docs/learn/encyclopedia/ledger-headers.mdx
+++ b/docs/learn/encyclopedia/ledger-headers.mdx
@@ -34,7 +34,7 @@ Hash of the transaction set applied to the previous ledger.
 
 ### Close time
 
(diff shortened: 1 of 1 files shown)


marta-lokhova:
@janewang PTAL

janewang:
@marta-lokhova I think we should specify our treatment of leap seconds, wdyt?

marta-lokhova:
Hi @janewang, the reason I left that off is because we specify that it's a UNIX timestamp, which by definition excludes leap seconds (so it's not unique to Stellar). But if you think we should keep the definition of UNIX system time, I can add it back. Lmk!

ElliotFriend:
For whatever it's worth, I was thinking the same as @marta-lokhova when I read the changes. Specifying it's a Unix timestamp "automatically" includes the fact that leap-seconds are excluded from the count. For anyone who doesn't already know that, I don't imagine they would _care_ to know it Hehehe

janewang (approved):

PR #576, opened 2024-05-15
Fix the "optional" part of `simulateTransaction`'s `stateChanges` list

Shaptic:
I forgot about https://github.com/stellar/soroban-rpc/pull/120#discussion_r1548753825

### DIFF
--- a/static/openrpc.json
+++ b/static/openrpc.json
@@ -1162,14 +1162,14 @@
                       "string",
                       "null"
                     ],
-                    "description": "Base64 or null - `LedgerEntry` state prior to simulation"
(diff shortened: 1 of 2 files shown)


psheth9:
Why before and after fields are coming as required on preview: 

image

static/openrpc.json
```diff
@@ -1139,7 +1139,9 @@
                 "type": "object",
                 "required": [
                   "type",
-                  "key"
+                  "key",
+                  "before",
```
psheth9, line 1143: These are optional now right?

Shaptic: No, see link in description: they're always present, sometimes null

psheth9 (approved):

PR #587, opened 2024-05-21
Update stellar-asset-contract.mdx so it recommends Soroban CLI over SDK

gglucass:
In my experience this was much easier to figure out with the Soroban CLI than the wall of words that accompanied the SDK approach.


ElliotFriend (commented):
🎉  Hi, @gglucass!! 🎉 

Thanks so much for this contribution! I really like the suggestion. It makes this section a lot more useable and friendlier for devs.

My only feedback is a pretty small nitpick, but otherwise this seems like a really great change.

Thanks again!

gglucass:
> My only feedback is a pretty small nitpick, but otherwise this seems like a really great change.

Thanks @ElliotFriend - made the change :-)

Julian-dev28:
@gglucass Thanks for the contribution!

Issue #299, opened 2024-02-08
`setOptions` operation on "List of Operations" page links to "Accounts" page unhelpfully


ElliotFriend:
The description of the [`setOptions` operation on the "List of Operations" page](https://developers.stellar.org/docs/fundamentals-and-concepts/list-of-operations#set-options) directs users to the [Accounts section](https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/accounts) when describing both the `setFlags` and the `clearFlags` parameter.

The problem is the accounts section doesn't really describe the flags, as the list of operations page suggests. You can follow a trail of links to _eventually_ get to the information you need, but it's a bit confusing and easy to get lost along the way. Here is where I _think_ a typical user might go through this process.

- List of Operations - "refer to the Accounts section" links to: https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/accounts
- Accounts - lists "Flags" with a link to learn more about it: https://developers.stellar.org/docs/glossary/#flags
- Glossary "Flags" entry - There's a very brief description of what the flags can do. Provides a link to "Controlling Access to an Asset section": https://developers.stellar.org/docs/glossary/issuing-assets/control-asset-access#controlling-access-to-an-asset-with-flags
  - _Note:_ This actually leads to a `404` error, but most of the returned search suggestions link to some part of this page: https://developers.stellar.org/docs/issuing-assets/control-asset-access.
- The "Asset Design Considerations" page does indeed provide a lot of helpful information about account and trustline flags. However, getting to that more useful information is a full four clicks for a developer to navigate, plus a `404` page in the mix.

I think we should:

- [x] Fix the `404` link on the glossary entry for "Flags." That's just a no-brainer. (accidentally solved in #354)
- [ ] The List of Operations page promises `For details about the flags, please refer to the Accounts section.` and it feels like this is too long of a trail to get to the actual useful information a developer might be looking for. We should link directly from the `setOptions` operation to the relevant part of the "Asset Design Considerations" page and take out all the intermediate steps.

PR #592, opened 2024-05-28
Re-write "set/clear flags" description in the list of operations to include more useful links

ElliotFriend:
added a note about using the account-level flags for asset issuers so it makes more sense to direct them to the asset design page.

Refs: #299


briwylde08 (approved):
Solid cross-referencing here.

PR #636, opened 2024-05-31
rpc: Add `getTransactions` documentation

aditya1702:
Add the endpoint page for the `getTransactions` endpoint in soroban-rpc: https://github.com/stellar/soroban-rpc/pull/174


openrpc/src/examples/Transactions.json
```diff
@@ -107,5 +107,168 @@
...
+    },
+    "GetTransactionsRequestUsingCursor": {
+        "name": "getTransactionsRequestUsingCursor",
+        "value": {
+            "pagination": {
+                "cursor": "8111217537191937",
+                "limit": 5
+            }
+        }
+    },
```
ElliotFriend, line 128: There is something particular about either a) OpenRPC in general, or b) the plugin/scripts we're using where using an `example` object for the request part of an example pairing works a little funky (screenshot below, where there's a superfluous intermediate object inside the `params` object in the request).

What you'll want to do is remove these request example objects, and create some examples for the individual parameters themselves, like `PaginationLimitTwo` in the `examples/Events.js` file (we're likely better off shuffling some stuff into an `examples/Pagination.js` file?). Then, add those parameter examples to the request in the example pairing object you've made, like in `NativeTransferEventsExample` in `EventsPairs.json`. (wow, that seems clear as mud hehe)

Screenshot 2024-05-31 at 12 11 41 PM

openrpc/src/schemas/Transactions.json
ElliotFriend, line 1: We could probably turn this into `Transaction`, since it's only describing a single one. Then, we could also use it as the "canonical" schema for responses to both `getTransactions` and `getTransaction` (provided they are indeed the same shape). 

Would the existing spec and descriptions in the `getTransaction` method return object be the same? (like the enum for a transaction's `status` (success, not_found, etc.)

openrpc/src/schemas/EventsPagination.json
ElliotFriend, line 1: Let's keep this file as Pagination, and include both the `EventsPagination` object and `TransactionsPagination` object in it. We could probably re-use a single `Cursor` ref between the two of them, as well.

openrpc/src/examplePairingObjects/GetTransactionsPairs.json
```diff
@@ -0,0 +1,22 @@
+{
+  "SuccessfulGetTransactionsRequest": {
+    "name": "Successful getTransactions Request",
+    "description": "Query a list of transactions starting from the given startLedger.",
+    "params": [
+      { "$ref": "#/components/examples/GetTransactionsRequest" }
```
ElliotFriend, line 6: This is where you'll probably want to use a couple of `$ref` objects for starting ledger, pagination with cursor, etc.

ElliotFriend (commented):
@aditya1702 This is amazing! Thanks!!

Added a couple notes about OpenRPC stuff.

openrpc/src/schemas/Transactions.json
aditya1702: @ElliotFriend I also thought the same but the problem here is that they both have different return schemas so the `Transaction` object will not work for both. We need 2 different schema examples.

openrpc/src/examples/Transactions.json
aditya1702: Ahh I did not see this. Thanks for spotting this issues. I will refactor it according to your suggestions.

janewang:
Hi @aditya1702 
The PR looks good. Could we please 
- rename all instances of soroban-rpc to RPC because we are going to do a rename soon, so I'd love to stop adding to docs references to the soroban-rpc

openrpc/src/methods/getTransactions.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "getTransactions",
+  "summary": "returns a list of transactions with their details",
+  "description": "The getTransactions method returns details of a list of transactions starting from the user specified start ledger or cursor. The soroban-rpc system maintains a restricted history of recently processed transactions, with the default retention window set at 1440 ledgers, approximately equivalent to a 2-hour timeframe. For private soroban-rpc instances, it is possible to modify the retention window value by adjusting the transaction-retention-window configuration setting. Due to this limited history, clients should make sure that the requests for transactions fall within the ledger range stored by their rpc instance.",
```
janewang, line 4: We are going to maintain a history of 7 days now.

Shaptic: We can only do that once the work to back events by a DB is complete, because otherwise you will have different views of history in memory vs in the database.

aditya1702: @janewang Yeah the 7 days window only goes into effect once all db changes are complete.

openrpc/src/schemas/Transactions.json
ElliotFriend: That makes sense. I might try, down the road, to "massage" the two a bit so there's less duplication to keep track of. But, that's not a blocker for this PR

openrpc/src/methods/getTransactions.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "getTransactions",
+  "summary": "returns a list of transactions with their details",
+  "description": "The getTransactions method returns details of a list of transactions starting from the user specified start ledger or cursor. The RPC system maintains a restricted history of recently processed transactions, with the default retention window set at 1440 ledgers, approximately equivalent to a 2-hour timeframe. For private RPC instances, it is possible to modify the retention window value by adjusting the transaction-retention-window configuration setting. Due to this limited history, clients should make sure that the requests for transactions fall within the ledger range stored by their rpc instance.",
```
Shaptic, line 4: Let's rephrase this to be more generic and read more like documentation rather than an operator guide, and especially not distinguishing between public/private instances (since we don't run one).

```suggestion
  "description": "The `getTransactions` method return a detailed list of transactions starting from the user specified starting point that you can paginate as long as the pages fall within the history retention of their corresponding RPC provider.",
```

(Just a suggestion, feel free to use your own words.)

openrpc/src/schemas/Cursor.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Cursor": {
+    "title": "cursor",
+    "type": "string",
+    "description": "A string ID that points to a specific location in a collection of responses and is pulled from the `paging_token` value of a record. When a cursor is provided Soroban-RPC will _not_ include the element whose id matches the cursor in the response. Only elements which appear _after_ the cursor are included."
```
Shaptic, line 5: Let's give full details, e.g. referencing TOIDs, plus minor rephrasing suggestion (e.g. matching verb tenses):

```suggestion
    "description": "A unique identifier (specifically, a [TOID](stellar.org/protocol/sep-35)) that points to a specific location in a collection of responses and is pulled from the `paging_token` value of a record. When a cursor is provided, RPC will _not_ include the element whose ID matches the cursor in the response: only elements which appear _after_ the cursor will be included."
```

openrpc/src/schemas/Pagination.json
```diff
@@ -1,16 +1,28 @@
...
+    },
+    "TransactionsPagination": {
+        "type": "object",
+        "properties": {
+            "cursor": {
+                "$ref": "#/components/schemas/Cursor"
+            },
+            "limit": {
+                "type": "number",
+                "description": "The maximum number of records returned. The limit for [getTransactions](https://developers.stellar.org/network/soroban-rpc/getTransactions) can range from 1 to 200 - an upper limit that is hardcoded in RPC for performance reasons. If this argument isn't designated, it defaults to 50."
```
Shaptic, line 23: Too detailed for public API docs

```suggestion
                "description": "The maximum number of records returned. For [getTransactions](https://developers.stellar.org/network/soroban-rpc/getTransactions), this ranges from 1 to 200 and defaults to 50."
```

(Isn't there a `default` for OpenAPI specs?)

openrpc/src/schemas/Transactions.json
```diff
@@ -0,0 +1,45 @@
...
+  "Transactions": {
+    "type": "object",
+    "properties": {
+      "status": {
+        "type": "string",
+        "description": "Indicates whether the transaction was successful or not."
+      },
+      "applicationOrder": {
+        "type": "number",
+        "description": "The index of the transaction among all transactions included in the ledger."
```
Shaptic, line 11: 0-based or 1-based?

openrpc/src/schemas/Transactions.json
```diff
@@ -0,0 +1,45 @@
...
+        "description": "A base64 encoded string of the raw TransactionResult XDR struct for this transaction."
+      },
+      "resultMetaXdr": {
+        "type": "string",
+        "description": "A base64 encoded string of the raw TransactionMeta XDR struct for this transaction."
+      },
+      "diagnosticEventsXdr": {
+        "type": "array",
+        "description": "(optional) A base64 encoded slice of xdr.DiagnosticEvent. This is only present if the transaction was not successful."
+      },
```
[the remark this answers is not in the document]
Shaptic: Also describe each string rather than at the top level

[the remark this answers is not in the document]
aditya1702: @ElliotFriend Good catch. I think that is wrong here. The diagnostic events are present even if the transaction is a SUCCESS. This is controlled by `ENABLE_SOROBAN_DIAGNOSTIC_EVENTS` in core config. I will change the description here.

openrpc/src/methods/getTransactions.json
aditya1702: No I agree. My default assumption is that these are docs for someone running their own instances. However, we should also cater to those who are depending on infra operators.

openrpc/src/schemas/Pagination.json
aditya1702: Ohh I did not know about any default OpenAPI specs. Let me check

janewang:
@aditya1702 Could we change mention of `Soroban RPC` to `RPC` please?

janewang:
Screenshot 2024-06-07 at 12 16 11 PM

These two links referred are broken.

aditya1702:
> @aditya1702 Could we change mention of `Soroban RPC` to `RPC` please?

@janewang Were you referring to the PR title? Because I already removed the `soroban-rpc` references from the documentation.

ElliotFriend (approved):
looks good to me! Thanks for the contribution!

janewang:
Thank you for documenting this endpoint!

PR #640, opened 2024-06-02
Update README.mdx

kushagra2jindal:
This is important to register the account, or else the server.loadAccount will give an error, creating confusion for the new developers.

### DIFF
--- a/docs/README.mdx
+++ b/docs/README.mdx
@@ -26,6 +26,8 @@ const {
 } = require("@stellar/stellar-sdk");
 const fetch = require("node-fetch");
 
+// Don't forget to fund these accounts, somehow. Have you met Friendbot?
+// https://developers.stellar.org/docs/tutorials/create-account#create-account
 const issuerKeypair = Keypair.random();
 const destinationKeypair = Keypair.random();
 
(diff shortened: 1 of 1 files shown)


ElliotFriend:
Hi, @kushagra2jindal! Thanks so much for the PR!

You bring up a good point about the accounts in the example being unfunded. I don't know that we really want to get into the weeds of all that here, though. This "quickstart" is more meant as an illustration, rather than a copy/paste snippet. It feels like Friendbot stuff could be left as an exercise for the reader?

Maybe we could have a comment pointing to our [create an account tutorial](https://developers.stellar.org/docs/tutorials/create-account#create-account) after the two keypairs are declared? something like:

```js
const issuerKeypair = Keypair.random();
const destinationKeypair = Keypair.random();
// Don't forget to fund these accounts, somehow. Have you met Friendbot?
// https://developers.stellar.org/docs/tutorials/create-account#create-account
```

What do you think?

kushagra2jindal:
Hi, @ElliotFriend adding the comments will solve the problem. Thanks!

briwylde08:
Hi @kushagra2jindal! Are you able to make the suggested edit? Thank you!

briwylde08 (approved):
Thank you for your contribution!

PR #644, opened 2024-06-03
Update setup.mdx

Frank-Fourier:
The soroban-cli version in this doc is outdated and will give the following error when invoking a deployed contract:

"error: xdr processing error: xdr value invalid"

As specified here an update of the CLI version is needed:

https://stellar.org/blog/developers/protocol-21-upgrade-guide

### DIFF
--- a/docs/smart-contracts/getting-started/setup.mdx
+++ b/docs/smart-contracts/getting-started/setup.mdx
@@ -84,7 +84,7 @@ The [Soroban CLI](http://github.com/stellar/soroban-cli) can execute Soroban con
 Install the [latest released version](https://github.com/stellar/soroban-cli/releases) of Soroban CLI using `cargo install`.
 
 ```sh
-cargo install --locked soroban-cli
+cargo install --locked soroban-cli --version 21.0.0-rc.1
 ```
 
 :::info


ElliotFriend:
@Frank-Fourier Thanks for the contribution!! 🎉 

Which deployed contracts are giving you this error? Probably contracts that have been deployed after the Testnet upgraded to protocol 21, I would think.

I guess this is a tricky line to straddle: even though the currently _released_ CLI is compatible with Mainnet contracts, do we tell people to upgrade to the release candidate, as it will be more useful as a developer using Testnet... 🤔

@chadoh @leighmcculloch Any thoughts on this one? I think I'd lean toward not specifying a particular version of the CLI to install, in the hopes that the instructions would be more "universal" and compatible over time and protocols. But, I'm far from an expert at cargo or Rust crates, so I'd defer to someone else's better judgment here.

leighmcculloch (approved):

leighmcculloch:
I think we could show both commands and note that right now you need the rc for testnet.

Frank-Fourier:
> @Frank-Fourier Thanks for the contribution!! 🎉
> 
> Which deployed contracts are giving you this error? Probably contracts that have been deployed after the Testnet upgraded to protocol 21, I would think.
> 
> I guess this is a tricky line to straddle: even though the currently _released_ CLI is compatible with Mainnet contracts, do we tell people to upgrade to the release candidate, as it will be more useful as a developer using Testnet... 🤔
> 
> @chadoh @leighmcculloch Any thoughts on this one? I think I'd lean toward not specifying a particular version of the CLI to install, in the hopes that the instructions would be more "universal" and compatible over time and protocols. But, I'm far from an expert at cargo or Rust crates, so I'd defer to someone else's better judgment here.

I was following the docs. I had no issues with the given install command when deploying and interacting with the token contract compiled from the soroban examples repo. However a custom contract, for example the vault contract itself or a simple hello world, will deploy successfully but will give the error when invoked. It's was fixed by updating the CLI to the new version.

chadoh (approved):
I think it probably makes sense to do this for now. We can remove it again once we're past `rc`?

ElliotFriend (approved):
thanks for the help keeping our examples working!

PR #645, opened 2024-06-03
Update smart-contract-deployment.mdx

Frank-Fourier:
The it is an issue in the vault contract's withdraw function which will incorrectly update the value of Reserve.

In this line

    let balance = get_token_balance(&e);

balance is updated AFTER the transfer of the user deposit + yield has already occurred. 

Thus the balance fetched is already the current "reserve" balance of the Vault. 

So when the following line of code executes:

put_reserve(&e, balance - amount);

the Reserve is updated to the balance of the Vault contract - the amount withdrawn by the user, so it will count the withdrawn 2 times. 

Indeed when following the test the result from "get_rsrvs" in the first scenario will be :

"-1" 

Initial balance of the Vault: 200 - amount: 100 + yield: 1 (from the transfer) - amount: 100 (from the put_reserve input). 

For further improvements the get_rsrvs should only account for the amount of reward tokens that the vault is holding which can be useful to obtain the deposited amount (vault balance - reserves).

Additionally the --wasm argument in the "invoke" command is not needed and it was removed.

### DIFF
--- a/docs/learn/migrate/evm/smart-contract-deployment.mdx
+++ b/docs/learn/migrate/evm/smart-contract-deployment.mdx
@@ -869,20 +869,20 @@ fn withdraw(e: Env, to: Address, amount: i128) -> i128 {
     // First transfer the vault shares that need to be redeemed
     let share_token_client = token::Client::new(&e, &get_token_share(&e));
     share_token_client.transfer(&to, &e.current_contract_address(), &amount);
+
+    // Calculate total amount including yield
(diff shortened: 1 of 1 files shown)


ElliotFriend (approved):
@Frank-Fourier Sorry I didn't see this until now. Thank you! You've been super-helpful with these PRs!

This is a great catch and fix! And I **so** appreciate you helping out with removing some of the leftover `--wasm` parameters we have scattered around.

Thanks again!

PR #658, opened 2024-06-07
Update MGI guide using TS wallet SDK

CassioMG:
Ticket: [[Documentation] Update MGI guide using TS wallet SDK](https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37?selectedIssue=WAL-1442)

Preview: https://developers-pr658.previews.kube001.services.stellar-ops.com/docs/building-apps/moneygram-access-integration-guide

Currently, we have this [Integrate with MoneyGram Access](https://developers.stellar.org/docs/building-apps/moneygram-access-integration-guide) guide in addition to the regular [wallet SDK SEP-24 integration guide](https://developers.stellar.org/docs/building-apps/wallet/sep24), which a lot of wallets seem to be referring to (as it covers every step of the MGI integration).

This PR updates the [Integrate with MoneyGram Access](https://developers.stellar.org/docs/building-apps/moneygram-access-integration-guide) guide to use `wallet SDK code examples`.

In addition, it also adds a link to the wallet SDK documentation in “Resources” section and creates a new “Installing the Wallet SDK” sub-section in the “Introduction” section that describes how to install the wallet SDK and links to existing documentation for more information.


docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -142,6 +157,22 @@ The following code demonstrates how to implement the application’s side of thi
 
 <CodeExample>
 
+```ts
+import { Wallet, SigningKeypair } from "@stellar/typescript-wallet-sdk";
+
+const wallet = Wallet.TestNet();
+
+// First we create an anchor object with MoneyGram's home domain.
+const anchor = wallet.anchor({ homeDomain: MGI_ACCESS_HOST });
```
Ifropc, line 166: Can you set this variable in the code please?
Maybe something like
```typescript
// Testnet
// MGI_ACCESS_HOST = "extstellar.moneygram.com"
// Main net
MGI_ACCESS_HOST = "stellar.moneygram.com"
```

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -185,10 +216,27 @@ You will need the following pieces of information:
...
+The following code can be used as a reference for implementing this logic yourself.
 
 <CodeExample>
 
+```ts
+// Use same "anchor" object from previous step.
+const { url, id } = await anchor.sep24().withdraw({
+  authToken: authToken, // Use same "authToken" string from previous step
+  withdrawalAccount: FUNDS_STELLAR_KEYPAIR.public_key,
+  assetCode: ASSET_CODE, // Should be "USDC"
```
Ifropc, line 228: I think we should follow our own guidelines from [SEP-24 guide](https://developers.stellar.org/docs/building-apps/wallet/sep24) 
```
const asset = info.currencies.find(({ code }) => code === "USDC").assetId;
```

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -185,10 +216,27 @@ You will need the following pieces of information:
...
+  lang: "en",
```
Ifropc, line 229: Should note that this field is optional

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -185,10 +216,27 @@ You will need the following pieces of information:
...
+  extraFields: {
+    amount: "<amount to withdraw / cash-out>",
```
Ifropc, line 231: Should note that this field is optional and only mandatory for custodial wallets

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -185,10 +216,27 @@ You will need the following pieces of information:
...
+  },
+});
+
+// Append the callback param for the wallet to be notified when the user has
+// completed the MGI experience and has requested to close the window.
+const urlWithCallback = `${url}&callback=postmessage`;
```
Ifropc, line 237: We advise against using postMessage approach, I think we should mention that it's being deprecated long term

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -247,20 +295,84 @@ In each case, the transaction submitted to Stellar must have a memo attached to
...
+    // runtime error, or the transaction comes back as
+    // no_market / too_small / too_large / error
+  },
+});
+
+// While the Watcher class offers powerful tracking capabilities, sometimes
+// it's required to just fetch a transaction (or transactions) once. The Anchor
+// class allows you to fetch a transaction by ID, Stellar transaction ID, or
+// external transaction ID:
+const transaction = await anchor.sep24().getTransactionBy({
```
Ifropc, line 362: Can you expand a bit more into how to use stellarTransactionId or externalTransactionId? Maybe write 3 examples

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -310,10 +422,57 @@ Code for submitting transactions to Stellar should be developed thoughtfully. Th
...
+```ts
+import { Wallet, IssuedAssetId } from "@stellar/typescript-wallet-sdk";
+
+const wallet = Wallet.TestNet();
+
+// This creates a Stellar instance to manage the connection with Horizon.
+const stellar = wallet.stellar();
+
+// Creates a stellar asset, in this case it should use USDC configuration.
+const asset = new IssuedAssetId(ASSET_CODE, ASSET_ISSUER);
```
Ifropc, line 440: Let's define on how to query asset above (see another comment)

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -310,10 +422,57 @@ Code for submitting transactions to Stellar should be developed thoughtfully. Th
...
+      .transferWithdrawalTransaction(transaction, asset)
+      .build();
+
+    // Signs it with the source (funds) account key pair
+    transferTransaction.sign(FUNDS_STELLAR_KEYPAIR);
+
+    // Finally submits it to the stellar network. This stellar.submitTransaction()
+    // function handles '504' status codes by keep retrying it until submission
+    // succeeds or we get a different error.
+    const response = await stellar.submitTransaction(transferTransaction);
```
Ifropc, line 469: Should we write an example on how to handle other errors?

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -358,6 +517,53 @@ Note that this code does not handle [path payments] or [claimable balances], two
 
 <CodeExample>
 
+```ts
+// The Wallet SDK does not support payments streaming yet so let's build
```
Ifropc, line 521: IMO it's a good functionality for wallet SDK to have 🤔

docs/building-apps/moneygram-access-integration-guide.mdx
JakeUrban: +1 on this, do we even need to recommend it? I thought MoneyGram supported not using postmessage as well.

docs/building-apps/moneygram-access-integration-guide.mdx
```diff
@@ -247,20 +295,84 @@ In each case, the transaction submitted to Stellar must have a memo attached to
 
 ### Poll Until MoneyGram is Ready
 
-Before the application can send funds or instruct the user to provide cash to a MoneyGram agent, the application should confirm with MoneyGram’s server that the transaction is ready to proceed.
+Before the application can send funds or instruct the user to provide cash to a MoneyGram agent, the application should confirm with MoneyGram’s server that the transaction is ready to proceed which is signaled by the `pending_user_transfer_start` status.
```
JakeUrban, line 298: Q: can/should we recommend the use of MoneyGram's webhooks now? When this tutorial was originally created they didn't have them. Now I would imagine this the preferred approach.

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: Sure thing. I was using Python's examples/vars as reference.

I'll set the var to make it more clear.

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: I believe the problem of not implementing the `postMessage` handler is that MGI is still displaying a red `OK/Close` button on the last screen. So if wallets don't listen for that button it'll behave like a broken button.

But one thing I just realized is that Vibrant is not appending this `&callback=postmessage` param to the url but the app still receives the `postMessage` event. So it looks to me that we don't need to ask wallets to append this param, but instead just listen to a postMessage event coming from that last red button so they know it's safe to close the webview and prevent the button from looking broken. What do you guys think?

Ifropc: Hmm so if we don't listen to it, the button will not work? I think we should describe it in the documentation then

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: yep

we have this task on the backlog: https://stellarorg.atlassian.net/jira/software/c/projects/WAL/boards/37/backlog?selectedIssue=WAL-1444

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: Nice catch.

Do we have any existing documentation explaining how to setup/integrate MGI's webhook?

docs/building-apps/moneygram-access-integration-guide.mdx
JakeUrban: We defined a migration plan with MGI to remove postmessages because of the UX complications in the event that something doesn't work. @Ifropc can you dig up the plans we had and revisit this with MGI? Feel free to loop in the Vibrant team as well.

docs/building-apps/moneygram-access-integration-guide.mdx
JakeUrban: I don't believe so, @Ifropc @nirojapanda are you familiar with any docs?

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: Thanks guys, would be nice to remove that button.

Just to clarify: Vibrant does not rely on the postMessage event anymore for a while now, users can close the webview without tapping MGI's red button. Which means MGI would just need to remove that red button and probably add some copy letting users know they can safely close the window.

Ifropc: @JakeUrban they have been stalling the plan for a very long time, it's somewhere in their backlog. We should raise this question again with their team

Ifropc: Yep the idea is to replace button with the next that says you can close the window now

docs/building-apps/moneygram-access-integration-guide.mdx
Ifropc: No, sorry, I don't think there are any docs for that. Their implementation also doesn't match the standard precisely BTW

JakeUrban: Ok, so do we still want to recommend this approach? If so, can you help create that documentation if you're familiar with the differences?

Ifropc: I think it's what Vibrant is doing: only listen to postMessage for closing button, and do NOT listen to the message returned by it, @CassioMG am I correct?

CassioMG: Correct, Vibrant only listen to postMessage for closing the button. But it only does that when the `transaction` that comes inside the event message is in a `pending` status to avoid closing the webview for ANY other random postMessage event that may happen during the interactive flow for whatever reason.
Like this:
```
    // If we get a postMessage and the status is any pending one,
    // let's interpret that as a signal to close the window and
    // take user back to the wallet experience
    if (eventMessage.transaction?.status?.includes("pending")) {
      handleClose();
    }
```

Does that make sense?

Sorry about the late response I was OOO these past days

JakeUrban: Ok, it sounds like using MGI's webhooks instead of postmessages may not be an option yet so we need to continue documenting the postmessage approach. @CassioMG if you think we need to instruct wallets to only close when the status is `pending_user_transfer_start` that is fine by me

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: @Ifropc this code appears to be broken. We can't access `assetId` directly from `info.currencies`.

I think this is the best we could do with current TS Wallet SDK:
```
    const tomlInfo = await anchor.getInfo();

    const tomlAsset = tomlInfo.currencies.find(({ code }) => code === "USDC");

    if(!tomlAsset?.code || !tomlAsset?.issuer) {
      throw new Error("Anchor does not support USDC asset or is not correctly configured on TOML file");
    }

    const asset = new IssuedAssetId(
      tomlAsset.code,
      tomlAsset.issuer,
    );
```

What do you think?

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/960e7cc8617115e256fe08e280c5c72d3592a9de

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/960e7cc8617115e256fe08e280c5c72d3592a9de

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: @Ifropc I'm wondering if we should display this same code above 👆  in our [SEP-24 guide](https://developers.stellar.org/docs/building-apps/wallet/sep24) in place of the broken `const asset = info.currencies.find(({ code }) => code === "USDC").assetId;` code. Wdyt?

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/d45f9d05bb7f78811a3ae68c7849eb6f56ea97a3

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/295515b170ec57706bff7d542ac9e2a0398b0334

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/295515b170ec57706bff7d542ac9e2a0398b0334

docs/building-apps/moneygram-access-integration-guide.mdx
CassioMG: https://github.com/stellar/stellar-docs/pull/658/commits/0f45a2dcd0cde262d3489b89fd64785877ef01cc

JakeUrban (approved):
LGTM! We can always do another iteration once @Ifropc gets back

Ifropc:
LGTM, thanks @CassioMG !

PR #659, opened 2024-06-07
Rewriting Horizon scaling documentation

mollykarcher:


network/horizon/admin-guide/scaling.mdx
```diff
@@ -3,38 +3,55 @@ title: Scaling
...
-![](/assets/horizon-scaling/Topology-2VMs.png)
+There are a few reasons you may choose to scale to multiple instances of Horizon.
 
-## Enterprise _n_-Tier
+- Horizontally scaling enables you to serve more API requests and at a faster rate
+- Redundancy enables zero downtime in the cases where Horizon requires downtime on upgrade (migrations, state rebuilds, etc)
+- Protection against potential ingestion lag, which could result in downtime for end-users
 
-This architecture services high request and data processing throughput with isolation and redundancy for each component. Scale the API service horizontally by adding a load balancer in front of multiple API service instances, each only limited by the database I/O limit. If necessary, use ALB routing to direct specific endpoints to specific request-serving instances, which are tied to a specific, dedicated DB. Now, if an intense endpoint gets clobbered, all other endpoints are unaffected.
+Multiple instances of Horizon can be configured to potin to the same database, and the ingestion process will not perform redudant work in these cases.
```
janewang, line 30: sp: `point`

janewang (approved):
Nit: sp

network/horizon/admin-guide/scaling.mdx
janewang: `redundant`

static/assets/horizon-scaling/Topology-ingestion-isolation.png
sreuland, line 1: minor nit, this sees to reference logical and physical designs at same time, i.e. the Horizon square seems to infer a physical process boundary, and at same time depict 'Captive Core' inside Horizon, which logically is good, but is it worth tweak this to highlight the physical deployment topology, i.e. 'Captive Core' is a second process on host, and 'Ingestion' is sub module in Horizon that connects to it:

topology

mollykarcher: Hmm yeah I see your point. But I think going more explicit like that complicates it without really providing much added clarity/benefit to the reader. What do you think about just actually removing the "captive core" boxes from the diagram. I think the main value I was trying to show here is that the API instances can be lighter weight in that they don't need captive core, but that's also captured in text. Having just the 3 "roles" inside Horizon will feel more symbiotic with the rest of the configuring/pre-reqs docs I thin. What do you think?

sreuland: yes, it makes the point as-is, just wanted to check on that aspect, thanks for consideration!

PR #661, opened 2024-06-07
Soroban RPC: Add document for getVersionInfo RPC

psheth9:


janewang:
Nit: My only suggestion is to define what captive core is in the description. It is a stellar-core that's managed by another piece of software. Maybe you could describe it better since it is a nomenclature that is only meaningful to some people, not all developers in the ecosystem.

janewang (approved):

PR #673, opened 2024-06-12
fix nginx config.

overcat:
1. `location / { deny all; }` conflicts with the later defined `location / {...}` and I think it is unnecessary, we can remove it.
2. The usage of `try_files` is incorrect, we need to correct it. https://nginx.org/en/docs/http/ngx_http_core_module.html#try_files

### DIFF
--- a/docs/validators/admin-guide/publishing-history-archives.mdx
+++ b/docs/validators/admin-guide/publishing-history-archives.mdx
@@ -58,9 +58,6 @@ server {
 
   server_name history.example.com;
 
-  # default is to deny all
-  location / { deny all; }
-
   # do not cache 404 errors
   error_page 404 /404.html;
   location = /404.html {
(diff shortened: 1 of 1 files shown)


mwtzzz (changes requested):
Please explain why you think `deny all` is unnecessary? And if it's interfering with another rule down below, then I think we should move it below that rule instead of eliminating it.

overcat:
> Please explain why you think `deny all` is unnecessary? And if it's interfering with another rule down below, then I think we should move it below that rule instead of eliminating it.

Hi @mwtzzz, Nginx does not allow adding duplicate locations, so moving it below is not feasible. Instead, any files in the `/mnt/xvdf/stellar-core-archive/node_001/` folder are not private and can be made public.

briwylde08:
Hi @mwtzzz! Are you okay reviewing this PR?

mwtzzz (approved):
LG

briwylde08 (approved):

briwylde08:
Hi @overcat! Are you able to fix the conflict? Then we're good to merge!

overcat:
> Hi @overcat! Are you able to fix the conflict? Then we're good to merge!

Hi @briwylde08, resolved.

PR #683, opened 2024-06-14
Correct the type of the `extras` field.

overcat:


Shaptic (approved):

ElliotFriend:
@overcat We just merged in #686 that reorganizes a lot of files which were located in the `/network` directory. Can you resolve the conflict that popped up?

ElliotFriend:
Thanks for taking care of that!

PR #686, opened 2024-06-14
Reorganize "network" section content

ElliotFriend:
We're going to:

- [x] move the core node docs out into their own navbar item
  - [ ] (don't forget to update the links in the core-node gh docs page)
- [x] move the disbursement/anchor platforms into the tools section
- [x] rename the navbar of what's left (horizon, rpc, hubble) to "Data"


briwylde08:
This is looking so great! My only question is: Node Hardware Requirements seems kind of hidden and also contains Core Validator Node reqs even though it's in the Data section. Would this section make more sense somewhere else? Maybe Technical Reference (Reference)?

ElliotFriend:
@briwylde08 I was wondering about putting it in the "Reference" section, too! I talked with Justin about it today, and my plan at the moment is:

- Leave it where it is, for now
- Create some follow-up PRs later on to...
  1. isolate and (strategically) duplicate core node (for example) h/w requirements into the validators admin guide (and similar with horizon, rpc, etc.)
  2. add links to each relevant admin guide in the table so people can know exactly what they need to get started, and then get to the instructions quickly.
- (perhaps? down the road?) move it somewhere else, depending on if/what feedback is given.

What do you think?

briwylde08:
Makes sense to me!

briwylde08 (approved):
Amazing!!

PR #692, opened 2024-06-17
Update dependency

Ifropc:
Updating `eslint-plugin-jsdoc` dependency: current version doesn't support node v20


ElliotFriend (approved):
great catch! thanks!!

PR #693, opened 2024-06-18
typo fix in cli guide

rahul-soshte:
A simple typo fix. The word 'isntalled' should be 'installed'. Nothing much apart from that.


briwylde08 (approved):
Good catch!

Issue #429, opened 2024-04-04
Docs Navigation Issue


janewang:
### What version are you using?

I think this is especially confusing with the how-to guides that have sub sections that are reflected in the breadcrumbs but not the sidebar, 

![image](https://github.com/stellar/stellar-docs/assets/734336/ee23388a-f900-4503-ad9f-cc9603d3af49)

![image](https://github.com/stellar/stellar-docs/assets/734336/bcc00ce1-f976-46f9-b854-c7df2ecd71e1)

Additionally: Some further issues with navigation. The categories listed in How-To guides are not clickable, though you can navigate to these pages through breadcrumbs and search, but these can be pretty barren. See example of the State Expiration how-to guides category page https://developers.stellar.org/docs/smart-contracts/guides/archival

briwylde08:
I think it could make sense to make the main categories clickable to follow the breadcrumbs but I would argue that putting all the How-To Guides in the sidebar would be super messy. Especially since we're aiming to create way more How-To Guides in the future. I don't know if putting the How-To Guides in the sidebar is worth how chaotic it will look! I would find it overwhelming coming into the docs.

briwylde08:
Maybe we can fold into docs merge v2

PR #704, opened 2024-06-20
show category pages for how-to guides in sidebar

ElliotFriend:
This change enables a dropdown on the "How-To Guides" sidebar category that only shows the further sidebar categories in the how-to guides "parent" category. This helps to orient the user while they are exploring various guides.

I haven't put the links directly onto the [how-to guides "parent" category page](https://developers.stellar.org/docs/smart-contracts/guides). Doing so made every piece of text on the page blue, and it was quite disorienting and hard to find where I would want to go. I'm open to doing that in the preview, if there's interest from others in seeing how it looks.

Refs: #429


ElliotFriend:
Here is a page in the preview that demonstrates pretty well how it'll look: https://developers-pr704.previews.kube001.services.stellar-ops.com/docs/smart-contracts/guides/events/ingest

janewang (approved):

ElliotFriend:
Jane pointed out that there are a [couple](https://developers-pr704.previews.kube001.services.stellar-ops.com/docs/smart-contracts/guides/conversions) [pages](https://developers-pr704.previews.kube001.services.stellar-ops.com/docs/smart-contracts/guides/fees) that aren't properly showing the category indexes. I'll try and figure out what's going on.

@briwylde08 What do you think? Does the sidebar feel too cramped? I _think_ it still feels ok, but if we add many more categories like this, I think it would feel not ok pretty dang quickly hehe

PR #708, opened 2024-06-20
Add info status of testnet and mainnet status

janewang:
This came back from an interviewee who provided feedback on not being able to find the status of testnet and mainnet.


briwylde08 (approved):
Woot!

PR #709, opened 2024-06-20
add sdk examples to send transaction

Myestery:
@briwylde08 
Added SDK examples to sendTransaction Page to close https://github.com/stellar/stellar-docs/issues/622

PR #720, opened 2024-06-25
Prevent OpenRPC-generated pages from using the `@theme/ApiItem` component

ElliotFriend:
As pointed out in #709 ([here](https://github.com/stellar/stellar-docs/pull/709#discussion_r1651473483) is the exact comment), the `@theme/ApiItem` component from the `docusaurus-theme-openapi-docs` package (as well as our swizzled `/src/theme/ApiItem/Layout` component) adds some styling/layout that we don't really want to see in the pages generated using the `open-rpc-docs-react` component. Namely, as pointed out in #709, any `<CodeBlock>` components are floated to the right-hand side of the screen (wrapping the code block component in a `<div>` seems to do the trick to workaround it, btw).

This PR aims to ensure that the URL of a page rendered with `@theme/ApiItem` doesn't include `rpc`, and will use the default `@theme/DocItem` component, instead.

### DIFF
--- a/docs/data/rpc/api-reference/methods/getTransaction.mdx
+++ b/docs/data/rpc/api-reference/methods/getTransaction.mdx
@@ -12,7 +12,6 @@ import { RpcMethod } from "@site/src/components/RpcMethod";
 
 The example above is querying details of a transaction using RPC methods directly. If you are using the Stellar SDK to build applications, you can use the native functions to get the same information.
 
-<div>
 <CodeExample>
 
 ```python
@@ -73,4 +72,3 @@ public class GetTransactionExample {
 ```
 
 </CodeExample>
-</div>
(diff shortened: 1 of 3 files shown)


ElliotFriend:
Relevant preview pages: The [`getTransaction`](https://developers-pr720.previews.kube001.services.stellar-ops.com/docs/data/rpc/api-reference/methods/getTransaction) and [`sendTransaction`](https://developers-pr720.previews.kube001.services.stellar-ops.com/docs/data/rpc/api-reference/methods/sendTransaction) pages are the only two pages which should be affected by this PR (at the moment). I've also remove the `<div>` workarounds on both of those pages.

briwylde08 (approved):
[🙏](https://emojipedia.org/folded-hands)

PR #727, opened 2024-06-27
Update stellar cli command in rpc admin guide

janewang:


docs/data/rpc/admin-guide.mdx
```diff
@@ -140,7 +140,7 @@ Replace `testnet` in that command with the name of your choice if you want to le
 The `alice` identity suggested for your Standalone network will still work here, since it's just a public/private keypair, but you'll need to remember to fund it for the Testnet network:
 
 ```bash
-curl "https://friendbot.stellar.org/?addr=$(soroban keys address alice)"
+curl "https://friendbot.stellar.org/?addr=$(stellar keys address alice)"
```
fnando, line 143: I think we can remove all calls to friendbot, because we're doing that already by default when calling `stellar keys generate`.

fnando (approved):

PR #731, opened 2024-06-28
Update section name to include operations

janewang:
I was having issue finding the 26 operations and I heard the same issue from another developer recently. I think the section header should indicate this section includes operations and transactions. Right now, the list of operations is buried under a section titled `Transactions` which makes it very difficult to find.


briwylde08 (approved):
Good point!!

PR #741, opened 2024-07-01
Update README to use more informative names

janewang:
Section title. Some users find it hard to know which section we are in because the breadcrumb is too small. I updated the section title in this PR.


briwylde08 (approved):
LGTM!

PR #745, opened 2024-07-02
Fix algolia crawler errors

ElliotFriend:
noticed a bunch of crawling errors in the algolia console. This should help fix them.

An incidental benefit is that this PR includes a patch for the `docusaurus-plugin-openapi-docs` plugin to enable our multi-instance setup.


briwylde08 (approved):
Inspired.

PR #772, opened 2024-07-10
create a Payments tag in horizon OpenAPI spec

ElliotFriend:
To give front more of a "front set" to the `/payments` endpoint, lets create its own tag/category.


briwylde08 (approved):
Inspired.

briwylde08 (approved):
Inspired.

PR #785, opened 2024-07-11
fix tx meta deprecation notice links to be relative filenames

ElliotFriend:
The links added in #782, aren't relative filepaths, and could lead to some build errors.


briwylde08 (approved):
Nice.

janewang (approved):

PR #789, opened 2024-07-12
Rename reference section to networks

leighmcculloch:
### What
Rename the "Reference" section to "Networks".

### Why
The section contains information about what deployed networks exist, what versions of software are deployed to those networks, and the resource limits and fees that are configured on those networks. All the sections within the reference section are about networks.

Also, the name "reference" is functional, it describes the types of docs are within, but doesn't provide any signal for what the content of those docs will be. That means that when someone thinks about "networks" they first need to categorise functionally to know that they'd find the information about networks in the reference section.

Most of the docs are categorised by domain rather than by function, which is easier to navigate. The reference section and the learn section seem to be the only ones not categorised by domain.


janewang (approved):

ElliotFriend:
Let's hold off on this, while we discuss further. https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1720748438913269

ElliotFriend:
There are some pretty compelling points raised here, and in the slack thread. There's certainly some things that could be done to improve how this information is presented. I'm gonna close this particular PR, for a couple reasons:

1. The "big bad migration v2" is still pretty fresh, and the site needs a little time to "settle." We want to have a really solid idea from the analytics folks about user behavior, redirect effectiveness, seo, etc. So, we're trying really hard to avoid any "structural" changes to the site for a while.
2. The word "Network**s**" in the top-nav feels like a bit of a misnomer. As someone familiar with developing on Stellar, I can understand it to mean what you've described. But, a dev (especially a new[er] dev) could just as easily get mixed up and assume it's referring to something akin to Cosmos' "internet of blockchains," or ethereum L2s, etc. I'm not opposed to using a domain-centric approach to group this set of information together (and possibly other info, as well). I just don't think "networks" is the right label to apply here.

We'd love to keep working on a specific plan for how we can make the improvements we want to see for this info. That'll make it much easier to "hit the ground running" down the road.

P.S. There's a non-trivial part of me that hates to close this PR, because you did **literally everything** to make it ready to go. You fixed links, fixed the sidebars, kept up-to-date with `main`, and you _even added the nginx redirects_! The work you put into this PR deserves an A++! Thank you @leighmcculloch!!

janewang:
I don't think this warrants closing this PR @ElliotFriend. We could leave this PR open until we decide what to do about reference. As I said to you, reference is very vague as to what the references is referencing to_. Only nearly all content is referencing to network attributes.

ElliotFriend:
@janewang 

I'll have to disagree, and say closing this PR is warranted. Even if we're still discussing what we want to do.

First, it's likely that during the "baking period" this PR will become very out-dated and merging down the road will be more difficult. The easier route will be to create a new PR once we do have a decision made.

Second, as I've said, even if `reference` isn't the best term to use here, I don't think `networks` is a better one. It gives the impression that there is more than one Stellar network, and a dev will have to choose which one to build on.

If we do become convinced, or if there's some data down the road that could support that change, we could even re-open this PR at that time. In the meantime, though, I don't think it makes sense to keep a "zombie" PR alive while we wait.

briwylde08:
I second @ElliotFriend here. While `Reference` may not be the perfect section title, I don't know if `Networks` is better. 

I'd also like to second the opinion that we're at a place where we should let the docs organization sit for a bit. We've made _a lot_ of structural changes over the last several months and it would be great to be able to truly gather some data to make decisions down the line. My two cents!!

janewang:
> First, it's likely that during the "baking period" this PR will become very out-dated

This sentence doesn't make sense to me at all. I thought the "baking period" is for the docs to stay put, you are now saying that this PR will become out-dated. This logically doesn't fly.

janewang:
We are keeping the doc structure as it during the "baking period" and I do not see how that this PR could become out-dated.

briwylde08:
While we aren't making large structural changes in the near term, there are smaller section changes that could happen to make this PR out of date. We can keep it open as we discuss this subject, but let's turn this PR into a draft.

ElliotFriend:
Good call, @briwylde08. A draft is the perfect approach here.

leighmcculloch:
> even if `reference` isn't the best term to use here, I don't think `networks` is a better one. It gives the impression that there is more than one Stellar network, and a dev will have to choose which one to build on.

There is more than one Stellar network, there's mainnet (pubnet), testnet, futurenet, local networks, ... This is an established paradigm. For example:
- https://ethereum.org/en/developers/docs/networks/

briwylde08 (approved):

PR #792, opened 2024-07-12
break out token interface code into a markdown partial

ElliotFriend:
create a token interface page in the reference section


ElliotFriend:
Dang. [This page] isn't supposed to be generated... 🤔 I'll have to figure out what went wrong

PR #754, opened 2024-07-04
Integrate a Stellar asset contracts

tupui:
closes #603 
Closes #601

PR #828, opened 2024-07-25
Fix token handling in SAC tests

tupui:
Follow up on #754 

Sorry for some reason I missed these.


tupui:
cc @ElliotFriend, I had a mistake on the `mint`. Sorry about that.

ElliotFriend (approved):
Great catch! Thanks!

PR #835, opened 2024-07-29
Anchor Platform: Unify Platform Transaction schemas

philipliu:
This PR updates the platform transaction schemas so that all SEPs have the same fields.


openapi/anchor-platform/bundled.yml
```diff
@@ -431,14 +461,17 @@ components:
             - withdrawal
         status:
           $ref: '#/components/schemas/StatusSEP24'
+        type:
```
Ifropc, line 464: Should it not be deprecated for SEP-6? Can't we get it with SEP-12 as a financial field?

Ifropc (approved):

openapi/anchor-platform/bundled.yml
philipliu: The business would need the type to know what financial fields to request from the user, right?

Ifropc: Yes, but now we have the idea of dynamic KYC. We can request financial field type first, and then request the remaining fields depending on it

philipliu: We discussed this offline. AP does not need to store the `type` of the transaction as wallets make SEP-12 requests with `transaction_id` as well as `type`. This would require a protocol change in potentially SEP-6 and SEP-9 so we will keep passing this field for now and investigate changing this in 4.0.

PR #842, opened 2024-07-31
Update RPC retention window to 7 days

janewang:
Also add max number of ledger keys in getLedgerEntries


Shaptic (approved):
Mergeable once getEvents support is released!

ElliotFriend:
> Mergeable once getEvents support is released!

@Shaptic When will that be ready?

Shaptic:
@ElliotFriend should be in the next RPC release :)

Issue #603, opened 2024-05-29
Create a guide explaining how to create a restoration footprint manually to restore archived data


briwylde08:
**Get awarded:** up to $200 worth of XLM
**Section:** How-To Guides - State Archival: https://github.com/stellar/stellar-docs/tree/main/docs/smart-contracts/guides/archival

**Guide could include:**

- Explanation of state archival functionality
- Code walkthrough
- Written explanation of how the guide works

KodeSage:
Hi @briwylde08 , I would love to work on this

briwylde08:
Great! You got it!

tupui:
Potentially closed by #754 😃

tupui:
Oh sorry I misread and it was a typo. 613 it was and not this 😅

tomita-militaru:
Hello @briwylde08 - may I work on this one?

briwylde08:
Heya @tomita-militaru! Super glad you'd like to participate! Let's give @KodeSage a chance to respond if he's still planning on working on this before you snag it. @KodeSage -- are you still working on this issue (and I guess another question is if you will be able to submit by the deadline of Wednesday next week?)

KodeSage:
Hi @briwylde08 , I just create a PR on it.

briwylde08:
Hi @tomita-militaru! Feel free to take this issue!

tomita-militaru:
@briwylde08 thank you!

tomita-militaru:
Hi @briwylde08 - I've just submitted a PR 🤞

briwylde08:
Thank you! We'll get it assigned and reviewed [👏](https://emojipedia.org/clapping-hands)

briwylde08:
Hi @tomita-militaru! 

Congratulations! Your contribution to Deep Dive Docs Bounty has been accepted! [🎉](https://emojipedia.org/party-popper/)
Your submission may qualify for an award of $200 worth of XLM! One last step: please fill out [this form](https://airtable.com/appOfFtPsiMxd27UZ/pagxH1WRxIpTFRGec/form) completely and accurately. Once we review the completed form and everything checks out, we will let you know the status of your potential award!

Please let us know if you have any questions.

All the best,
Bri

tupui:
Youhou 🥳 congrats Tomi 👏

tomita-militaru:
Thank you!

tomita-militaru:
Hi @briwylde08 - I've just filled in the forms, thanks!

PR #845, opened 2024-08-01
chores: manually create restoration footprints

KodeSage:
Bounty Submisson for #603 

@briwylde08


Shaptic:
The language here appears to be AI generated and the content itself completely incorrect: valid `createRestorationFootprint` code should have nothing to do with paginating transactions from Horizon.

Issue #404, opened 2024-03-28
Provide code examples to make http requests using Stellar SDK in docs


janewang:
### What problem does your feature solve?
Missing examples on using the SDK

### What would you like to see?
For RPC documentation and Horizon documentation: we could switch all our Javascript and Python code examples to make http requests using a Stellar SDK (as opposed to general http libs like requests) and show encoding/decoding of XDR (as opposed to using opaque b64 strings)

ElliotFriend:
For Horizon, this is much more easily done. I recently learned about the [`x-codeSamples` openapi extension](https://redocly.com/docs/api-reference-docs/specification-extensions/x-code-samples), and it will enable us to bake that right into the API specfile, to be generated at build/deploy time.

I don't _think_ there is an openrpc equivelant to this, though. So, for RPC we might be out of luck for now

chadoh:
@janewang @ElliotFriend I created #860 to make it easier to maintain our own fork of the `open-rpc-docs-react` package, which is how we add those tabs to show various ways of doing the RPC calls. However, my fork does not modify the tabs or examples yet.

I'm not sure we should remove the plain-JS `fetch`-style examples. Seems kinda nice to show how simple it is. Maybe we could add a `JS SDK` tab in addition?

And as for Python—I've never used this. I guess lots of Stellar developers must? Maybe we could add another tab for `Python SDK` also.

janewang:
Need to publish the npm package

sreuland:
Considering the large amount of content across the API docs:
* [Horizon API](https://developers.stellar.org/docs/data/horizon/api-reference)
* [RPC API](https://developers.stellar.org/docs/data/rpc/api-reference)

The permutations over each language example expressed as SDK usage(python, js, go) per each endpoint will be a considerable effort to manually copy code samples(which should be tested) as stubs into the docs repo per each language per each endpoint. This increases the amount of maintenance to keep each SDK code sample up to date going forward related to protocol updates or sdk blasts.

Before getting deeper into the solution for achieving this, another thought which may be considered more decentralized - Can we defer to each SDK to generate its own docs and examples? Has this already been considered and/or ruled out?

mollykarcher:
@sreuland there's some related discussion on this as well over on this issue - https://github.com/stellar/stellar-docs/issues/441#issuecomment-2712332988

ElliotFriend:
[...]

PR #860, opened 2024-08-02
build: fork, don't patch, open-rpc-docs-react

chadoh:
Working toward https://github.com/stellar/stellar-docs/issues/404

We were essentially forking `@metamask/open-rpc-docs-react` already, but in  a more brittle way using `"post-install": "patch-package"` and a `.patch` file that applied a patch directly to the built artifacts from the library.

I've now created an actual fork at https://github.com/AhaLabs/open-rpc-docs-react Which is published at https://www.npmjs.com/package/@ahalabs/open-rpc-docs-react

I'd like to move this to `@stellar/open-rpc-docs-react` in both GitHub and NPM, but this demonstrates that it will work, at least.


janewang:
I think we should move this to `@stellar/open-rpc-docs-react` first. cc @ElliotFriend could you help with this

PR #861, opened 2024-08-02
Globally rename Laboratory to Lab

janewang:


janewang:
I think we should probably merge this and the new URL when Lab is ready to launch, wdyt @quietbits?

quietbits:
> I think we should probably merge this and the new URL when Lab is ready to launch, wdyt @quietbits?

Yes, @janewang , I agree. :+1:  We will also need to update the `laboratory.stellar.org` (or `lab.stellar.org`) links because the routes will be different in v2.

janewang:
Yes we could redirect beta and laboratory to the new lab url. I think there are already people using the beta :)

docs/build/guides/freighter/connect-testnet.mdx
```diff
@@ -10,4 +10,4 @@ description: Create a Stellar account on Testnet with Freighter.
 
 3. Next, switch to Testnet. Testnet is available from the network dropdown.
 
-4. If your account does not exist on the selected network, Freighter will prompt you to fund it the account using Friendbot. Alternatively, you can do so in the Stellar [Laboratory](https://laboratory.stellar.org/#account-creator).
+4. If your account does not exist on the selected network, Freighter will prompt you to fund it the account using Friendbot. Alternatively, you can do so in the [Stellar Lab](https://lab.stellar.org/account/create?$=network$id=testnet&label=Testnet&horizonUrl=https:////horizon-testnet.stellar.org&rpcUrl=https:////soroban-testnet.stellar.org&passphrase=Test%20SDF%20Network%20/;%20September%202015;;).
```
quietbits, line 13: We could omit the network settings to make the URL shorter. We'll get user's network setting from the local storage, if it exists, or use testnet by default.

```suggestion
4. If your account does not exist on the selected network, Freighter will prompt you to fund it the account using Friendbot. Alternatively, you can do so in the [Stellar Lab](https://lab.stellar.org/account/create).
```

quietbits (approved):

Issue #345, opened 2024-02-28
Add versioning to Anchor Platform documentation


briwylde08:
### What problem does your feature solve?

Request from Jake and Gleb: we'd like to include versioning to the Anchor Platform documentation. Elliot is working on it!

### What would you like to see?

Be able to make different versions of documentation accessible and useable. Currently in conversation between Gleb and Elliot (primarily).

janewang:
@JakeUrban Is this an urgent ask?

JakeUrban:
It is not urgent my understanding is that it is not a significant lift to support. @ElliotFriend would have the best idea.

Ifropc:
We want this task to be done before any significant work on SEP-6 is being done, IMO (I don't think the work on SEP-6 feature will start before Q2)

ElliotFriend:
Saving for future reference: https://stackoverflow.com/questions/75229184/docusaurus-how-can-i-have-multiple-versions-of-different-docs-in-the-docs-direc

PR #872, opened 2024-08-05
[AP] cleanup specfiles and sidebar

ElliotFriend:
## What's this, then?

To get the site ready for us to add docs versioning to AP (#345), there is some preparation that should be done. First, was upgrading to docusaurus v3, and now it's cleaning up a bit how we have docusaurus handling the AP files. This PR aims to tackle that second part.

~~We're consolidates the callback, platform, and custody specfiles into a single `main.yaml` file (notwithstanding the external schemas, etc. files).~~ (Edit: We're not consolidating, but still doing some cleanup) This also upgrades the SDP specfile to openapi v3.1, and fixes some spec errors in the file.

## Remains to be done:

- [x] reconfigure build to break on broken links
- [ ] add some nginx redirects since some URL paths get shifted up a directory

## TL;DR

There's a lot happening in this PR, so here's a recap of what I've done:

- consolidated the three distinct API specifications into a single specfile
- made the rpc spec behave/generate more like the soroban rpc spec (including the documentation pages generated from it)
- fiddled with the sidebar definition so we don't need to custom-code a generator
- because the custom sidebar generator is no more, i had to make some sidebar adjustments to the SDP docs, too. no big changes there, though.
- other bits-n-bobs related to the redocly cli (updates, configs, etc.)


openapi/anchor-platform/main.yaml
```diff
@@ -1,26 +1,241 @@
-openapi: 3.0.0
+openapi: 3.1.0
```
Ifropc, line 1: Do we need to store all openapi stuff in `main.yaml` from now on?

openapi/anchor-platform/main.yaml
```diff
@@ -1,26 +1,241 @@
...
-    anchor's responses to these requests affect the Platform responses to the client application.
-  title: Synchronous Callbacks API
+    The platform also includes features for managing assets, transactions, and user accounts, and supports a variety of deployment options, including using Docker or Kubernetes. Overall, the Anchor Platform aims to simplify the process of building and managing a Stellar-based financial service, allowing businesses to focus on providing value to their customers.
+
+    Learn about integrating with the Anchor Platform in the [Admin Guide](https://developers.stellar.org/platforms/anchor-platform/admin-guide) or get API information in the [API Reference](https://developers.stellar.org/platforms/anchor-platform/api-reference).
+  termsOfService: https://stellar.org/terms-of-service
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
 tags:
```
Ifropc, line 16: Everything below is copy of Platform API + Custody API right?

Ifropc (commented):
Do you mind adding a small `README` in `/openrpc` folder to maybe add some information on how to add new method? (I think changing existing methods is relatively straight-forward but if you think it should also be documented it'd be a nice addition as well)

patches/@metamask+open-rpc-docs-react+0.1.2.patch
```diff
@@ -21,12 +21,15 @@ index 354ce73..129371f 100644
 +            result: examplePairing.result.value
 +        }
 +        // const jsCode = `await window.ethereum.request(${JSON.stringify(methodCall, null, "  ")});`;
++        const serverUrl = methodName.indexOf('_') > -1
```
Ifropc, line 24: Not sure if method name containing `_` is a good enough way to distinguish it... We may add something on either RPC side that could break this logic. Is there anything else we can use? Tags maybe?

docs/data/rpc/api-reference/methods/getEvents.mdx
```diff
@@ -4,5 +4,8 @@ description: Returns contract events
 ---
 
 import { RpcMethod } from "@site/src/components/RpcMethod";
+import rpcSpec from "@site/static/stellar-rpc.openrpc.json";
 
-<RpcMethod method="getEvents" platform="soroban" />
+<RpcMethod
+  method={rpcSpec.methods.filter((meth) => meth.name === "getEvents")[0]}
```
Ifropc, line 10: Could it be refactored so it's possible to call in like before?
`<RpcMethod method="getEvents">`

Ifropc (approved):

patches/@metamask+open-rpc-docs-react+0.1.2.patch
ElliotFriend: You're absolutely right, this is a **really fragile** way to make this check. We're well on our way to publishing [our own fork](https://github.com/stellar/open-rpc-docs-react) of the package, and this patch file will (thankfully) be out of the picture soon. So, it seemed like this quick-n-dirty bandaid might be sufficient just for the time-being.

We could definitely do a more robust check now, if you'd prefer, though. iirc the stellar-rpc doesn't use tags for _any_ of its methods, so that probably is a better way to go about it.

docs/data/rpc/api-reference/methods/getEvents.mdx
ElliotFriend: The short answer is "I'm not sure if we can."

The reason for co-locating the specfile _into_ the `/platforms/anchor-platform` directory is to make it a part of the versioned docs so when someone clicks from v2 to v3, the proper specfile gets rendered and displayed.

But, the `RpcMethod` React component lives entirely outside of that context. I'm not sure how to make the component work across differing versions (or non-versioned like stellar-rpc) while still passing in `getEvent`-like strings like we had before (which i do agree has much better ergonomics). I'm open to ideas or suggestions, for sure, I was just drawing a blank.

openapi/anchor-platform/main.yaml
ElliotFriend: At the moment, yes. That's the "heart" of the specfile, kindof like the "index.html" page. In the future, we could certainly break it into a more manageable series of smaller files (similar to how Horizon is done (but also done better, that ones a bit of a rat's nest lol)) to get bundled together (like how `yarn api` bundles them already).

ElliotFriend: Even if we did break things into sub-files, they will still need to get `$ref`ed in the `main.yaml` file in order to be bundled up.

openapi/anchor-platform/main.yaml
ElliotFriend: correct

ElliotFriend: Actually, Platform API + Custody API + Callbacks API

patches/@metamask+open-rpc-docs-react+0.1.2.patch
Ifropc: Sounds good, if it's temporary let's just leave the glue there for now

docs/data/rpc/api-reference/methods/getEvents.mdx
Ifropc: > to make it a part of the versioned docs so when someone clicks from v2 to v3, the proper specfile gets rendered and displayed.

Got it, I didn't think about it, makes sense! 
Not sure if it can be done in a better manner then 😔

openapi/anchor-platform/main.yaml
Ifropc: Got it, any reason for merging them back together? I liked the previous one because it's easy to navigate, but don't have any strong preference. 
Also cc @philipliu for opinion on merging this yaml files together

openapi/anchor-platform/main.yaml
Ifropc: Yep I think that was how it was done before, right?

openrpc/README.md
```diff
@@ -13,17 +14,58 @@
...
+If you're here to add a new method to one of the specfiles, you'll want to:
+
+1. Create a new `<method_name>.json` file in the relevant
+   `/openrpc/src/{stellar-rpc,anchor-platform}/methods` directory. Follow the
+   OpenRPC specification concerning [method objects] for defining this file.
+2. Feel free to use any existing (or create new) `$ref` objects along the way.
+   More about them [later on](#json-refs).
+3. Once your method is defined the way you want it, run the [build](#building)
+   and [validate](#testing) scripts to generate updated "bundled" specfiles.
+4. Create a new Docusaurus page for the method. You can use one of the existing
```
Ifropc, line 28: Perfect thank you very much! I honestly spent waaay too long to figure out how to do that when I needed to 😓

openapi/anchor-platform/main.yaml
philipliu: I would prefer if we could keep them separate

Ifropc:
LGTM, thanks @ElliotFriend ! We can always get back to splitting API files later on

Ifropc (approved):

openapi/anchor-platform/main.yaml
ElliotFriend: The reasoning was that it's a bit of a "pre-emptive" move for the versioning setup. It would be much simpler to coordinate and manage the versions of a single specification file, rather than 3 or 4 different files.

That said, my mind has been moving in the same direction as you both over the last couple days. I think  there are a couple options to address it. I'm gonna dig a bit into the [versioning capabilities](https://docusaurus-openapi.tryingpan.dev/versioning) of the openapi plugin we're using, as well as this [experimental `join` command](https://redocly.com/docs/cli/commands/join) in the Redocly CLI. I think both could provide a reasonable option for what we're after (which is: A useful, maintainable specfile (or set of them) that cleanly and nicely generates the documentation pages for us).

ElliotFriend:
Updates since last major review:

- We've gone back to keeping the various specs of the Anchor Platform as different files. file management in the repo is a _little_ more cumbersome, but the trade-offs are worth it.
- The sidebar is generated now in a much simpler manner, it's just an autogenerated directory 🎉 
- The `redocly` CLI is configured to include and manage commands for the different parts of the AP specs
- The sidebar is now "flattened" out a bit more, so there are fewer dropdowns that need to be navigated through for users
- I've moved the RPC spec documentation _into_ the "Platform Server" category, since the methods exist on the same instance/server. (we can easily undo this, it felt like it made some sense, though?)

Here's the PR Preview:
https://developers-pr872.previews.kube001.services.stellar-ops.com/platforms/anchor-platform

Would love another review to see how we think this'll work out.

Ifropc (approved):

Issue #837, opened 2024-07-29
Use of prepareTransaction missing from RPC JS example code


leighmcculloch:
Docs page: https://developers.stellar.org/docs/data/rpc/api-reference/methods/sendTransaction

Example code for the JS SDK doesn't include the use of `prepareTransaction`.

Reported discussion:
- https://discord.com/channels/897514728459468821/1267544718313521173/1267607906409513036

PR #886, opened 2024-08-06
Fix `sendTransaction` example by simulating the transaction prior to submission

Shaptic:
Soroban transactions need to be simulated prior to submission, so we need a `prepareTransaction` call which will simulate and assemble for us. 

Closes https://github.com/stellar/stellar-docs/issues/837

### DIFF
--- a/docs/data/rpc/api-reference/methods/sendTransaction.mdx
+++ b/docs/data/rpc/api-reference/methods/sendTransaction.mdx
@@ -76,7 +76,7 @@ async function sendTransaction() {
     const sourceKeypair = StellarSdk.Keypair.fromSecret(sourceSecretKey);
     server.prepareTransaction(tx).then((result) => {
       result.sign(sourceKeypair);
-      return sendTransaction(transaction);
+      return sendTransaction(result);
     }).then((result) => {
       console.log("hash:", result.hash);
       console.log("status:", result.status);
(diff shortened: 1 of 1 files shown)


Myestery:
Hello @Shaptic, I'll like to also add that the `Server` variable imported here wasn't imported well
it is `import { Server } from "@stellar/stellar-sdk/rpc" `
But this server instance has issues when used to prepareTransaction
My bad here for not picking them out when I wrote these examples, it should be

```
import {  SorobanRpc } from "@stellar/stellar-sdk";
const server = new SorobanRpc.Server("https://soroban-testnet.stellar.org");
```

Does it make sense to write an issue on the SDK repo for the issue with `prepareTransaction` ?

ElliotFriend:
@Myestery Thanks for pointing that out.

In the source where the SDK [exports the RPC server](https://github.com/stellar/js-stellar-sdk/blob/5ab7797aff177f2179fb1559ae613899e73919a1/src/index.ts#L19-L34), they're working to deprecate the `import { SorobanRpc } from ...` convention you've mentioned. I'm curious what kind of problems you've had with `prepareTransactions` using the other method?

I think it would be really useful to create an issue in the [SDK repo](https://github.com/stellar/js-stellar-sdk/issues/new/choose) for that!

ElliotFriend (approved):
Great catch, @Shaptic! Thanks for the fix!

PR #913, opened 2024-08-15
[DOC] Update dapp-frontend.mdx so it includes how to use the StellarWalletsKit to connect the wallets

earrietadev:
This change includes a quick guide of how to use the StellarWalletsKit when using Freighter to request the public key and signing a transaction in the dapp-frontend guide: https://developers.stellar.org/docs/build/apps/dapp-frontend

This PR comes to complete this task: https://github.com/stellar/stellar-docs/issues/766

PR #923, opened 2024-08-16
dApp tutorial: fix wrap selector

tupui:
Follow up of #913

I was just going through this now and saw this leftover as I was running the snippet 😃 

There is one other instance bellow but not critical.

cc @ElliotFriend @earrietadev


ElliotFriend (approved):
Really great catch! Thank you!!

PR #928, opened 2024-08-20
[Wallet SDK] adding more methods for Stellar Network section

piyalbasu:


CassioMG (approved):
LGTM!

Will you leave ​​`transferWithdrawalTransaction` out of this?

piyalbasu:
> LGTM!
> 
> Will you leave ​​`transferWithdrawalTransaction` out of this?

So I noticed that there's a reference to transferWithdrawalTransaction in the SEP-24 docs: https://developers.stellar.org/docs/build/apps/wallet/sep24

Were you thinking some other information needed to be added here? From my understanding of the method, it seemed like it wouldn't be useful without the context of SEP-24. But happy to add it under Stellar Network if you think it's useful!

CassioMG:
> > LGTM!
> > Will you leave ​​`transferWithdrawalTransaction` out of this?
> 
> So I noticed that there's a reference to transferWithdrawalTransaction in the SEP-24 docs: https://developers.stellar.org/docs/build/apps/wallet/sep24
> 
> Were you thinking some other information needed to be added here? From my understanding of the method, it seemed like it wouldn't be useful without the context of SEP-24. But happy to add it under Stellar Network if you think it's useful!

@piyalbasu I think you are right, the Sep-24 context makes more sense for this function!

PR #909, opened 2024-08-13
[AP] Versioning Setup

ElliotFriend:
Follow-up from #872, we can now setup and configure the Anchor Platform docs to support versioning for the upcoming v3 release.

PR #929, opened 2024-08-20
[AP] Create 2.9.0 docs

philipliu:
This was generated using `VERSION=2.9.0 yarn ap:versions:new`.


ElliotFriend:
I'm honestly a little surprised and delighted that the command (looks to have) worked out reasonably well 🤣

openapi/anchor-platform/versions/callbacks-2.9.0.yaml
```diff
@@ -0,0 +1,1309 @@
+openapi: 3.1.0
+info:
+  version: 3.0.0
```
ElliotFriend, line 3:
```suggestion
  version: 2.9.0
```

When I was working on #909, I changed the `info.version` in the specfiles to `3.0.0`, 'cause I thought that would be next up. You'll probably want to change that in these files. Maybe even in the `/openapi/anchor-platform/main-*.yaml` files, too, depending on what your release cycles look like

philipliu: Thanks for the catch! `2.9` should be the last minor version before our `3.0` release, so I think it makes sense to only update the versions in `/openapi/anchor-platform/*-2.9.0.yaml`s.

lijamie98: Is there a way to know what's changed from `2.8.x` to `2.9.0`?

Ifropc:
Do we want to create docs for each new minor version? Or will we skip if there are no changes in docs? 🤔

ap_versioned_docs/version-2.9.0/admin-guide/getting-started.mdx
```diff
@@ -0,0 +1,384 @@
...
+## Installation
+
+import { CodeExample } from "@site/src/components/CodeExample";
+
+The easiest way to install the Anchor Platform is to pull the [docker image][anchor-platform-image].
+
+<CodeExample>
+
+```bash
+docker pull stellar/anchor-platform:latest
```
Ifropc, line 15: We still pill `latest` here, I think we should start pinning versions from now on. Maybe do replacement of 
`stellar/anchor-platform:latest` -> `docker pull stellar/anchor-platform:<vesion>` as part of `yarn ap:versions:new` in the future? (It's a nice to have but don't think is mandatory)

philipliu: Good call. I'll rename it manually for now.

philipliu:
> Do we want to create docs for each new minor version? Or will we skip if there are no changes in docs? 🤔

I think you mean the patch version. Yeah, I think it makes sense to publish docs for each minor version instead. i.e. have docs for `2.8`, `2.9`, and 3.0`, etc.

PR #931, opened 2024-08-21
[AP] Remove CONTRIBUTING.md from 2.9 docs

philipliu:


ElliotFriend:
I should probably ultimately modify the config and/or script to automatically remove the `CONTRIBUTING.md` file from the newly created versions. Good catch here!

ElliotFriend (approved):

PR #933, opened 2024-08-21
[AP] Make `amount_out` optional for request funds

philipliu:
A lot of the changes are from prettier formatting. The main change is in the `amount_out` description for the `request_onchain_funds` RPC.

### DIFF
--- a/openrpc/src/anchor-platform/methods/request_onchain_funds.json
+++ b/openrpc/src/anchor-platform/methods/request_onchain_funds.json
@@ -3,15 +3,29 @@
   "summary": "Request onchain funds",
   "description": "The user has to initiate transfer to the anchor",
   "paramStructure": "by-name",
-  "tags": [
-    { "name": "SEP-6" },
-    { "name": "SEP-24" }
-  ],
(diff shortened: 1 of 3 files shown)


openrpc/src/anchor-platform/methods/request_onchain_funds.json
```diff
@@ -3,15 +3,29 @@
...
-  ],
+  "tags": [{ "name": "SEP-6" }, { "name": "SEP-24" }],
   "params": [
     { "$ref": "#/components/contentDescriptors/transaction_id" },
     { "$ref": "#/components/contentDescriptors/message" },
     { "$ref": "#/components/contentDescriptors/amount_in" },
-    { "$ref": "#/components/contentDescriptors/amount_out" },
+    {
+      "name": "amount_out",
+      "description": " This is required if the transaction is associated with a firm quote and for transactions where `amount_in.asset` and `amount_out.asset` are the same.",
```
philipliu, line 13: Individual fields cannot be updated using a  `$ref`, so we need to copy the entire component. I think we should just not include `description` and `required` in the components.

PR #946, opened 2024-08-27
Add tip about missing wasm32 target.

fnando:
CleanShot 2024-08-27 at 12 09 50


willemneal (approved):
I don't think I have I approval permissions in this repo but it looks good to me!

leighmcculloch (approved):

PR #950, opened 2024-08-29
⬆ Update LICENSE copyright year

JFWooten4:
This should really include community contributors, [too](https://github.com/bitcoin/bitcoin?tab=MIT-1-ov-file#readme).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2022 Stellar Development Foundation
+   Copyright 2024 Stellar Development Foundation
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2022 Stellar Development Foundation
+   Copyright 2024 Stellar Development Foundation
```
ElliotFriend, line 190: Should we keep the starting year like `Copyright 2022-2024 ...`?

ElliotFriend (commented):
I like your idea about adding a line with contributors here!

Unless @janewang or @briwylde08 might know of some specific reason not to? I say we go for it! 🎉

LICENSE
JFWooten4: It's my understanding that the docs repos have been moving around between forks/clones since the network launched. Might the SDF claims start back in 2014, and community contributions sometime thereafter? 🤔 When were the docs open-sourced / first external commit?

briwylde08 (approved):
Lgtm! Thank you!

PR #955, opened 2024-09-01
Two not three typo

rahul-soshte:
only two methods are listed for handling inclusion fee fluctuations and surge pricing. So someone accidentally must have wrote as 'three' methods


briwylde08 (approved):
Thanks for the catch!

PR #959, opened 2024-09-04
Indexer Ordering

briwylde08:
-alphabetized
-description edits


JakeUrban (approved):
This can be solved later, but it is a bit odd to have Data Indexers in "Tools" and not "Data"

Issue #985, opened 2024-09-17
Incomplete description of automatic Quorum Set generation


nano-o:
QUALITY="CRITICAL" is not mentioned on this page: https://developers.stellar.org/docs/validators/admin-guide/configuring#automatic-quorum-set-generation

marta-lokhova:
@bboston7 would you mind taking a stab at this one?

bboston7:
> @bboston7 would you mind taking a stab at this one?

Will do!

ElliotFriend:
Hi @bboston7 👋 I'm doing a triage pass over our open issues, and this one hasn't seen activity in a while. Two quick questions:

- Is this still relevant / still on your radar?
- Is there anything you need to move it forward — review, context, a decision, someone to pair with?

If it's no longer on your plate, no worries at all — just say the word and we'll open it up for other contributors. If I don't hear back in a couple weeks, I'll free up the assignment so it doesn't look claimed.

ElliotFriend:
hey, @bboston7! unassigned this one, as promised earlier. if you still want to tackle this, please let me know, and i'll be happy to hand it back to you!

kaankacar:
Automated triage bot, acting for the docs maintainers.

I am looking at this issue now. I will report what I find.

kaankacar:
Automated triage bot, acting for the docs maintainers.

**PR open: #2823**

You are right. `QUALITY` accepts `CRITICAL`, and this page listed only `HIGH`, `MEDIUM` and `LOW`. The word `CRITICAL` appeared nowhere under `docs/`.

The PR adds it in four places:

- Both `QUALITY` field tables.
- A new "CRITICAL Quality" subsection.
- The automatic quorum set generation rules, where every `CRITICAL` entity is required.
- The diagram caption, because `validator_complete.png` still shows three levels.

I checked every statement against stellar-core v28.0.1. `kQualities` in `src/main/Config.cpp` is `{"LOW", "MEDIUM", "HIGH", "CRITICAL"}`, and `generateQuorumSet` starts at the critical level with an all-required threshold.

The PR also corrects one nearby error on the same page. It called redundancy "good practice". In fact stellar-core stops with a configuration error when a `HIGH` or `CRITICAL` home domain holds fewer than three validators.

The docs review lane takes the PR from here. A maintainer merges it.

PR #990, opened 2024-09-18
Add description of nomination leader election algorithm

bboston7:
Part of stellar/stellar-core#4387

This change adds a section to the admin guide explaining the impact of validator quality levels on SCP nomination leader election.

### DIFF
--- a/docs/validators/admin-guide/configuring.mdx
+++ b/docs/validators/admin-guide/configuring.mdx
@@ -259,16 +259,26 @@ Without those settings, your validator depends on other nodes on the network to 
 
 The Stellar Consensus Protocol uses validator quality levels in determining which validator should nominate the next transaction set for inclusion in a ledger. This process of choosing the validator that will nominate the next transaction set is called _leader election_. To start, the leader election algorithm assigns organizations and validators weights as follows:
 
-- Each `HIGH` quality organization has a weight of `1`.
-- Each `MEDIUM` quality organization has a weight of `1/(10*n)` where `n` is the number of `HIGH` quality organizations.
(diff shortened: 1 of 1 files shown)


nano-o:
Are we leaving out "CRITICAL" on purpose? I opened an issue about it yesterday for another part of the docs: https://github.com/stellar/stellar-docs/issues/985

briwylde08:
Couple editorial nits but will defer to @marta-lokhova for content correctness and approval [🙏](https://emojipedia.org/folded-hands)

marta-lokhova:
Thanks for documenting leader election! I have a similar question to @nano-o's - do we want to mention CRITICAL quality as well?

bboston7:
The problem is that there's no mention of CRITICAL anywhere else in the docs, so I think it would be wrong/confusing to mention it only in the nomination leader election section. We could add a description of CRITICAL to the admin docs, but I'm not sure whether CRITICAL was intentionally left out of the docs, which is what I think @nano-o's question is getting at.

bboston7:
After some discussion with @marta-lokhova we decided to add a generic, level-agnostic description of the algorithm, as well as an example using HIGH, MEDIUM, and LOW quality validators. I'll make that change and ping @marta-lokhova when it's ready.

bboston7:
> we decided to add a generic, level-agnostic description of the algorithm, as well as an example using HIGH, MEDIUM, and LOW quality validators.

I just pushed up a change that accomplishes this, and rebased on `main`. @marta-lokhova, this is ready for another round of review when you have a moment!

marta-lokhova (approved):

PR #723, opened 2024-06-26
📝 Refresh six classic guides

JFWooten4:
Solving #338 🫱🏽‍🫲🏿

PR #991, opened 2024-09-18
🌐 Enabling transparent developer collaboration

JFWooten4:
### Motivation

Anyone should be invited to collaborate on our documentation. I think we can agree that the most common way to uncover the `list-all-assets` endpoint is through the public reference site, yes? ✏ Perhaps the pages thereon should have a direct modification link so we can [build](https://github.com/stellar/stellar-protocol/issues/1552) the best docs together!

### Summary

Haven't heard back on obfuscation [logic](https://github.com/stellar/stellar-docs/commit/788cc251929030b4f4879e8bf5df4d586ee2a9b2#r146788687) here yet, so I've implemented an open-access policy. 🤝

These changes enable the `Edit this page` button on the API endpoint specification [docs](https://developers.stellar.org/docs/data/horizon/api-reference/list-all-payments) and rectify minor syntax items. 👨‍💻

By making the data structure pages open to public upgrades, the introductions and underlying architecture efficiencies have a chance [improve](https://github.com/stellar/stellar-docs/pull/723/commits/fff271364bd9c3492baeeceb054347fa7116812d) with each and every new reader. 🚀


ElliotFriend:
Hi, @JFWooten4!

Thanks for the suggestions! Having these API docs as collaborative as possible is a **great idea**!! There's a technical hiccup here, though. All the pages you've modified are auto-generated using the [`docusaurus-plugin-openapi-docs` package](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs). All the specification files that feed into these pages are located in the `/openapi/horizon` [directory](https://github.com/stellar/stellar-docs/tree/main/openapi/horizon). So, any changes/corrections to these `*.api.mdx` files will eventually be overwritten when the files get re-generated. Also, I'm not sure if there's a way to tell that plugin that we want the `custom_edit_url` to be left alone and/or specified to something custom.

I thought we had included some instruction on editing these pages in the main repo README, but evidently I'm mistaken. I think what I was remembering is the [README](https://github.com/stellar/stellar-docs/blob/main/openapi/horizon/README.md) that we wrote when this was the `/api` directory, and _only_ included Horizon docs. Needless to say, that text is in a less-than-useful/obvious place, and we should have it far more prominently displayed.

ElliotFriend:
Hi, @JFWooten4! I'm gonna close this PR, since we're at a bit of a "hands are tied" place with it. I'll keep thinking of how we can best "advertise" the method by which these files should be edited, thanks for capturing the current state of this problem (and more) in #1013

PR #1006, opened 2024-09-26
Revamp 'Choosing the right storage' article.

dmkozh:
I just wanted to add some clarification on independence of storage types, but ended up revamping it completely:

- Remove/update outdated info (such as reference to Wasm being stored in the contract instance)
- Remove/rephrase/move around less relevant info. The old article had a very long and irrelevant diversion on the role of state archival, that has nothing to do with the purpose of the guide.
- Add a table that highlights the differences between the storage types
- Update sections dedicated to individual storage types
- Rewrite the examples in order to avoid demonstrating anti-patterns (like auth-less storage modificaitons) and demonstrate intricacies of the storage (especially the temp storage)

I still feel like some of the info might be too verbose, but I think this is still a step in the right direction in terms of the information usefulness.


briwylde08 (approved):
Went through for editorial! LGTM otherwise!

PR #1019, opened 2024-10-03
[Fix] Remove deprecated amount_fee schema and examples

JiahuiWho:
Remove more `amount_fee` usage, including schemas, examples, references

Verified by `yarn rpcspec:build` and no more `amount_fee` will be generated


lijamie98:
Should this PR target `v3`?

JiahuiWho:
I would love to but things is kind stuck in between now. Part of the removal is already merged to `main` after 2.9 release because we were assuming 3.0 is the next release.

I would prefer just merge now to keep the doc in consistent at least

lijamie98 (approved):
LGTM

PR #1029, opened 2024-10-07
Stop redefining the same utility function everywhere

Shaptic:
People keep copy-pasting the helper `yeetTx` and it makes it hard to update the implementation.

### DIFF
--- a/docs/build/guides/transactions/submit-transaction-wait-js.mdx
+++ b/docs/build/guides/transactions/submit-transaction-wait-js.mdx
@@ -12,7 +12,7 @@ import {
   FeeBumpTransaction,
   SorobanRpc,
 } from "@stellar/stellar-sdk";
-import { Server, Api } from "@stellar/stellar-sdk";
+import { Server, Api } from "@stellar/stellar-sdk/rpc";
 
 const RPC_SERVER = "https://soroban-testnet.stellar.org/";
 const server = new Server(RPC_SERVER);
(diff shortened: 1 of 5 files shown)


JFWooten4:
Re style guide, how do we feel about the [diction](https://www.today.com/parents/teens/yeet-slang-meaning-rcna163660) here @Shaptic?

docs/build/guides/transactions/submit-transaction-wait-js.mdx
```diff
@@ -12,14 +12,15 @@ import {
   FeeBumpTransaction,
   SorobanRpc,
 } from "@stellar/stellar-sdk";
+import { Server, Api } from "@stellar/stellar-sdk";
```
ElliotFriend, line 15:
```suggestion
import { Server, Api } from "@stellar/stellar-sdk/rpc";
```

I think we want the `/rpc` at the end, right? (legit asking, i'm not entirely sure lol)

ElliotFriend (commented):
this is amazing! Thank you for tackling this problem! I know for sure people have over-copy/pasted this, and I've wanted to comb through and get rid of the cruft for a while! 🙏🏻

docs/build/guides/transactions/submit-transaction-wait-js.mdx
Shaptic: Yep! Nice catch

Shaptic:
@JFWooten4 I mean it's leftover as an informal name from an age-old quick reference lol (yeet as in throw), but you're right that it probably doesn't belong as the name for proper documentation.

JFWooten4:
Frankly, I don't know the exactly legacy contexts, but this sort of context is super helpful in drafting community contribution standards, be the ideas explicit in writing or implicit through example(s). 🫱🏻‍🫲🏽 Appreciate your perspective, @Shaptic! You are the reason my dreams revolve around getting Microsoft to allow 🧠 reactions to comments

PR #1039, opened 2024-10-14
[SDP-1302] Update `PATCH /receiver` with the option to patch the phone number

marcelosalloum:
### What

Update `PATCH /receiver` with the option to patch the phone number.
Also, add the new 409 response there.

### Why

Address https://stellarorg.atlassian.net/browse/SDP-1302.


marwen-abid (approved):
LGTM

Issue #1041, opened 2024-10-15
🏛️ Decentralize SDF specification explainer


JFWooten4:
### What problem does your feature solve?

Many Docs pages reference the SDF, which receives only a Glossary entry at most.

While a number of other blockchain networks employ foundations with varying levels of control over the network, Stellar seems unique due to its lack of mining. 🌐

In this light, might it help to clarify for new users the difference between Stellar and the SDF more explicitly?

### What would you like to see?

A standalone page extending the [small footnote](https://developers.stellar.org/docs/learn/glossary#stellar-development-foundation-sdf) detailing the history of the SDF from early SF startup days to global changemaker status. 🤝🏾

As much transparency around operations as might be prudent and not already available on public sites (which could be helpfully linked).

An explicit detailing of the SDF's role in Tier 1 operations and the communication methods used between its trusted validators.

### What alternatives are there? 🔍

We could leave this documentation scarce and vague, opening the door for legal structural scrutiny.

While this is an easier option, I respectfully submit that further community-centric development explainers might best serve ongoing decentralization efforts. 💡

Alternatively, we might create a separate site hosting internal publications and such, as presently shared by certain core devs through their personal blogs and such.

briwylde08:
Heya @JFWooten4! While this is interesting, I'm not sure it belongs in the developer docs. Happy to continue the discussion but we hope to keep the docs as straightforward as possible for developing on Stellar!

JFWooten4:
Hi @briwylde08! Thanks for all the recent cleanups, getting the action items for the docs in order recently! I appreciate the focus on developer transparency and activity throughout. Nothing more punctual or centered than explaining important development concepts.

That said, there is a relative lack of an official organizational activity timeline, especially related to blockchain operations. It's my current understanding that the majority of operational disclosures exist in a scant few blog posts over the last ten+ years. Might a singular archival page explaining some of these important ongoing facts best serve the community?

As it stands, there is effectively no definitive source to visit when it comes to the exact operative roles of the SDF. This concerns me given the relative centralization of key data points on ongoing functions implicated in the response [here](https://github.com/stellar/stellar-docs/issues/794#issuecomment-2261278083). Could a simple page[^wher] help act as an authoritative source?

The present alternatives rely on either extensive work parsing through block explorers or relying on secondhand summaries from non-SDF or community-audited public sites. This just makes it challenging for outside scrutiny, which I'll argue matters greatly given our present distribution status of the remaining lumens.

[...]

PR #1043, opened 2024-10-15
♻️ Clarify past tense action

JFWooten4:
minor implications, would ideally `href` to #1041

### DIFF
--- a/docs/data/horizon/README.mdx
+++ b/docs/data/horizon/README.mdx
@@ -5,7 +5,7 @@ sidebar_position: 10
 
 :::info
 
-Starting on August 1, 2024, SDF will be truncating historical data on SDF-owned Horizon to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
+On August 1, 2024, the SDF truncated historical data on its Horizon instances to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
 
 :::
 
--- a/docs/validators/tier-1-orgs.mdx
+++ b/docs/validators/tier-1-orgs.mdx
@@ -5,7 +5,7 @@ sidebar_position: 30
 
 To help with Stellar’s decentralization, the most advanced teams building on Stellar run validators and strive to join the ranks of “Tier 1 organizations.”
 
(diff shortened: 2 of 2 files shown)


docs/data/horizon/README.mdx
```diff
@@ -5,7 +5,7 @@ sidebar_position: 10
 
 :::info
 
-Starting on August 1, 2024, SDF will be truncating historical data on SDF-owned Horizon to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
+On August 1, 2024, the SDF truncated historical data on its Horizon instances to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
```
ElliotFriend, line 8: Only thing I might quibble with is the `the` you've added in `... 2024, the SDF ...`. It doesn't quite "fit" the way we've (generally) used `SDF` more like a proper noun (almost like a name lol).

I think we should remove that, but I'm **certainly** not planning to die on that hill hehe

ElliotFriend (commented):
Great catch here! Thanks for putting together a more accurate past-tense note.

docs/data/horizon/README.mdx
JFWooten4: It's a material style choice imo. The way I thought about it was if I was vocally saying "_the_ Stellar Dev Foundation." Certainly it's been used [both](https://github.com/stellar/stellar-docs/blob/90e01c94a7ed6982bf1b5b680ac54c02a9f9cc74/docs/data/hubble/README.mdx?plain=1#L22) [ways](https://github.com/stellar/stellar-docs/blob/90e01c94a7ed6982bf1b5b680ac54c02a9f9cc74/openrpc/scripts/build.mjs#L149) throughout. 💬

For me, the larger question is how do we frame the Foundation as the single network participant that it is. Namely, while you, at the Foundation, may review this PR, the underlying network infrastructure always stays in the hands of actual validators. 🌐 As we all know, this leaves the SDF as a single network peer, which I think we ought focus on by continually [identifying](https://github.com/stellar/stellar-docs/issues/1041) its limited foundational role (e.g. via isolating _the_ introducing article).[^ind]

[^ind]: More to the point, this begins the necessary process of decentralizing not only the underlying `core` tech stack, but also the community documentation which is so key to permissionlessly educating the next billion about the power of Stellar. 🌌

JFWooten4: Does that make sense? Feel free to lmk your initial reaction and perhaps we could move overarching items into a discussion. 💭 Frankly, I think this should be included in the SDF's [404 Horizon response](https://horizon.stellar.org/transactions/90de1ff887cd708e1d30db0f17e9a0f6e646f86b5c04154d87957c20127080fc) `detail` rather than the definitive network docs.

briwylde08: I think we're okay to leave the `the` in there. Let's merge it!

briwylde08 (approved):
Thank you!

Issue #799, opened 2024-07-16
🎨 Create a style guide


JFWooten4:
### What problem does your feature solve?
GitHub content, community interactions, and actual code all coalesce into our collective documentation.[^max_tx]

Accordingly, [we](https://github.com/stellar/stellar-docs/pull/737) [can](https://github.com/stellar/stellar-docs/pull/666) [create](https://github.com/stellar/stellar-docs/pull/162) stellar docs detailing our knowledge, collectively describing how the network functions. 👩‍💻

Today's nonexistent contribution style guide leaves one person voicing opinions with a centralized SDF reviewer.[^code_rev] This isn't fair for the reviewer's time, the contributor's confusion, or the browser's sanity. Moreover, a clear style guide can help developers independently build in line with our shared vision. 🌄

### What would you like to see? 📝

~~A style guide under the `CONTRIBUTING.md` page.~~ A `CONTRIBUTING.md` file which explicitly details general [collaboration styles](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#stellar-core-contribution-specifics).

We respectfully submit that an independent Contributing page outside of `stellar-docs/README.md` would significantly further community collaboration since:
- The Contributing page could act as a single frame of reference for cooperative ecosystem developers, 🤝
- The README page could be specialized to introduce visitors to the open-source network documentation, 👀 and
- Nuanced author participation tips could be collected to support an inclusive submission opportunity. 🌐

Ultimately, might these guideposts steer us all towards a newfound documentation experience, without unnecessary missteps?[^conflict]

### What alternatives are there?
1. We could leave the community reviewers up to their own devices when manually looking over pull requests. This is what happens now, and contributions often get merged despite conflicting syntax. 🤔 I'm not saying that you must follow the guidelines 100% or else get rejected; just proposing that they exist in the first place when someone does want to follow the standard practice or clean up existing work.

2. We could use Stella, stricter `prettier` logic, or some other AI to automate the formatting amendments. 🤖

3. Do nothing and leave my conscious unbearably restless, likely causing nightmares. 👻

[...]

PR #1044, opened 2024-10-15
🧑🏾‍🎨 Open source contributing foundations

JFWooten4:
solving #799

@briwylde08 mentioned that "networks tabs you can't use .mdx at the end of the link" - but I'm unfamiliar with the nuances here 🤔


briwylde08:
Hey! Great job on proposing a contribution guideline section for the docs — this will be really helpful as we encourage more community involvement. I especially like the details around writing code examples, defining Stellar terms, and cross-referencing.

Some general feedback I have is on language in general.

I think the language could be clearer and more neutral to better suit developer documentation. For example, in the Code Examples section, you wrote:

'If you're technically inclined, example applications of documented functions can go a long way to further the network's understanding. Writing a quick snippet is an easy way to connect with developers, and the local simulation translates code!'
Instead, I’d suggest simplifying it to:

'Here's how to write correctly-formatted code blocks in Markdown.'

The goal is to make the language more straightforward and focused on practical guidance. I’d also recommend removing language like:

'Let's very carefully implicate some of the ethos behind the community in an effort to distill Balancing the World's humanitarian principles into something that's subtly nudging towards an empowering, creative, and inclusive readership tone.'

This kind of language is too complex and opinionated for readers who are looking for clear and actionable guidance on contributing. I think we should aim for concise, clear, and direct phrasing to ensure it’s accessible for all contributors.

Again, I think this is a great start and I'm excited to get a contribution guide out there!

ElliotFriend:
Great start @JFWooten4! As I'm thinking about what needs to be in the contribution guidelines, here are some "sections" that come to mind. You likely already have thought of (most of) these, but I'll just write them out for the sake of collaboration.

- referencing other docs pages (i.e., `*.mdx` vs non-mdx files, and when to use which)
- code example indentations (i.e., 4 spaces for python, 2 spaces for javascript, etc.) (ideally, we should have that codified in a prettier config or something)
- how to use the custom react components (seems like you've already gotten that ball rolling pretty well)
- how to run the formatting/linting scripts, when/why you should, and if you should commit _everything_ or just the files you're working on.
- handling certain characters to avoid markdown compilation errors (`<` for example will cause the compilation to fail, and it should be escaped like `\<`)
- special-case source files for some pages (openapi specs, cli guides, etc.)

Again, great start! Glad to have this in the works!! Thanks again

JFWooten4:
@ElliotFriend @briwylde08 Thank you both so much for voicing your initial reactions to my very rought draft! 💜 I will implement these thoughts and [more](https://github.com/users/JFWooten4/projects/3) in due time, and then circle back. One last thing I was wondering, I [noticed](https://github.com/stellar/stellar-docs/blob/94890315304141b999bee831a5ca86201b205eb6/docs/data/hubble/analytics-platforms.mdx?plain=1#L17) that a lot of the pages use `[link](./FILENAME.mdx)` while a simpler syntax could use `[link](FILENAME.mdx)`&mdash;any opinionated preference on same-dir referencing?

JFWooten4:
How do you two feel about simplifying Go type declarations when multiple parameters share the same type?[^info][^contra]

[^info]: In Go, you can group parameters of the same type to simplify function signatures: `func addLiquidity(source *keypair.Full, maxReserveA, maxReserveB float64) {}`[^insfo] (option A)

[^insfo]: _Note_ the `maxReserveA` does not have a subsequent type defined, since it is assumed that it's a `float64` when `maxReserveB` compiles.

[^contra]: _Compare_ behavior to that in Java, whereby you must explicitly declare the type for each parameter, even if they share the same type: `public static void addLiquidity(KeyPair source, String maxReserveA, String maxReserveB) {}` (option B)

In the limited context of Go functions, do you have a preference for using entirely explicit typedefs v. inheriting previous?

If it's a binary answer, perhaps a 🎉 for "defining all" (option B) or a ❤️ for "item grouping" (option A)?

JFWooten4:
We'll also need to clarify heading trailing commas, sectional `href` names, and cutoff lengths.

PR #1046, opened 2024-10-17
[AP] [ANCHOR-848] Rename AP versions from major.minor.path to major.minor

lijamie98:
Rename AP versions from `major.minor.patch` to `major.minor`.

- `2.8.4` to `2.8`
- `2.9.0` to `2.9`
- `2.10.0` to `2.10`
- `2.11.0` to `2.11`


JiahuiWho:
The version format in scripts may needs to be changed as well?

https://github.com/lijamie98/stellar-docs/blob/ap/anchor-848-rename-versions/scripts/ap_new_version.mjs

lijamie98:
> The version format in scripts may needs to be changed as well?
> 
> https://github.com/lijamie98/stellar-docs/blob/ap/anchor-848-rename-versions/scripts/ap_new_version.mjs

The API version still has `major.minor.0`. I hesitate to change it at this moment because it may break the code.

ElliotFriend:
Hey, friends! Just wanting to understand the context: Is there a reason we want to do away with the `.patch` portion of the version numbers? I could potentially foresee some problematic behavior from our openapi plugin (though I'm not certain that it _would_ behave poorly), depending on what you're after here.

JiahuiWho:
Hi @ElliotFriend, thanks for checking on and we'd also like to get your input on this. We plan to release new documentation only with minor and major versions, not with patch versions, as updating documentation that frequently would be too cumbersome. Is there a way to omit the patch version number? Or should we consistently use `.0` in the version instead?

lijamie98:
> Hey, friends! Just wanting to understand the context: Is there a reason we want to do away with the `.patch` portion of the version numbers? I could potentially foresee some problematic behavior from our openapi plugin (though I'm not certain that it _would_ behave poorly), depending on what you're after here.

@ElliotFriend As @JiahuiWho mentioned, we don't plan to change the documentation between patches. Is it going to cause issues with stellar docs if we omit the patch ?

lijamie98:
@ElliotFriend can you give us a go if we use `major.minor` to replace `major.minor.patch` is not an issue for future releases?

JiahuiWho (approved):

lijamie98:
We will merge for now. If removing the `patch` from the version causes issues, we can resolve later.

PR #1055, opened 2024-10-22
Update stellar-cli docs to support the latest protocol 22.

fnando:


docs/build/smart-contracts/getting-started/setup.mdx
```diff
@@ -127,48 +127,14 @@ Users of non-bash shells may need to adapt the above commands to suit their need
 
 ### Configuring the CLI for Testnet
 
-Stellar has a test network called Testnet that you can use to deploy and test your smart contracts. It's a live network, but it's not the same as the Stellar public network. It's a separate network that is used for development and testing, so you can't use it for production apps. But it's a great place to test your contracts before you deploy them to the public network.
```
fnando, line 130: stellar cli has default networks, so this is no longer required.

src/helpers/stellarCli.ts
```diff
@@ -0,0 +1,2 @@
+// Stub file: the real file is generated in build time by `yarn stellar-cli:build`.
+export const latestVersion = "0.0.0";
```
Ifropc, line 2: Should we still commit latest version? If you build locally, there'll be unstated git changes

fnando: I don't think we should. Running local dev doesn't update this file, it's unlikely that you'll build the final site (we do something similar with the manual), but if you do, having 0.0.0 is a great way of knowing that it didn't work properly.

leighmcculloch (approved):

PR #1060, opened 2024-10-25
`getLedgerEntries`: Outline building and parsing all sorts of `LedgerKey`s

### LINKED ISSUE #469: getLedgerEntries: add meaningful coding examples
The `getLedgerEntries` method is pretty powerful, you can retrieve any piece of state from the ledger: accounts, trustlines, offers, data, liquidity pools, contract wasms, contract instances, contract storage, yada yada.

However, there are two challenges associated with that: 
1. building the relevant xdr key for the type of ledger entry you're looking for
2. parsing the xdr result

However, our coding examples use pre-built keys and don't parse the results. I think that modifying our JS and Python examples in this page to include these can help unlock human potential

```[tasklist]
### Tasks
- [x] Add a full list of keys and description for each key
- [x] Add examples for how to use the keys to retrieve accounts and balances
- [x] Add examples for how to use the keys to retrieve trustlines
- [x] Add examples for how to use the keys to retrieve contract wasms
- [x] Add examples for how to use the keys to retreive contract state / storage
- [x] Add examples for how to use the keys to retrieve contract instances
```

Shaptic:
This makes the following changes:
 1. Add a guide on building + parsing account and trustline entries
 2. Improve the guide on building + parsing contract data, instances, and WASMs

As part of (2) I tried to morph it to be in line with the way (1) is explained. Notice that I've removed all of the base64 cruft: this is intentional as these can quickly become out of date and the guide is structured to teach someone to fish rather than giving them a base64-encoded fish.

Closes #469.

## For Reviewers
I recommend viewing the entire Markdown file in this PR as-is and comparing it to the [existing page](https://developers.stellar.org/docs/data/rpc/api-reference/methods/getLedgerEntries) rather than going line-by-line as the changes are significant and hard to parse out of a diff.

[Page preview link](https://developers-pr1060.previews.kube001.services.stellar-ops.com/docs/data/rpc/api-reference/methods/getLedgerEntries).


docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -3,28 +3,125 @@ hide_title: true
...
+The Stellar ledger is, on some level, essentially a key-value store. The keys are `LedgerKey`s and the values are `LedgerEntryData`s in the XDR. An interesting product of the store's internal design is that the key is a _subset_ of the entry: we'll see more of this later.
 
-The example above is querying a deployment of the [`increment` example contract] to find out what value is stored in the `COUNTER` ledger entry. This value can be derived using the following code snippets. You should be able to extrapolate from the provided examples how to get `keys` parameters for other types and values.
+The `getLedgerEntries` method returns the "values" (or "entries") for a given set of "keys". `LedgerKey`s come in a lot of forms, and we'll go over the valuable ones on this page alongside tutorials on how to build and use them.
 
-#### Python
+## Types of `LedgerKey`s
 
-:::note
+The source of truth should always be the XDR defined in the protocol. `LedgerKey`s are a union type defined in [Stellar-ledger-entries.x](https://github.com/stellar/stellar-xdr/blob/529d5176f24c73eeccfa5eba481d4e89c19b1181/Stellar-ledger-entries.x#L600). There are 10 different forms a ledger key can take:
```
leighmcculloch, line 20: Can this link be to a protocol version. XDR is tagged.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -3,28 +3,125 @@ hide_title: true
...
 
-If you are using the [Python](https://stellar-sdk.readthedocs.io/en/latest/) `stellar_sdk` to generate these keys, you will need to install the latest version of the SDK. This can be done like so:
+1. **Account:** holistically defines a Stellar account, including its balance, signers, etc. (see [Accounts](https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/accounts))
```
leighmcculloch, line 22: At the beginning of the list, the list is describe as "forms a ledger key can take," however the list describes the ledger entries, rather than the keys. For example:

> **Account:** holistically defines a Stellar account, including its balance, signers, etc.

Maybe:

> There are 10 different forms a ledger key can take and each refers one-to-one with the following ledger entries:

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -3,28 +3,125 @@ hide_title: true
...
+2. **Trustline:** defines a balance line to a non-native asset issued on the network (see [`changeTrustOp`](https://developers.stellar.org/docs/learn/fundamentals/transactions/list-of-operations#change-trust))
```
leighmcculloch, line 23: The 'defines' is pretty repetitive in this section, suggest dropping it:
```suggestion
1. **Account:** a Stellar account, including its balance, signers, etc. (see [Accounts](https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/accounts))
2. **Trustline:** a balance holding an issued asset (see [`changeTrustOp`](https://developers.stellar.org/docs/learn/fundamentals/transactions/list-of-operations#change-trust))
```

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +145,191 @@ print(
...
-2. Then, we can look up the raw Wasm byte-code using that hash.
+2. Then, we can look up the raw WASM byte-code using that hash.
 
-#### Request the `LedgerKey` for the Contract Code
+#### 1. Find the ledger key for the contract instance
 
-##### Python
+<CodeExample>
 
 ```python
```
leighmcculloch, line 186: Some examples are Python and JavaScript, some are Python and TypeScript, and some are unlabelled and I assume just JavaScript. Can we make them consistent, maybe all JS instead of TS, and all also have Python?

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +145,191 @@ print(
...
 
-```json
+Then you can inspect them accordingly. Each of the above entries follows the XDR for that `LedgerEntryData` structure precisely. For example, the `AccountEntry` is in [`Stellar-ledger-entries.x#L191`](https://github.com/stellar/stellar-xdr/blob/529d5176f24c73eeccfa5eba481d4e89c19b1181/Stellar-ledger-entries.x#L191) and you can use `.seqNum()` to access its current sequence number, as we've shown. In JavaScript, you can see the appropriate methods in the [type definition](https://github.com/stellar/js-stellar-base/blob/6930a70d7fbde675514b5933baff605d97453ba7/types/curr.d.ts#L3034).
+
+## Viewing and understanding XDR
+
+If you don't want to parse the XDR out programmatically, you can also leverage both the [Stellar CLI](https://developers.stellar.org/docs/tools/developer-tools/cli/stellar-cli) and the [Stellar Lab](https://lab.stellar.org/xdr/view) to get a human-readable view of ledger keys and entries. For example,
+
+```bash
+echo 'AAAAAAAAAAAL76GC5jcgEGfLG9+nptaB9m+R44oweeN3EcqhstdzhQ==' | stellar xdr decode --type LedgerKey | jq
```
leighmcculloch, line 329: `jq` makes some assumptions about a developers environment and assumes some familiarity with the tool. Suggest tweaking as below to be more inclusive, and to narrowly focus on the Stellar tools:
```suggestion
echo 'AAAAAAAAAAAL76GC5jcgEGfLG9+nptaB9m+R44oweeN3EcqhstdzhQ==' | stellar xdr decode --type LedgerKey --output json-formatted
```
This command with the use of echo and pipe probably won't work on Windows. Could you add a tabbed box for Windows with the equivalent Windows command? The docs don't do this consistently yet but in new docs we're putting some effort into working on windows. (@fnando has suggested using UTM to try out windows commands.)

leighmcculloch (commented):
Few comments inline. Looks like there's some incomplete docs here.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: imo `jq` is as ubiquitous as `curl` these days, but native to the CLI is def better :+1: 

I'm 100% not going to figure out how to spin up a VM+Windows on Linux just to get this PR merged. I'm also 99% sure this will work: `echo` and `|` work the same way in Batch ([example](https://stackoverflow.com/a/33292435)). If it doesn't, I'm sure someone will raise an issue or figure out the equivalent.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: None of them are unlabeled; I'm not sure what you mean? It's all TS and/or Python. Also I don't use the Python SDK enough to make examples for them - the ones included here are copy-pasta from elsewhere iirc.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: This is a writing style choice, no? I actually made it repetitive on purpose to grammatically have parallel structure.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -3,28 +3,125 @@ hide_title: true
...
+    .toString()} stroops of ${asset} with a limit of ${trustlineEntryData
+    .balance()
+    .toString()}`,
+);
 ```
 
-:::
+### Contract Data
+
+Suppose we've deployed the [`increment` example contract] and want to find out what value is stored in the `COUNTER` ledger key. To build the key,
```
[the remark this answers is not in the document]
Shaptic: The code example is the rest of the sentence. Replaced `,` with `:` for clarity!

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +147,179 @@ print(
...
 
-#### JavaScript
-
-If you are using the [JavaScript](https://stellar.github.io/js-stellar-sdk/) `stellar-sdk` to generate these keys, you will need to install the latest pre-release version of the SDK. This can be done like so:
-
-```bash
-yarn add @stellar/stellar-sdk
-```
-
 ```js
```
tamirms, line 150: not sure if it matters but, should this be a typescript code block like the other examples above?

tamirms:
I think you might want to update the description [here](https://github.com/stellar/stellar-docs/blob/main/openrpc/src/stellar-rpc/methods/getLedgerEntries.json#L4) :

> For reading the current value of ledger entries directly.\n\nThis method enables the retrieval of various ledger states, such as accounts, trustlines, offers, data, claimable balances, and liquidity pools. It also provides direct access to inspect a contract's current state, its code, or any other ledger entry.

"retrieval of various ledger states" doesn't seem like the terminology. Perhaps you should change it to "retrieval of various ledger entries"

tamirms:
Screenshot 2025-03-07 at 3 45 15 PM

The way "LedgerKeys" and "LedgerEntries" are formatted with the backticks looks a bit strange, especially considering the plural of ledger entries is not ledgerentrys. I think it would be better to remove the back ticks for `LedgerKey`s

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +147,179 @@ print(
...
 import { xdr, Address } from "@stellar/stellar-sdk";
 
 const getLedgerKeySymbol = (contractId, symbolText) => {
-  const ledgerKey = xdr.LedgerKey.contractData(
+  return xdr.LedgerKey.contractData(
     new xdr.LedgerKeyContractData({
       contract: new Address(contractId).toScAddress(),
       key: xdr.ScVal.scvSymbol(symbolText),
+      // or, possibly, .temporary(), depending on your contract
```
tamirms, line 158: the [increment contract example](https://developers-pr1060.previews.kube001.services.stellar-ops.com/docs/build/smart-contracts/getting-started/storing-data) stores this variable in instance storage. So this comment might be confusing. Maybe you can rephrase the comment to be 
```suggestion
      // the increment contract stores its state in persistent storage but other contracts may use
      // temporary storage which have a durability of xdr.ContractDataDurability. temporary()
```

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +147,179 @@ print(
...
+</CodeExample>
 
-This can be a bit tricky to wrap your head around, but the conventions do make sense once you let it sink in.
+### Contract Wasm Code
 
-In the previous examples, the `COUNTER` _key_ was used as a `LedgerKey` while the incremented _value_ was stored in a **`LedgerEntry`**. "Ledger Entry" is the relevant term to keep in mind during this discussion. That `LedgerEntry` was stored on the Stellar ledger, and was associated with a corresponding `LedgerKey`. `LedgerKey: LedgerEntry` works the same way you would think of almost any `key: value` storage system.
+To understand this, we need a handle on how smart contract deployment works:
 
-#### How Soroban Contract Deployment Works
+- When you deploy a contract, first the code is "installed" (i.e. uploaded onto the blockchain), creating a `LedgerEntry` with the Wasm byte-code that can be uniquely identified by its hash (that is, the hash of the uploaded code itself).
```
tamirms, line 176: I think you should try to use the same terminology / description as https://developers-pr1060.previews.kube001.services.stellar-ops.com/docs/build/smart-contracts/getting-started/deploy-increment-contract . That page states that deployment is a two step process with first step being uploading / installing the code and the second step being instantiation.

In this snippet, you refer to the second step as deployment which might be confusing since the other page defines it as creation / instantiation.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +147,179 @@ print(
...
+
+## Actually fetching the ledger entry data
+
+Once we've learned to _build_ and _parse_ these (which we've done above at length), the process for actually fetching them is always identical. If you know the type of key you fetched, you apply the accessor method accordingly once you've received them from the `getLedgerEntries` method:
+
+```typescript
+const s = new Server("https://soroban-testnet.stellar.org");
+
+// assume key1 is an account, key2 is a trustline, and key3 is contract data
+const keys = await s.getLedgerEntries(key1, key2, key3);
```
tamirms, line 288: `keys` does not seem like an appropriate variable name because it actually contains the ledger entries. Maybe `response` would be a better name.

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
```diff
@@ -48,291 +147,179 @@ print(
...
+
+const account = keys.entries[0].account();
```
tamirms, line 290: might be overkill but it could be useful to highlight the case where a ledger entry doesn't exist. Also, I'm not sure if it's guaranteed that the order of `entries` will match the order of the ledger entry keys you passed into `s.getLedgerEntries()`

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: Lol oopsie, good callout

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: Good call, updated :+1:

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
Shaptic: For missing ones, I think this section could be expanded with the restoration case once we've got that in P23. Then you could write very similar stuff for missing vs. archived, but right now it seems like overkill since you'd juts get 404-like exception like any other endpoint.

tamirms:
@Shaptic do you need to run `yarn rpcspec:build`?

tamirms (approved):
not sure if you need to run `yarn rpcspec:build`. But other than that, everything looks good!

docs/data/rpc/api-reference/methods/getLedgerEntries.mdx
leighmcculloch: By unlabelled, I mean that some code blocks have no context as to which SDK they're using:

Screenshot 2025-03-08 at 10 34 38 pm

And some code examples start with Python, then continue with JS/TS, where the first code sample is labeled as Python because of the tab, then the second one isn't:

https://github.com/user-attachments/assets/603bc480-be24-400b-af9d-b425bd683fed

Shaptic: Ahh I see what you mean by "label", I thought you meant there was no highlight. Yeah, I can see how that can be misleading. I'll add `<CodeExample>`s everywhere in a new PR on Monday :+1:

leighmcculloch: Moved to an issue since this PR is merged:
- https://github.com/stellar/stellar-docs/issues/1385

Issue #836, opened 2024-07-29
🧮 Calculating claimable balance ID


JFWooten4:
https://github.com/stellar/stellar-docs/blob/71435158a2fcdbb824eae4c6445eb5219982c64a/docs/data/hubble/data-catalog/data-dictionary/enriched-history-operations.mdx?plain=1#L14

Claimable balance IDs use a "history operation id that created the balance" to get the final CB_ID. You are supposed to be able to do this outside of the network:

https://github.com/stellar/stellar-docs/blob/71435158a2fcdbb824eae4c6445eb5219982c64a/docs/learn/encyclopedia/transactions-specialized/claimable-balances.mdx?plain=1#L265

However, the documentation isn't clear on how to precisely find your CB_ID, which adds [operational complexity](https://discord.com/channels/897514728459468821/1199438502098645103/1205134179650637854).

Shaptic:
There are examples in three languages? Ctrl-f "method 1"

JFWooten4:
@Shaptic I am not entirely sure what you mean, as each instance of `method 1` in `stellar-docs/docs/learn/encyclopedia/transactions-specialized/claimable-balances.mdx` has no implementation in their respective SDKs. If this is incorrect, then please do correct me for #723, but I can't find the exact docs in any language for precomputing offer IDs. I also searched the entire doc repo, but please do let me know if I'm missing the reference you're referring to.

Shaptic:
@JFWooten4 err yeah, you're right (I had the Go tab open by default). There are still the other methods of course, but it'd be good to document computation of the claimable balance :+1: and add a JS helper for it if that's a feature request.

Shaptic:
Oh lol there already is one in JS: https://stellar.github.io/js-stellar-sdk/Transaction.html#getClaimableBalanceId, it's just not updated in the docs.

Shaptic:
The claimable balance ID is the sha256 hash of the claimable balance operation body encoded as a claimable balance ID. Namely,

```
claimableBalanceIdTypeV0(sha256(
  the operation/transaction source account's public key,
  the transaction's source account sequence number
  the operation index within the transaction
))
```

JFWooten4:
> just not updated in the docs.

Thank you for the help, George. I drafted an update in [P1516](https://github.com/stellar/stellar-protocol/pull/1516) and will get the docs updated tomorrow. Mocking up the code examples now, all per your insights!

PR #1070, opened 2024-10-30
Add JS way to fetch claimable balance ID

Shaptic:


quietbits (approved):

JFWooten4:
@Shaptic Thanks for implementing this! I have my own [instance](https://github.com/stellar/stellar-docs/blob/91b40fda94d11a46d900bbc2bd1c7b92edca7231/docs/learn/encyclopedia/transactions-specialized/claimable-balances.mdx?plain=1#L354-L356) pending submission, and just wanted to circle around on this since I saw you used `txResponse`. As I understood the existing docs and community SDKs, `method 1` is meant using only an [analysis](https://github.com/stellar/stellar-docs/issues/836) of the pre-submitted but built tx.

Clearly, you helped figure this one out and have gone the extra mile to implement it here! But I think it's intended to get fetched on the actual local object before touching Horizon. Just want to make sure we get this one right 🫱🏻‍🫲🏽

Shaptic:
@JFWooten4 ah, very nice point!! I used `txResponse` out of laziness here as a way to get the transaction envelope, but you're absolutely correct that it's intended to enable calculation prior to submission entirely.

PR #1077, opened 2024-11-06
Add a new public RPC instance.

overcat:


ElliotFriend (approved):
thanks so much @overcat!!

PR #1101, opened 2024-11-26
feat: use JS SDK's Client for "invoke contract"

chadoh:
This page contained a very complicated, old-school example of how to invoke a smart contract. I've updated it with how to use `contract.Client`.

I made this somewhat more complicated than it needs to be, showing how to also deploy a new contract from an already-uploaded Wasm hash. I wanted to do this for a few reasons:

- The part where we actually call `increment` has become so simple that
  the page feels almost unnecessary, if that's all we show. (Although
  maybe that would be good?)
- The example contract ID is almost certainly invalid. If people want to
  walk through the whole thing, it would be nice to show them how to
  walk through _the whole thing._
- It's actually fairly straightforward! It allows us to show off more of
  Client's features.

That said, I would love to add the ability to deploy directly from a file:

```ts
Client.deploy({ wasm: "path/to/some/file.wasm" })
```

If the JS SDK gains this ability, then we can clean up this example even more, removing the `stellar contract install` CLI step.

## Should we keep the old example?

Maybe it's nice to see a side-by-side of low-level JS SDK logic next to higher-level Client logic? Sort of a "peek under the hood", to show how much Client is doing. This seems most useful to SDK/client authors, since I can't imagine most app developers caring to dive into that logic. Unless they encounter some problem, and need to understand these lower-level details? Not sure if this is the document where we should do that, though. Honestly, it might be fine to refer them to the source code of `Client`.

### DIFF
--- a/docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
+++ b/docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
@@ -18,117 +18,111 @@ Please go to [the project homepage](https://github.com/stellar/js-stellar-sdk) o
 
 :::
 
-```javascript
+First, upload the bytes of [the example contract](https://developers.stellar.org/docs/build/smart-contracts/example-contracts/auth) onto the blockchain using [Stellar CLI](../../../tools/developer-tools/cli/stellar-cli). This is called "install" because, from the perspective of the blockchain itself, this contract has been installed.
+
+```bash
+stellar contract build
+stellar contract install --wasm target/wasm32-unknown-unknown/release/test_auth_contract.wasm --network testnet
+```
+
+This will return a hash; save that, you'll use it soon. For this example, we will use `bc7d436bab44815c03956b344dc814fac3ef60e9aca34c3a0dfe358fcef7527f`.
+
+No contract has yet been deployed with this hash. In Soroban, you can have many Smart Contracts which all reference the same Wasm hash, defining their behavior. We'll do that from the following JavaScript code itself.
+
+```ts
+import { Keypair } from "@stellar/stellar-sdk"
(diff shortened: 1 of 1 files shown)


docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
```diff
@@ -18,117 +18,111 @@ Please go to [the project homepage](https://github.com/stellar/js-stellar-sdk) o
...
+const rpcUrl = "https://soroban-testnet.stellar.org"
+const networkPassphrase = "Test SDF Network ; September 2015"
+const wasmHash = "bc7d436bab44815c03956b344dc814fac3ef60e9aca34c3a0dfe358fcef7527f"
+
+/**
+ * Generate a random keypair and fund it on Testnet
+ */
+async function generateFundedKeypair() {
+  const keypair = Keypair.random();
+  await fetch(`https://friendbot.stellar.org/?addr=${
```
willemneal, line 46: I wish this was a method on Keypair or an account type?

briwylde08:
@chadoh and @willemneal, howdy! Are we still working on this one? Waiting on anything? Thanks! :)

docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
chadoh: Looks like there's a [server.requestAirdrop]() function that does exactly that. Terrible name, though!

chadoh: Tried out using `requestAirdrop` here

- https://github.com/stellar/js-stellar-sdk/pull/1125

Shaptic (approved):
LGTM! :shrug:

PR #1105, opened 2024-11-29
Update Python example

overcat:
Calling contracts with the latest version of the Python SDK has become very simple, so let's update this example.


ElliotFriend (approved):
Thanks so much @overcat!!! Love the new python bindings, and love the updated example (much more concise 😁)

Thanks again!

PR #1121, opened 2024-12-11
fix: update frontend tutorial to latest libs

chadoh:
And just make sure it works, in general. It was pretty broken.

Relies on these fixes:

- https://github.com/stellar/soroban-template-astro/pull/13
- https://github.com/stellar/soroban-template-astro/pull/14

And can be cleaned up once this is merged:

- https://github.com/Creit-Tech/Stellar-Wallets-Kit/pull/49

### DIFF
--- a/docs/build/apps/dapp-frontend.mdx
+++ b/docs/build/apps/dapp-frontend.mdx
@@ -19,25 +19,32 @@ Let's get started.
 
 You're going to need [Node.js](https://nodejs.org/en/download/package-manager/) v18.14.1 or greater. If you haven't yet, install it now.
 
-We want to initialize our current project as an Astro project. To do this, we can again turn to the `stellar contract init` command, which has a `--frontend-template` flag that allows us to pass the url of a frontend template repository. As we learned in [Storing Data](../smart-contracts/getting-started/storing-data.mdx#adding-the-increment-contract), `stellar contract init` will not overwrite existing files, and is safe to use to add to an existing project.
+We want to initialize our current project as an Astro project. To do this, we can clone a template. You can find Soroban templates on GitHub by [searching for repositories that start with "soroban-template-"](https://github.com/search?q=%22soroban-template-%22&type=repositories). For this tutorial, we'll use [stellar/soroban-template-astro](https://github.com/stellar/soroban-template-astro). We'll also use a tool called [degit](https://github.com/Rich-Harris/degit) to clone the template without its git history. This will allow us to set it up as our own git project.
 
-From our `soroban-hello-world` directory, run the following command to add the Astro template files.
+Since you have `node` and its package manager `npm` installed, you also have `npx`. Make sure you're no longer in your `soroban-hello-world` directory and then run:
 
-```sh
-stellar contract init ./ \
-  --frontend-template https://github.com/stellar/soroban-astro-template
+```
+npx degit stellar/soroban-template-astro first-soroban-app
+cd first-soroban-app
+git init
+git add .
+git commit -m "first commit: initialize from stellar/soroban-template-astro"
 ```
 
-This will add the following to your project, which we'll go over in more detail below.
+This project has the following directory structure, which we'll go over in more detail below.
 
 ```bash
+├── contracts
+│   ├── hello_world
+│   └── increment
 ├── CONTRIBUTING.md
+├── Cargo.toml
+├── Cargo.lock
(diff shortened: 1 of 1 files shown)


docs/build/apps/dapp-frontend.mdx
```diff
@@ -19,25 +19,32 @@ Let's get started.
 
 You're going to need [Node.js](https://nodejs.org/en/download/package-manager/) v18.14.1 or greater. If you haven't yet, install it now.
 
-We want to initialize our current project as an Astro project. To do this, we can again turn to the `stellar contract init` command, which has a `--frontend-template` flag that allows us to pass the url of a frontend template repository. As we learned in [Storing Data](../smart-contracts/getting-started/storing-data.mdx#adding-the-increment-contract), `stellar contract init` will not overwrite existing files, and is safe to use to add to an existing project.
+We want to initialize our current project as an Astro project. To do this, we can clone a template. You can find Soroban templates on GitHub by [searching for repositories that start with "soroban-template-"](https://github.com/search?q=%22soroban-template-%22&type=repositories). For this tutorial, we'll use [stellar/soroban-template-astro](https://github.com/stellar/soroban-template-astro). We'll also use a tool called [degit](https://github.com/Rich-Harris/degit) to clone the template without its git history. This will allow us to set it up as our own git project.
```
leighmcculloch, line 22: > You can find Soroban templates on GitHub by

> `soroban-template-`


'Soroban template' doesn't feel specific enough and is also oddly over specific.

Soroban is the contract runtime. But a dapp / web app is primarily interacting with the Stellar network, not the contract runtime.

Templates could exist that don't use contracts at all too.

`stellar-template`?

leighmcculloch (commented):
One ask inline, otherwise looks good.

Reading this doc also reminded me the docs are pronoun heavy that increases the effort to read them, and shifts focus to the author and reader, rather that than the topic under discussion. Fixing that is out of scope of this PR though.

docs/build/apps/dapp-frontend.mdx
chadoh: Good point, @leighmcculloch. I think `stellar-template-*` is probably a better prefix.

So far, these templates have been focused on interacting with smart contract projects, which is why we started with `soroban-template-*`.

If we want to update this, we will need to update all the [existing templates](https://github.com/search?q=%22soroban-template-%22&type=repositories). These are mostly owned by @ElliotFriend. What do you think, Elliot?

chadoh:
> the docs are pronoun heavy that increases the effort to read them, and shifts focus to the author and reader, rather that than the topic under discussion

I'm not quite sure I understand the complaint. You mean all the switching between "we" and "you", @leighmcculloch? 

Regardless, it seems like maybe both this and the "rename to `stellar-template-` can be dealt with in a follow-up PR?

If you agree with that @leighmcculloch, then maybe you can review and merge if you like it, @ElliotFriend?

docs/build/apps/dapp-frontend.mdx
leighmcculloch: Did we consider using repo topics?

Repo topics are convenient because you can actually search and browse them on GitHub, where-as searching for repos with a particular prefix is harder / less accurate and will match on unrelated things.

For example:
https://github.com/topics/stellar

I think we could create a topic `stellar-template`:
https://github.com/topics/stellar-template

Vs searching for repos with a particular prefix, which requires devs to have knowledge of GitHub search queries:
https://github.com/search?q=soroban-template-+in:name&type=repositories

leighmcculloch: Another pro for repo topics is that any repo assigned a topic will have a little blue circle with the topic name in it on it's home page, which when folks click will take them to a browsable list of repos sorted by popularity that are also tagged with the topic. i.e. Repo topics are self discoverable.

Ifropc:
@chadoh should we bump review for all PRs that this PR depends on?

docs/build/apps/dapp-frontend.mdx
ElliotFriend: I know this conversation is a bit old at this point, sorry. I'm not very familiar with repo topics, but it _sounds_ like that's an easier, more discoverable way to go. Maybe we can also _recommend_ that `stellar-template-*` is a convention that's often used for templates?

To the point @chadoh made, it does seem like something we could tackle in a subsequent PR

PR #1124, opened 2024-12-12
add an oracles page to dev tools

ElliotFriend:
With the launch of DIA, I realized we don't have any content in the dev tools section concerning oracles. This creates that page, and adds information about the [Reflector Network](https://reflector.network/) and [DIA](https://www.diadata.org/) to start with.

Relevant preview page: https://developers-pr1124.previews.kube001.services.stellar-ops.com/docs/tools/developer-tools/oracles


janewang (approved):
Thanks @ElliotFriend. It would be great if we could order these pages alphabetically. At least that's the order currently :)

PR #1125, opened 2024-12-12
Update quickstart.mdx

ozgunozerk:
For the newcomers it may not be obvious what do the below lines do:

```js
await fetch(friendbotUrl + `?addr=${issuerKeypair.publicKey()}`);
await fetch(friendbotUrl + `?addr=${destinationKeypair.publicKey()}`);
```

they seem like just fetch / ping operation on a URL. 
In fact, these lines pre-fund the accounts. 

I had to do some extra research to be able to learn this info. Optimally, the newcomers shouldn't have to leave the docs and search it elsewhere to figure out what these lines do.


carstenjacobsen (approved):
LGTM

PR #1132, opened 2024-12-13
Remove Dapps Challenge / Soroban Quest from docs

briwylde08:


ElliotFriend (approved):
I think this covers everything now. Once the preview builds successfully, i think we're good to merge.

PR #1136, opened 2024-12-13
Update RPC OpenRPC spec with P22 fields

aditya1702:


openrpc/src/stellar-rpc/contentDescriptors/Format.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Format": {
+    "name": "format",
+    "summary": "indicates which the response format to return - json or xdr",
+    "description": "Lets the user choose the format in which the response should be returned - either as json or base64 encoded xdr strings.",
```
Shaptic, line 5:
```suggestion
    "description": "Lets the user choose the format in which the response should be returned - either as unpacked JSON or as base64-encoded XDR strings. Note that you should not rely on any schema for the JSON, as it will change when the underlying XDR changes.",
```

openrpc/src/stellar-rpc/methods/getVersionInfo.json
```diff
@@ -12,28 +12,27 @@
...
           "example": "21.1.0"
         },
-        "commit_hash": {
+        "commitHash": {
           "type": "string",
           "description": "The commit hash of the RPC server.",
           "example": "fcd2f0523f04279bae4502f3e3fa00ca627e6f6a"
         },
-        "build_time_stamp": {
+        "buildTimeStamp": {
```
Shaptic, line 25: it's `Timestamp`

```suggestion
        "buildTimestamp": {
```

Shaptic (commented):
Coupla small corrections needed! Otherwise LGTM!

Shaptic (approved):

PR #1141, opened 2024-12-17
Update software-versions.mdx

Shaptic:


briwylde08 (approved):
October __ isn't the right date? Weird.

PR #1143, opened 2024-12-18
Fix crowdin key passing

ElliotFriend:
with the change in the Dockerfile's `RUN` command, it looks like the crowdin token isn't getting properly passed to the `yarn crowdin:sync` command. This PR aims to resolve that problem.


briwylde08 (approved):
Why can't things just work?

Issue #1127, opened 2024-12-12
📌 Community-crowdsourced Horizon instance URLs


JFWooten4:
Similar to the list of [RPC providers](https://developers.stellar.org/docs/data/rpc/rpc-providers#publicly-accessible-rpc-urls), might we curate some notable public Horizon servers available for starting developers? Having a place to query live data was quite helpful for me starting out, and this could be something harder to find now that the SDF's deprecating some of these ops while rolling out Hubble. Perhaps we could add them to [this page](https://github.com/stellar/stellar-docs/blob/main/docs/data/horizon/horizon-providers.mdx), so that the default mantra for new users isn't just "spin up your own instance" right off the bat?

[<img width="640" alt="few-server-options" src="https://github.com/user-attachments/assets/3d29c462-2d42-4fc1-812b-e53e444e41d7">](https://publicnode.org/)

ElliotFriend:
@JFWooten4 Since we merged #1145, can this issue be closed? Or, do you still have more work to do towards this end?

JFWooten4:
Yes, my bad. I wish you didn't need write access to associate issues with PRs.

PR #1145, opened 2024-12-19
🌐 Add community Horizon endpoint

JFWooten4:
Implements starting #1127

### DIFF
--- a/docs/data/horizon/horizon-providers.mdx
+++ b/docs/data/horizon/horizon-providers.mdx
@@ -27,7 +27,13 @@ These providers allow access to the Testnet and Mainnet networks.
 
 †_Please note that Blockdaemon and Validation Cloud provide full historical data for Horizon._
 
-## SDF Provided Horizon (Futurenet and Testnet) {#sdf-provided-horizon}
+## Publicly Accessible Horizon URLs
+
+| Provider | Mainnet Endpoint | Testnet Endpoint |
+| -------- | ---------------- | ---------------- |
+| [LOBSTR](https://lobstr.co) | `https://horizon.stellar.lobstr.co` | N/A |
+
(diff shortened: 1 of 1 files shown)


ElliotFriend:
This is a great addition! I love the idea of aligning this page with the corresponding RPC providers page. Let's add an entry to the table with an entry for Public Node's [Artemis instance](https://publicnode.org/node/artemis/). I think the URL is `https://horizon.publicnode.org/`

ElliotFriend:
Actually, scratch that. Looks like Public Node has stopped providing a public Horizon instance. Forget what I said lol

ElliotFriend:
I peaked through the other validators that are listed on stellarbeat.io to see if anybody else has horizon instances listed in their stellar.toml file, and i can't see any others, sadly.

Also noticed that the `history_eldest_ledger` for Lobstr's (at time of writing) is from December 1, so about a month retention window.

All that begs the question: if we only have one community horizon instance and it only has a month's worth of data, do we really need to list it? Of course, if we begin normalize that there are at least _some_ Horizon instances out in the wild, then the list might begin to grow over time. So, that's certainly something to consider.

I'm not really sure either way, what do you think @JFWooten4?

JFWooten4:
Hey @ElliotFriend, completely with you—adding Public Node was my first instinct, too! Unfortunate that Horizon server stopped execution. My range of expertise also started at the Stellarbeat list, where I could only establish a connection with Lobstr (despite their relatively limited lookback period).

Notwithstanding, something is better than nothing! And a prominent server feature section here ought to inspire other community operators to add their own deployments. For instance, I shared the page with the [VAP winners](https://discordapp.com/channels/897514728459468821/1247973387779706931/1328010751008309271) and am spinning up our own resources as fast as possible to publish permissionlessly.

I say let it grow 🌱

ElliotFriend (approved):
looks great! 🌱

PR #1147, opened 2024-12-20
Add CROWDIN how to

janewang:


CROWDIN.md
```diff
@@ -0,0 +1,62 @@
...
+
+As new English source files are added to the Stellar Docusaurus Documentation GitHub repository, they are automatically uploaded to Crowdin. From there, an automated AI translation is run on the files to translate them into other target languages (currently only Spanish at the time of writing).
+
+In cases where the AI translation is inaccurate, a user with the Translator role will need to update and edit the translations manually.
+
+---
+
+## Translator User
+
+To get started, you need to set up a Crowdin account and be added to the Stellar Documents Crowdin project. This involves creating a Crowdin user account and requesting access to the project from a project owner.
```
ElliotFriend, line 17: theres a project setting that we could use to configure this, but I _think_ right now the project is "open" and anybody could join. We could enable "member moderation" (or it's called something similar), though, and we'd have to approve anybody who wants to join.

CROWDIN.md
```diff
@@ -0,0 +1,62 @@
...
+[Working in the Editor](https://support.crowdin.com/for-translators/#working-in-the-editor) 
+
+For a more in-depth look, see: 
+[Online Editor Guide](https://support.crowdin.com/online-editor/)
+
+---
+
+## Glossary
+
+For common words or phrases, it is good practice to add them to the glossary. The glossary helps the AI translator provide more accurate translations.
```
ElliotFriend, line 42: looks like for our project, the glossary is "read-only" for any member who's not a manager or owner. i'm not exactly sure which of the glossary access settings are best for our use-case, though.

CROWDIN.md
```diff
@@ -0,0 +1,62 @@
...
+
+---
+
+## Approving Translations
+
+As documents are translated, they will be assigned a translation percentage highlighted in blue. 
+
+To manage files that have been checked or proofread, it is recommended to mark them as approved. Approved files will display an approved percentage highlighted in green. 
+
+**Note:** All translations are synced with the Stellar Docusaurus Documentation regardless of whether they have been marked as approved. Approval is purely for managing translation files that have been checked.
```
ElliotFriend, line 59: this is probably something we should think about as people begin contributing. The risk of chatgpt doing "something" weird with the markdown/react syntax and causing build problems is definitely non-zero. Having an "approved" state that we're confident will build and work isn't necessarily a bad idea

ElliotFriend (commented):
great start! I'm curious if we want to name this file something else, though. if a person is just looking to contribute, they might not necessarily know what a file called `CROWDIN.md` is meant to discuss. Maybe we could do something like `LOCALIZATION.md` or `TRANSLATIONS.md`?

CROWDIN.md
janewang: @ElliotFriend I think it's possible to update Glossary permission: https://support.crowdin.com/glossary/#:~:text=To%20allow%20translators%20and%20proofreaders,access%20level%20for%20project%20members.

I'm looking into it now

janewang: Ok Glossary is updated to allow members to edit.

CROWDIN.md
janewang: Yes, let me edit this note actually. It's a good idea to have an approval process. Also crowdin is set up with roles to enable this review flow

janewang:
@ElliotFriend Please take another look! 👀

ElliotFriend (approved):
looks good to me!

PR #1152, opened 2024-12-20
Add note on setting `QUALITY` to tier 1 page

bboston7:
This change adds a note to the docs for tier 1 orgs to help them properly set their org's quality level.

### DIFF
--- a/docs/validators/tier-1-orgs.mdx
+++ b/docs/validators/tier-1-orgs.mdx
@@ -33,7 +33,11 @@ To configure a quorum set for your validator, we recommend including several Tie
 
 To see what the current recommended quorum set looks like, check out the [example Full Validator config file](https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg).
 
-**NOTE:** Be sure to properly add your own validators to your validators' quorum sets. As a Tier 1 organization with three validators and history archives, SDF recommend declaring your organization [`HIGH` quality](./admin-guide/configuring.mdx#validator-quality) in your quorum set configuration. Declaring your organization at a lower quality level [may limit its participation in consensus](./admin-guide/configuring.mdx#impact-of-validator-quality-on-nomination).
+:::note
+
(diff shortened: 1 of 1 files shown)


docs/validators/tier-1-orgs.mdx
```diff
@@ -33,6 +33,8 @@ To configure a quorum set for your validator, we recommend including several Tie
 
 To see what the current recommended quorum set looks like, check out the [example Full Validator config file](https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg).
 
+**NOTE:** Be sure to properly add your own validators to your validators' quorum sets. As a Tier 1 organization with three validators and history archives, we recommend declaring your organization [`HIGH` quality](admin-guide/configuring#validator-quality) in your quorum set configuration. Declaring your organization at a lower quality level [may limit its participation in consensus](admin-guide/configuring#impact-of-validator-quality-on-nomination).
```
ElliotFriend, line 36:
```suggestion
**NOTE:** Be sure to properly add your own validators to your validators' quorum sets. As a Tier 1 organization with three validators and history archives, SDF recommend declaring your organization [`HIGH` quality](./admin-guide/configuring.mdx#validator-quality) in your quorum set configuration. Declaring your organization at a lower quality level [may limit its participation in consensus](./admin-guide/configuring.mdx#impact-of-validator-quality-on-nomination).
```

ElliotFriend: Mostly markdown linking nitpicks here. They're silly and dumb, but it makes our build more reliable, especially with the newly introduced AI translations

However, I think it might be better suited to use `SDF`, rather than `we` in this instance. Not a deal-breaker, but worth thinking about, maybe.

If you wanted to wrap this note in an admonition (entirely optional, and also not a deal-breaker imo), you could do it like this:

```markdown
:::note

Be sure to properly ...

:::
```

ElliotFriend (changes requested):
This is a really great addition! Thanks for the contribution!

bboston7:
Thanks for the review Elliot! I wrapped the note in an admonition, which I think helps it stand out. Please take a look when you get a chance!

ElliotFriend (approved):
wonderful! Thanks!!

PR #1153, opened 2024-12-21
Add example for BLS signature

jayz22:
Corresponding example source code: https://github.com/stellar/soroban-examples/tree/v22.0.1/bls_signature

### DIFF
--- a/docs/build/smart-contracts/example-contracts/bls-signature.mdx
+++ b/docs/build/smart-contracts/example-contracts/bls-signature.mdx
@@ -45,7 +45,7 @@ $$
 
 Where $pk$ is the public key, $H(m)$ is hash of the message onto the `G2` group, $g1$ is the generator point in the `G1` group and $sigma$ is the signature. $e(,)$ denotes the bilinear pairing between a point in `G1` and a point in `G2`.
 
-The nice thing about pairing based signature is it enables signature aggregation. I.e. if you have multiple signers `pk0 .. pk0` on the a single message, you can compute the aggregate public key $pk_{agg}$ by adding up all the public keys (recall each public key is just a point on the G1 group), the aggregate signature $\sigma_{agg}$ by adding up individual signatures (which is just a point on the G2 group).
+The nice thing about pairing based signature is it enables signature aggregation. I.e. if you have multiple signers `pk_0 .. pk_n` on the a single message, you can compute the aggregate public key $pk_{agg}$ by adding up all the public keys (recall each public key is just a point on the G1 group), the aggregate signature $\sigma_{agg}$ by adding up individual signatures (which is just a point on the G2 group).
 
 Then the aggregate signature verification is just
 
--- a/docs/build/smart-contracts/example-contracts/bls-signature.mdx
+++ b/docs/build/smart-contracts/example-contracts/bls-signature.mdx
@@ -1,0 +1,531 @@
+---
+title: BLS Signature
+description: A custom account contract with BLS signature verification.
+sidebar_position: 50
+---
+
+<head>
+  <title>BLS Signature</title>
+  <meta charSet="utf-8" />
+  <meta
+    property="og:title"
+    content="A custom account contract with BLS signature verification."
+  />
+  <meta
+    property="og:description"
+    content="A basic custom account contract implementing the `FastAggregateVerify` function in [BLS Signatures](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-fastaggregateverify) draft."
+  />
+</head>
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+import { getPlatform } from "@site/src/helpers/getPlatform";
+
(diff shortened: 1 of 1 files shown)


docs/build/smart-contracts/example-contracts/bls-signature.mdx
```diff
@@ -0,0 +1,531 @@
...
+}
+```
+
+This contract is fairly simple and standard. On `init()`, it initializes the aggregate public key of all the owners, the domain separation tag `DST`, and initializes the counter to 0.
+
+It contains a single function `increment` which calls `require_auth` that will check the authorization condition defined later, and if success, increment and return the counter.
+
+### BLS Signature verification
+
+```
```
dmkozh, line 255: This should be ```rust

docs/build/smart-contracts/example-contracts/bls-signature.mdx
```diff
@@ -0,0 +1,531 @@
...
+    env.budget().reset_default();
+    env.try_invoke_contract_check_auth::<AccError>(&client.address, &payload, sig_val, &vec![&env])
+        .unwrap();
+    env.budget().print();
+}
+```
+
+Most of what's here is needed in order to create the contract client and ensure the invocation of its custom account interface for signature authorization. After the setup, calling `env.try_invoke_contract_check_auth` on the client will invoke the `__check_auth` logic we've defined in our contract.
+
+The invocation will return nothing on success, and will panic on failure. The other thing we are interested here is budget consumption for this call. At the beginning setting `env.budget().reset_unlimited()` ensure all the prep work (which emulates the off-chain work needed to be performed by e.g. a smart wallet) will not run out of budget. This isn't strictly necessary since the default budget should be large enough. `env.budget().reset_default()` right before the invocation makes sure the budget is reset to default state (also resetting metering counters) so that we can get an accurate capture of the budget for the invocation alone. `env.budget().print()` at the end prints out the budget.
```
dmkozh, line 334: Since SDK 22.0.4 we shouldn't need to do budget resets anymore. We should update the example first, but I think for the doc we can just omit this, so that there is less to change when the example is updated.

docs/build/smart-contracts/example-contracts/bls-signature.mdx
```diff
@@ -0,0 +1,531 @@
...
+        msg_g2.clone(),
+    ];
+    bls.g2_msm(vec_msg, vec_sk).to_bytes()
+}
+```
+
+Running this test will produce the following budget output, the total cpu consumption for signature verification is around 31M. And you can add as many additional public keys as you like. In general `pairing_check` is a function with linear cost, so the more unique messages that needs to be signed, the higher cost. Here because all signers sign the same content (hash of the call stack of this contract), we can do this in constant time.
+
+```
+---- test::test stdout ----
```
dmkozh, line 440: Would it make sense to omit all the cost type rows here and just leave the non-zero BLS rows?

docs/build/smart-contracts/example-contracts/bls-signature.mdx
jayz22: fixed

docs/build/smart-contracts/example-contracts/bls-signature.mdx
jayz22: fixed the example as well as removed the mentioning of budget reset

docs/build/smart-contracts/example-contracts/bls-signature.mdx
jayz22: fixed it

docs/build/smart-contracts/example-contracts/bls-signature.mdx
```diff
@@ -0,0 +1,490 @@
...
+        env.current_contract_address().require_auth();
+        let mut count: u32 = env.storage().instance().get(&DataKey::Counter).unwrap_or(0);
+        count += 1;
+        env.storage().instance().set(&DataKey::Counter, &count);
+        count
+    }
+}
+
+#[contractimpl]
+impl CustomAccountInterface for IncrementContract {
```
dmkozh, line 160: This is rather misleading, why is increment contract also an account? These should be two separate, unrelated contracts. I don't think the doc needs to contain the increment contract at all. The same actually goes for the example - tests don't even cover the increment logic - why overload the reader with non-relevant code?

dmkozh (approved):
I'll approve this for the time being, but I think the example (and the the doc) should be simplified.

docs/build/smart-contracts/example-contracts/bls-signature.mdx
jayz22: Sure, the increment example is just the bare minimal logic that's used as the basis for many existing examples, so I don't think it should hurt to include it, but I see your point.

dmkozh: But it's completely redundant here, isn't it? You're not using it to demonstrate anything.

PR #1173, opened 2025-01-12
FIX - Update bindings.mdx

armsves:
there's an error when generating the typescript bindings, source is not needed to generate them. also made the secret-key clearer and made it clear you can also use a stored identity which is safer


fnando (approved):
thanks! lgtm!

PR #1181, opened 2025-01-15
better trustline description

tomerweller:
remove the language about the relationship between trustlines and trading. In flight, as a hop in a path payment, an account can trade an asset that it does not have a trustline for


leighmcculloch (commented):
👏🏻

leighmcculloch (approved):
👏🏻

PR #1184, opened 2025-01-16
fix fuzzing example

leighmcculloch:
### What
Import the arbitrary module in the fuzzing example test.

### Why
It's required because of how arbitrary depends on it being in scope. Somehow this was missed when the test was run, and I have no idea how it worked. I tested it, and must have changed it to self thinking it was fine just before merging it.


leighmcculloch:
TODO:
- [ ] Add the fuzz tests to the `make test` run so that they actually run in CI.

sisuresh (approved):

Issue #1138, opened 2024-12-16
soroban getting started guide broken


leighmcculloch:
The main getting started guide for Soroban/Stellar contracts appears to be broken:
https://developers.stellar.org/docs/build/smart-contracts/getting-started/storing-data

The guide proposes to perform:
```
stellar contract init ./ --with-example increment
```

However, the `--with-example` option was deprecated and removed from the CLI.

cc @stellar/devx

leighmcculloch:
The guide needs updating to use:
```
stellar contract init . --name increment
```

And then instruct the user to copy the code in.

janewang:
Will close after merging https://github.com/stellar/stellar-docs/issues/1035

chadoh:
This is a more comprehensive fix: https://github.com/stellar/stellar-docs/pull/1121

janewang:
> This is a more comprehensive fix: [#1121](https://github.com/stellar/stellar-docs/pull/1121)

Add https://github.com/stellar/stellar-docs/pull/1121 to my PR, also fixed [this one spot](https://github.com/stellar/stellar-docs/pull/1121/files#diff-0cf38822505630f6628223efad614371edaa0489ece6e05d030e1c08fc84a57bR547) that's missed in  https://github.com/stellar/stellar-docs/pull/1121

PR #1162, opened 2025-01-06
Update cli command in getting started guide and closes #1138

janewang:
Also closes https://github.com/stellar/stellar-docs/issues/1035

PR #1185, opened 2025-01-16
Fix/1138 followup

elizabethengelman:
Some additional fixes/updates that i noticed while looking into [1138](https://github.com/stellar/stellar-docs/issues/1138)

getting start
- updated hello world and increment contracts to be the same as they are in the examples repo
- use aliases when deploying so that our first `contract bindings` command works in dapp-frontend.mdx
- add a tip about the --alias flag
- update description of `init` command, and how the `--name` flag works

dapp-frontend.mdx
- have the user copy the contract ids from `.stellar` over into the `irst-soroban-app` directory so we can reuse the deployments from the first secions
- use alias for first `contract bindings` example
- small update to the Counter.astro code

### DIFF
--- a/docs/build/apps/dapp-frontend.mdx
+++ b/docs/build/apps/dapp-frontend.mdx
@@ -79,7 +79,7 @@ stellar contract bindings typescript \
 
 :::tip
 
-Notice that we were able to use the contract alias, `hello-world`, in place of the contract id!
+Notice that we were able to use the contract alias, `hello_world`, in place of the contract id!
 
 :::
 
--- a/docs/build/apps/dapp-frontend.mdx
+++ b/docs/build/apps/dapp-frontend.mdx
@@ -19,11 +19,13 @@ Let's get started.
 
 You're going to need [Node.js](https://nodejs.org/en/download/package-manager/) v18.14.1 or greater. If you haven't yet, install it now.
 
-We want to initialize our current project as an Astro project. To do this, we can clone a template. You can find Soroban templates on GitHub by [searching for repositories that start with "soroban-template-"](https://github.com/search?q=%22soroban-template-%22&type=repositories). For this tutorial, we'll use [stellar/soroban-template-astro](https://github.com/stellar/soroban-template-astro). We'll also use a tool called [degit](https://github.com/Rich-Harris/degit) to clone the template without its git history. This will allow us to set it up as our own git project.
+We want to create an Astro project with the contracts from the previous lesson. To do this, we can clone a template. You can find Soroban templates on GitHub by [searching for repositories that start with "soroban-template-"](https://github.com/search?q=%22soroban-template-%22&type=repositories). For this tutorial, we'll use [stellar/soroban-template-astro](https://github.com/stellar/soroban-template-astro). We'll also use a tool called [degit](https://github.com/Rich-Harris/degit) to clone the template without its git history. This will allow us to set it up as our own git project.
 
-Since you have `node` and its package manager `npm` installed, you also have `npx`. Make sure you're no longer in your `soroban-hello-world` directory and then run:
+Since you have `node` and its package manager `npm` installed, you also have `npx`.
 
-```
+We're going to create a new project directory with this template to make things easier in this tutorial, so make sure you're no longer in your `soroban-hello-world` directory and then run:
+
+```sh
 npx degit stellar/soroban-template-astro first-soroban-app
 cd first-soroban-app
 git init
@@ -56,7 +58,11 @@ This project has the following directory structure, which we'll go over in more 
 └── tsconfig.json
 ```
 
-The `contracts` are the same ones you walked through in the previous steps of the tutorial.
+The `contracts` are the same ones you walked through in the previous steps of the tutorial. Since we already deployed these contracts with aliases, we can reuse the generated contract ID files by copying them from the `soroban-hello-world/.stellar` directory into this project:
+
+```sh
+cp -R ../soroban-hello-world/.stellar/ .stellar
+```
 
 ## Generate an NPM package for the Hello World contract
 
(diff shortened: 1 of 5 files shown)


briwylde08:
Hi @elizabethengelman! Is 1138 an issue somewhere?

elizabethengelman:
> Is 1138 an issue somewhere?

@briwylde08 Good question - I was doing this work as a follow up to [Jane's PR](https://github.com/stellar/stellar-docs/pull/1162) that was addressing https://github.com/stellar/stellar-docs/issues/1138

docs/build/apps/dapp-frontend.mdx
```diff
@@ -56,7 +58,11 @@ This project has the following directory structure, which we'll go over in more
 └── tsconfig.json
 ```
 
-The `contracts` are the same ones you walked through in the previous steps of the tutorial.
+The `contracts` are the same ones you walked through in the previous steps of the tutorial. Since we already deployed these contracts with aliases, we can reuse the generated contract ID files by copying them from the `soroban-hello-world/.stellar` directory into this project:
+
+```sh
+cp -R ../soroban-hello-world/.stellar/ .stellar
```
ElliotFriend, line 64: TIL the difference between `cp -R` and (what I've been using for years) `cp -r` (at least for macOS). Thanks!

docs/build/apps/dapp-frontend.mdx
```diff
@@ -67,17 +73,23 @@ This is going to use the CLI command `stellar contract bindings typescript`:
...
 stellar contract bindings typescript \
   --network testnet \
-  --contract-id $(cat .stellar/contract-ids/hello_world.txt) \
+  --contract-id hello_world \
   --output-dir packages/hello_world
 ```
 
+:::tip
+
+Notice that we were able to use the contract alias, `hello-world`, in place of the contract id!
```
ElliotFriend, line 82:
```suggestion
Notice that we were able to use the contract alias, `hello_world`, in place of the contract id!
```

docs/build/smart-contracts/getting-started/hello-world.mdx
```diff
@@ -60,7 +61,7 @@ members = [
 ]
 
 [workspace.dependencies]
-soroban-sdk = "20.3.2"
+soroban-sdk = "22"
```
ElliotFriend, line 64: Genuine question, because I'm not sure: Are there "best practices" or common patterns that are recommended somewhere for how "granular" `Cargo.toml` dependencies should be?

I just realized you can do this, and it could save me a lot of headaches to just specify the major version. But, I don't know if the tutorial should "buck against" conventional wisdom just for the sake of convenience, **if** there are common/accepted reasons to explicitly specify the full semantic version.

docs/build/smart-contracts/getting-started/hello-world.mdx
```diff
@@ -150,14 +151,14 @@ All contracts should begin with `#![no_std]` to ensure that the Rust standard li
...
-use soroban_sdk::{contract, contractimpl, symbol_short, vec, Env, Symbol, Vec};
+use soroban_sdk::{contract, contractimpl, vec, Env, String, Vec};
 ```
 
 Many of the types available in typical Rust programs, such as `std::vec::Vec`, are not available, as there is no allocator and no heap memory in Soroban contracts. The `soroban-sdk` provides a variety of types like `Vec`, `Map`, `Bytes`, `BytesN`, `Symbol`, that all utilize the Soroban environment's memory and native capabilities. Primitive values like `u128`, `i128`, `u64`, `i64`, `u32`, `i32`, and `bool` can also be used. Floats and floating point math are not supported.
 
 Contract inputs must not be references.
 
-The `#[contract]` attribute designates the Contract struct as the type to which contract functions are associated. This implies that the struct will have contract functions implemented for it.
+The `#[contract]` attribute designates the `HelloContract` struct as the type to which contract functions are associated. This implies that the struct will have contract functions implemented for it.
```
ElliotFriend, line 161: ~~The CLI (at least `v22.2.0`, which I _think_ is the current version) creates a struct called `Contract`.~~

Edit: Never mind, we're talking about the astro template, not the `stellar contract init` command... 🤦🏻‍♂️ 

Edit Edit: Oh, wait, no... This page isn't talking about the astro template, right? We might need to modify instances of `HelloContract` to just `Contract` then?

ElliotFriend (commented):
Looks great! I left a couple minor thoughts/questions, but I think it's pretty close to ready as-is! Thanks @elizabethengelman!!

docs/build/smart-contracts/getting-started/hello-world.mdx
elizabethengelman: That's a good question! My main reason for making this change was so that it would match the version that is currently being used when a user creates a new project with `stellar contract init`. 

In terms of best practices, I think it depends on the dependency you're including in the Cargo.toml and if you're confident that it is using semantic versioning, and won't introduce breaking changes in minor release updates. For the sdk since we know that it does use semantic versioning, I think that we're safe just using the major version, which gives Cargo the ability to use any 22 version.

docs/build/smart-contracts/getting-started/hello-world.mdx
elizabethengelman: Good catch! You're right, the current version of the CLI just has `Contract`. 

I have a branch to update the code that `stellar contract init` creates to make it `HelloContract` so that it matches the soroban-examples repo, but I'll revert this back to `Contract` for now. And we can make this change when the cli change is also merged. I can open 2 new PRs.

elizabethengelman:
@ElliotFriend just made the updates from your review - thank you! I also noticed one more small thing that needed to be fix, and pushed up a commit for that as well.

docs/build/smart-contracts/getting-started/hello-world.mdx
elizabethengelman: For reference, here are the 2 new Prs: https://github.com/stellar/stellar-docs/pull/1239 and https://github.com/stellar/stellar-docs/pull/1185

ElliotFriend (approved):
looks great @elizabethengelman!! Thanks so much!!

Issue #1104, opened 2024-11-28
Cross contract calls  file path error


KoxyG:
### What version are you using?

stellar 22.0.0 (7abcaeb65cf254602a7e0fe6c552d3813bcb5bfe)
stellar-xdr 22.0.0-rc.1.1 (72e523004b5906eb1829990f9b14d2f0fa3018f0)
xdr curr (529d5176f24c73eeccfa5eba481d4e89c19b1181)

### What did you do?

I followed the docs and imported the wasm of contractA
https://developers.stellar.org/docs/build/smart-contracts/example-contracts/cross-contract-call

### What did you expect to see?
I expected to have a successful build of contract B. The wasm importation file path in contractB appears to be wrong.

### What did you see instead?
I saw an error of Client in contractA not found.

briwylde08:
Fixed via https://github.com/stellar/stellar-docs/pull/1208

PR #1208, opened 2025-01-21
Fix test steps #1104

carstenjacobsen:
It's necessary to build Contract A before running the test, otherwise Contract A cannot be called from Contract B.


ElliotFriend (approved):
looks great! thanks!

Issue #1076, opened 2024-11-06
Storage example is missing from contract examples


leighmcculloch:
We used to have a storage example in the docs, but it's not showing up anymore. Most of the examples used progressively build on the storage example. From best I can tell we don't have a simple storage example anymore.

Looking at:
https://developers.stellar.org/docs/build/smart-contracts/example-contracts

![Image](https://github.com/user-attachments/assets/12aced69-5c19-43ef-a6f7-fe8ac3b193b9)

leighmcculloch:
The example still exists in the repo here:
- https://github.com/stellar/soroban-examples/tree/main/increment

briwylde08:
https://github.com/stellar/stellar-docs/pull/1210

PR #1210, opened 2025-01-21
Adding Storage to example contracts #1076

carstenjacobsen:
The Increment smart contract is used. The documentation is based on the Getting Started - Storing Data guide.

### DIFF
--- a/docs/build/smart-contracts/example-contracts/storage.mdx
+++ b/docs/build/smart-contracts/example-contracts/storage.mdx
@@ -81,26 +81,21 @@ Ref: https://github.com/stellar/soroban-examples/tree/v22.0.1/increment
 
 ## How it Works
 
-This contract will get a counter value from storage (or use the value 0 if no value has been stored yet), 
-and increment this counter every time it's called. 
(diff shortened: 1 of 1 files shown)


ElliotFriend (approved):
i like what i'm seeing here. Very nice to have it back among the other examples! If it builds, it gets the thumbs-up from me

@leighmcculloch I know these example contracts are in (more or less) order of increasing complexity/usefulness. Does this location in the sidebar line up for the `increment` example, or should we move it up or down?

leighmcculloch:
> Does this location in the sidebar line up for the `increment` example, or should we move it up or down?

I've move storage up to the top position, as it's one of the simplest entry points. Many of the other examples, like events, custom types, errors, auth, etc, all do storage already.

The oddity in the list in terms of progression is logging, because it actually appears very simple, but it should stay where it is after events, because it's similar to events, but less important than events.

briwylde08:
Merged ^ sorry for being too trigger happy on this one.

leighmcculloch:
Merge fast, it's easy to make these follow up changes 👍🏻

PR #1223, opened 2025-01-22
Add nightly rust requirement to fuzzing how-to guide

leighmcculloch:
### What
Update fuzzing how-to guide to include installation of the nightly Rust toolchain.

### Why
Cargo-fuzz requires the nightly Rust channel, so the documentation now explicitly guides users to install the nightly toolchain before setting up fuzzing. The documentation previously assumed it was already installed which would lead devs to see an error when first running the fuzzer.


ElliotFriend (approved):
Great catch!! Thanks

PR #1224, opened 2025-01-22
Fix fuzzing guide code block formatting

leighmcculloch:
### What
Change the formatting of the code blocks on the fuzzing guide from 'shell' to nothing.

### Why
Shell formatting shouldn't be used for simple commands, otherwise the formatter will highly random arguments that happen to fit shell keywords.

### DIFF
--- a/docs/build/guides/testing/fuzzing.mdx
+++ b/docs/build/guides/testing/fuzzing.mdx
@@ -15,13 +15,13 @@ The following steps can be used in any Stellar contract workspace. If experiment
 
 1. Install `cargo-fuzz`.
 
-   ```
+   ```text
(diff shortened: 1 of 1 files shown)


ElliotFriend (approved):
I'm with you, the weird highlighting sometimes feels... well, weird lol.

The nit-picker in me likes to have _something_ specified, so i typically use `text` as the "language" when i don't want any highlighting. but that just stems from this `markdownlint` plugin for vs code i use, and it flags that there "should always be a language specified." but, honestly, it's more of a preference thing hehe.

leighmcculloch:
I don't mind whether it is text or nothing. What should I do?

leighmcculloch:
I'll change it.

PR #1234, opened 2025-01-27
Create a quick tutorial for a custom SAC admin address

### LINKED ISSUE #1160: Add an example and/or documentation for setting and using a custom admin on a SAC
Slack Thread: https://stellarfoundation.slack.com/archives/C049E67BFSP/p1735830210567219

ElliotFriend:
Closes #1160


docs/build/guides/tokens/custom-sac-admin.mdx
```diff
@@ -0,0 +1,161 @@
+---
+title: Set a custom admin account for a Stellar Asset Contract (SAC)
+hide_table_of_contents: true
+description: Set a custom administrator account on a deployed SAC
+---
+
+The [Stellar Asset Contract (SAC)](../../../tokens/stellar-asset-contract.mdx) includes functionality to set a custom administration account on the contract itself. This can be seen as a security precaution, similar to locking an asset's issuer account after creating the total supply of tokens. It should also be noted that an asset's issuer account can have a distinct set of signatures and weights, and these are unrelated to the SAC admin address.
```
dmkozh, line 7: I don't think this can be seen as security precaution. The issuer account is not locked in this case and it can keep miniting/burning/clawbacking tokens (if enabled). We can say that the issuer account _can_ be locked thus moving the token administration _exclusively_ to Soroban. But I don't think this has anything to do with security precautions. Basically, it's important to note that changing the admin does not automatically revoke any permissions from the original issuer.

docs/build/guides/tokens/custom-sac-admin.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+## Considerations
+
+To effectively utilize the SAC admin functionality, you should be aware of a few things first:
+
+- When a SAC is initially enabled for an issued asset, the `Admin` address on the contract defaults to the issuer account.
+- A SAC `Admin` address can be either a regular Stellar account (`G...`), or it can be a smart contract (`C...`) address. This opens up the possibility of programmatically minting and taking other administrative actions for an asset by way of a smart contract invocation.
+- Changing a SAC `Admin` address (initially) requires authorization from the asset's issuer account. After the `Admin` address has been set the first time, subsequent changes will require authorization from the _current_ `Admin` address.
```
dmkozh, line 15: Instead of specifying the exceptions, we can just say that the admin change must be authorized by the current admin of the token. When token is first instantiated, its admin is set to its issuer.

docs/build/guides/tokens/custom-sac-admin.mdx
```diff
@@ -0,0 +1,161 @@
...
+For this part of the example, we'll use a very simple "minter" contract as the `Admin` address. It has a very simple implementation, and only one main function, `start_mint`:
+
+```rust
+#[contractimpl]
+impl Contract {
+    pub fn __constructor(env: Env, sac_address: Address) {
+        env.storage().instance().set(&symbol_short!("SAC_ADDR"), &sac_address);
+    }
+
+    pub fn star_mint(env: Env, to: Address, amount: i128) {
```
dmkozh, line 103: Even though this is just a simple example,  I still don't think it's ever a good idea to post examples of permissionless operations. We should at least leave a disclaimer about that.

docs/build/guides/tokens/custom-sac-admin.mdx
ElliotFriend: you're right. i think i was making an unfounded connection between changing the SAC admin and modifying the weights/signers for the issuer `G...` address. I'm cleaning up this language in a new PR

docs/build/guides/tokens/custom-sac-admin.mdx
ElliotFriend: good call. i'll try to improve this part of the document, and will probably ask you for some help and feedback about what i put together

docs/build/guides/tokens/custom-sac-admin.mdx
ElliotFriend: I did want to be a little more verbose in regard to this, though. When a dev is first considering a custom SAC admin, I think it's a bit counter-intuitive about having an issuer account that can mint/clawback, as well as a totally unrelated SAC admin that can do the same things. So, spending a bit more time and re-iterating in a couple ways seemed like a good approach to me.

That said, these points can be cleaned up and refined, for sure!

ElliotFriend: On second thought, I may have misunderstood what you were getting at hehe. Yes, line 15 was fairly messy, and a much cleaner version will be in the upcoming PR

PR #1245, opened 2025-01-31
refine and clarify some points in the custom SAC admin guide

ElliotFriend:
Addresses the comments from the initial publishing of this page:

- incorrectly describing a custom SAC admin as a "security precaution" ([link to comment](https://github.com/stellar/stellar-docs/pull/1234#discussion_r1936007543))
- unclear and overly wordy description of which admin needs to authorize when ([link to comment](https://github.com/stellar/stellar-docs/pull/1234#discussion_r1936037753))
- using an example that would result in a "permissionless operation" ([link to comment](https://github.com/stellar/stellar-docs/pull/1234#discussion_r1936055012))

Refs: #1234

### DIFF
--- a/docs/build/guides/tokens/custom-sac-admin.mdx
+++ b/docs/build/guides/tokens/custom-sac-admin.mdx
@@ -27,9 +27,21 @@ Depending on the use-case, it can be beneficial for asset issuers to mint an ent
 - If an asset's issuer account is locked while the SAC `Admin` address has not been changed, it will _never_ be possible to change the SAC `Admin` address. This is because the issuer account can no longer authorize the first `set_admin` invocation.
 - If an asset's issuer account is locked after the SAC `Admin` address has been changed, tokens will still be mint-able and/or clawback-able (if enabled) from the SAC as long as it's authorized by the _current_ `Admin` address.
 
+:::danger
+
+When changing a SAC `Admin` address, the new admin address provided is not validated at that time. This means you can lock down administration from the SAC **forever**. Consider your choice of SAC `Admin` address carefully and thoroughly.
+
+:::
+
 ## Example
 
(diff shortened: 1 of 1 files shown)


docs/build/guides/tokens/custom-sac-admin.mdx
```diff
@@ -91,7 +102,15 @@ If you examine the asset's `GCS5...` issuer account, you won't see anything diff
 
 ### Set the `Admin` address to some contract `C...` address
 
-For this part of the example, we'll use a very simple "minter" contract as the `Admin` address. It has a very simple implementation, and only one main function, `start_mint`:
+For this part of the example, we'll use a very simple "minter" contract as the `Admin` address. It has a very simple implementation, and only one main function, `start_mint`.
+
+:::warning
+
+The following "minter" contract calls the `require_auth()` function for itself. This means you would need to include some `__check_auth` logic in the contract, which is demonstrated in the [custom account example contract](../../smart-contracts/example-contracts/custom-account.mdx). We'll omit those details here, for the sake of simplicity and brevity.
```
dmkozh, line 109: I think this is way too nuanced and really not related to the topic. This makes the example incomplete without check_auth, and it doesn't really demonstrate something close to reality (I don't expect token contracts to also be custom accounts).

I can think of the following simple example that's not completely permissionless, but also demonstrates something interesting: consider a `claim_airdrop` function (or something along these lines), that mints some small fixed amount to a given address. It would take the receiver address, call require_auth for it in order to make sure it's a real address, then make sure the address hasn't claimed before (using storage), then mint some amount of token to it. This shouldn't add more than a couple lines of code to the example (there is no need to even use some UDTs for the keys, just add the addresses directly to the storage for brevity).

ElliotFriend: this is perfect! i'll use that instead. thanks!

dmkozh (approved):

PR #1263, opened 2025-02-07
Run the RPC spec bundler

Shaptic:
A few updates have been made to the specs but they haven't been reflected in the bundled spec (`yarn rpcspec:build`).


ElliotFriend (approved):
This is a great practice! Thanks!

To be sure, though. the `yarn rpcspec:build` command is run at build-time, so all the changes "should" still be reflected in the published version of the file. All the same, it's really good to keep the committed file up-to-date.

Thanks!

Issue #1177, opened 2025-01-13
[Soroban] Cargo test does not work in Soroban examples


lijamie98:
### What version are you using?
Stellar CLI `22.1.0`

### What did you do?

In https://developers.stellar.org/docs/build/smart-contracts/example-contracts/custom-types#run-the-contract,

`stellar contract invoke –wasm` does not working.

### What did you expect to see?

`stellar contract invoke –wasm` should invoke the smart contract.

### What did you see instead?

```
jamie ~/projects/soroban-examples/custom_types (tags/v22.0.1) % stellar contract invoke \
    --wasm target/wasm32-unknown-unknown/release/soroban_custom_types_contract.wasm \
    --id 1 \
    -- \
    increment \
    --incr 5
error: unexpected argument '--wasm' found

  tip: to pass '--wasm' as a value, use '-- --wasm'

Usage: stellar contract invoke [OPTIONS] --id <CONTRACT_ID> --source-account <SOURCE_ACCOUNT> [-- <CONTRACT_FN_AND_ARGS>...]

For more information, try '--help'.

```

carstenjacobsen:
Resolved by #1269

PR #1269, opened 2025-02-11
Fixed invoke example #1177

carstenjacobsen:
The invoke command used the `--wasm` parameter, which is not supported.


briwylde08 (approved):

anataliocs (approved):
🚢

PR #1273, opened 2025-02-11
Move Sora details into a note on CLI page

leighmcculloch:
### What
Move the Sora details into a note block on the CLI documentation page rather than being a section.

### Why
It's somewhat confusing going to the page and seeing Sora listed under CLI. It makes it sound like the Sora tool is another CLI, when it's a GUI ontop of the CLI. A note looks to me like a good way to present that information. The page is primarily about the Stellar CLI, and notes are a good way to present ancillary information that take the user away to other related things.

### DIFF
--- a/docs/tools/developer-tools/cli/README.mdx
+++ b/docs/tools/developer-tools/cli/README.mdx
@@ -1,12 +1,10 @@
 ---
-title: CLI
-description: Explore a suite of developer tools for Stellar, including CLI.
+title: Stellar CLI
+description: The CLI for interacting with the Stellar network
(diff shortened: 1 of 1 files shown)


jeesunikim (approved):

briwylde08:
@leighmcculloch that was a long-winded answer but in this particular case I would just add a little language, which I did here: https://github.com/stellar/stellar-docs/pull/1275

leighmcculloch:
Approved the PR, thanks, and 👍🏻 some language or way to signal products are external would be helpful and improve how projects are framed in general.

PR #1291, opened 2025-02-18
Update llvm-cov command syntax in fuzzing docs

leighmcculloch:
### What
Modify llvm-cov command syntax and fix formatting in fuzzing documentation.

### Why
The existing command was valid, but longer than necessary


ElliotFriend (approved):
thanks Leigh!!! looks great to me!

PR #1298, opened 2025-02-21
Update `SorobanRpc` to `rpc` and use `StellarRpc`

jeesunikim:
1. js-stellar-sdk [v13.0.0](https://github.com/stellar/js-stellar-sdk/releases/tag/v13.0.0) removed `SorobanRpc`. 

> The SorobanRpc import, previously deprecated, has been removed. You can import rpc instead:

2. We rebranded `SorobanRpc` to `StellarRpc` in November 2024

This PR is to update the doc accordingly.

### DIFF
--- a/docs/build/guides/archival/restore-data-js.mdx
+++ b/docs/build/guides/archival/restore-data-js.mdx
@@ -32,7 +32,7 @@ import {
   Keypair,
   TransactionBuilder,
   SorobanDataBuilder,
-  SorobanRpc,
+  rpc as SorobanRpc,
   xdr,
 } from "@stellar/stellar-sdk"; // add'l imports to submitTx
 const { Api, assembleTransaction } = SorobanRpc;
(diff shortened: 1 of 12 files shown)


Shaptic (approved):
I don't get why you'd do `rpc as StellarRpc` instead of just using `rpc`, but I don't have an opinion either way :) LGTM!

jeesunikim:
@Shaptic so i didn't wanna change the entire existing `SorobanRpc` to `rpc` so I started as only to update `rpc as SorobanRpc` and leave the rest as it is then I realized we renamed `SorobanRpc` to `StellarRpc` so I updated that then I realized 'oh I could've just used rpc as it is..' okay maybe i will just use rpc

Shaptic:
@jeesunikim either way works for me! But to be clear we didn't rename `SorobanRpc` -> `StellarRpc` _in the SDKs_, just as a "product name". In the SDK it's only `rpc`.

jeesunikim:
yep i am aware!

PR #1346, opened 2025-03-05
Feature branch/platform docpoc25

karthikiyer56:
NOT TO BE MERGED until all code for docpoc25 is ready for public viewing


docs/data/ingest-library/README.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+The library is a programmatic SDK in the Go programming language composed of several packages for acquiring and parsing data from the Stellar network. It provides language level bindings which convert the [binary XDR encoded](https://developers.stellar.org/docs/learn/encyclopedia/data-format/xdr) streams emitted from the network into fluent programmatic data model bindings.
+
+## Why use the Ingest Library?
+
+Applications can leverage the library to rapidly develop ingestion pipelines capable of acquiring real-time or historical Stellar network data and deriving custom data models. The library enables applications to traverse the hierarchal data structures of the network: [history archives](https://developers.stellar.org/docs/validators/admin-guide/environment-preparation#history-archives), [ledgers](https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/ledgers), transactions, operations, ledger state changes, and events.
+
+Use the library for an intuitive, compile-time, type-safe developer experience to work with the main types of network data:
+
+### Ledger Entries
```
urvisavla, line 16: This tells you about getting two kinds of data (entries and metadata) that you can get from the ingest library which is a good start but it doesn't explain when to use each one. It looks like Ledger Metadata has everything, so why would anyone use Ledger Entries from the history archives? Some specific examples of what you can get from Ledger Entries would be really useful. Also, details like BucketEntry are probably too much for introduction page.

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,213 @@
...
+type ChangesInfo struct {
+	LedgerEntriesCreated    int32
+	LedgerEntriesUpdated    int32
+	LedgerEntriesDeleted    int32
+	FeeRelatedChanges       int32
+	TxRelatedChanges        int32
+	OperationRelatedChanges int32
+}
+
+func changeCausedBy(change ingest.Change) xdr.LedgerEntryChangeType {
```
urvisavla, line 76: This can be remove after protocol-23 branch is merged into main -  https://github.com/stellar/go/blob/protocol-23/ingest/change.go#L46

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,127 @@
+---
+title: Get statistics from a ledger range
```
urvisavla, line 2: The term "statistics" feels a bit vague here. It would be clearer if the title specified the exact data like for eg "Get successful/failed transactions breakdown for a ledger range"

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,213 @@
+---
+title: Get changes from a ledger range
```
urvisavla, line 2: Similar to the comment below, I think it would be more helpful if the title specified the exact data this example provides.

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,129 @@
+---
+title: Get ledger entries from History Archive
```
urvisavla, line 2: Similar to the comment below, I think it would be more helpful if the title specified the exact data this example provides.

docs/data/ingest-library/README.mdx
sreuland: I'll link `ledger entries` to https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/ledgers

and add a second sentence to reference real life examples of ledger entries such as accounts, liquidity pools, contract data.

sreuland: done - https://github.com/stellar/stellar-docs/pull/1346/commits/02aead1d7e0d76a382240e335a6394cde5e2a3a9

docs/data/ingest-library/README.mdx
```diff
@@ -0,0 +1,38 @@
+---
+title: Ingest Library Introduction
+sidebar_position: 0
+---
+
+## What is the Ingest Library?
```
sreuland, line 6: Should we catalog the list of go packages that comprise the ingest library here for completeness?

We haven't mentioned release tags as a convention for importing the ingest library into a project yet. Should we defer on notion of release for now, and add reference in here to use [go/releases](https://github.com/stellar/go/releases) once we start tagging for ingest library?

docs/data/ingest-library/examples/changes.mdx
sreuland: could you add a suggestion with what you have in mind?

docs/data/ingest-library/examples/changes.mdx
sreuland: Will the current code be forwards compatible with the p23 changes? 
sounds like no-op for now and a follow on ticket to track updating this for clarity. 
This is somewhat related to notion of releases of ingest library also, [commented on that here](https://github.com/stellar/stellar-docs/pull/1346/files#r1992461225).

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+```
+
+</CodeExample>
+
+**Sample Response:**
+
+```bash
+>> go run ./change_entries.go
```
sreuland, line 155: @karthikiyer56 , is a `go.mod` needed to run these examples? I think the imports on external pkgs like  `github.com/stellar/go` will need it, maybe can inline the go.mod in each example here or up in `examples/README.mdx` as a pre-req?

karthikiyer56: Good idea. 
Didn't think about that 
If you are in the vicinity, go ahead, or else I'll make the change tonight

docs/data/ingest-library/examples/changes.mdx
karthikiyer56: Agreed 
We can mark it no-op for now

docs/data/ingest-library/README.mdx
karthikiyer56: Thats true. And perhaps, as a part of p23 release, we could start a new tag name convention for the release library

docs/data/ingest-library/examples/changes.mdx
urvisavla: It's fine to treat this as a no-op, but the main point is why introduce something new that will soon require an update.
Alternatively, does this need to display the created, deleted and updated entry counts to demonstrate how "change" entries can be used? Or is there other useful information that can be derived from the 'change' and shown here instead?

karthikiyer56: The idea was to get the user familiar with the change struct.
To that end, i simply put together a sample example that highlights the usage of the change struct, without going through too many nuances. 
I had half a mind to write a account balance diff example that goes through updated accounts and shows their balance pre and post, but that might be overkill. 
I find there to be merit in including an example that shows number of entries updated/deleted etc. 
Even when this addition of `ChangeType` goes out in p23, it doenst necessarily negate the correctness of the `func changeCausedBy`, right? It merely makes it redundant. 

Which is why i am of the opinion that we let it go in the way it is, and then rework it to use `ChangeType` field after p23 is released.

docs/data/ingest-library/examples/transaction_statistics.mdx
karthikiyer56: replaced with `Get successful/failed transactions from a ledger range`

docs/data/ingest-library/examples/changes.mdx
karthikiyer56: done
I couldnt come up with anything better than `Retrieve ledger entry changes within a range`

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done.
replaced with `Get count of each ledger entry type from history archive`
i couldnt come up with anything better here either

docs/data/ingest-library/examples/changes.mdx
karthikiyer56: done.
add a simple example of `go.mod` in the examples/README.mdx

docs/data/ingest-library/examples/changes.mdx
sreuland: ok, if it's compatible with current and going forwards after the p23 merge, that means it's always working, provides time to take on later enhancements w/o rush, thanks!

urvisavla (approved):

PR #1349, opened 2025-03-06
Add an example usage for ingest library to fetch statistics

karthikiyer56:


docs/data/ingest-library/examples/README.mdx
```diff
@@ -0,0 +1,12 @@
...
+sidebar_position: 10
+---
+
+This section showcases real-world use cases of the Ingest Library, focusing on key components like `ingest.LedgerTransaction`, `ingest.LedgerChange`, `ingest.LedgerTransactionReader`, and `ingest.ChangeReader`.
+
+These examples either taken in a Stellar Ledger (`xdr.LedgerCloseMeta`) or a ledger sequence range (start-end). Each of the examples listed here is complete, and can be taken as-is, compiled and run in your IDE/local dev.
+
+## Prerequisites
+
+Some of the examples listed here might require a local instance of captive-core. At a minimum, you will need a `stellar-core` binary that has been compiled for your platform. Refer "admin guide" for more details.
```
karthikiyer56, line 12: "admin_guide" will point to the admin guide once that work is done. For now, this shud suffice

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,136 @@
...
+---
+
+This example connects to the Stellar History Archive using `historyarchive.Connect` and initializes a checkpoint reader with `ingest.NewCheckpointChangeReader` to read ledger entry changes at a given checkpoint. It iterates over the ledger changes, processing different entry types such as `Account`, `ClaimableBalance`, `Trustline`, and `Offer`, using the `ingest.Change` structure. The code tracks statistics in an `entriesInfo` struct, counting the total entries and specific types.
+
+<CodeExample>
+
+```go
+// Filename: example_ledger_entries.go
+
+package main
```
karthikiyer56, line 13: Reworked the code examples from the go monorepo for both these examples quite a bit.
Made it so that they can be copy-pasted as-is to make them work

docs/data/ingest-library/examples/README.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: Example Usages
+sidebar_position: 10
+---
+
+This section showcases real-world use cases of the Ingest Library, focusing on key components like `ingest.LedgerTransaction`, `ingest.LedgerChange`, `ingest.LedgerTransactionReader`, and `ingest.ChangeReader`.
+
+These examples either taken in a Stellar Ledger (`xdr.LedgerCloseMeta`) or a ledger sequence range (start-end). Each of the examples listed here is complete, and can be taken as-is, compiled and run in your IDE/local dev.
```
sreuland, line 8:
```suggestion
These examples take a Stellar Ledger (`xdr.LedgerCloseMeta`) or a ledger sequence range (start-end) as input. Each of the examples listed here is complete, and can be compiled and run as-is in your IDE/local dev.
```

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,136 @@
...
+    NumClaimableBalances int32
+    NumOffers            int32
+}
+
+func ledgerEntryStats() {
+    archiveURLs := network.TestNetworkhistoryArchiveURLs
+    networkPassphrase := network.TestNetworkPassphrase
+
+    // Open a history archive using our existing configuration details.
+    historyArchive, err := historyarchive.Connect(
```
sreuland, line 49: Using [ArchivePool](https://github.com/stellar/go/blob/master/historyarchive/archive_pool.go#L35) would be best convention when configuring from a list of urls, probably less code as well, such as:

```
historyArchive, err := NewArchivePool(archiveURLs, ArchiveOptions{})
```

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,136 @@
...
+        },
+    )
+    panicIf(err)
+
+    // First, we need to establish a safe fallback in case of any problems
+    // during the history archive download+processing, so we'll set a 30-second
+    // timeout.
+    //
+    // NOTE: We're using the testnet here, whose archives are much smaller. For
+    //       the pubnet, a 30 *minute* timeout may be more appropriate.
```
sreuland, line 66: Would be nice to just remove lines 65-66 if possible. A 30 minute delay sounds rough, like something is wrong and may discourage people from considering it.

Can you try this with pubnet archive urls for sequence=123455 and also for a recent ledger sequence from this year and see how it performs? 123455 is very old ledger sequence, so, the archive file from pubnet should be just as small.

It may also be worthwhile to use pubnet instead of testnet in general because testnet gets reset occasionally, and the checkpoint ledger used in this example won't exist until a reset network is around at least 7 days old.

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,138 @@
+---
+title: Get statistics from a ledger range
+sidebar_position: 10
+---
+
+This example illustrates how to connect to a Stellar history archive using the `ledgerbackend.CaptiveCoreConfig`, prepare a range of ledgers for ingestion, and then processes transactions within those ledgers. It reads each ledger's transactions using the `ingest.LedgerTransactionReader`, categorizes them as successful or failed, and tracks the operations associated with each transaction.
```
sreuland, line 6:
```suggestion
This example illustrates how to run and connect with a Stellar network watcher node, referred to as 'captive core' using the `ledgerbackend.CaptiveStellarCore`. It then requests a historical bounded range of ledgers to be replayed. The captive core instance will emit a stream of `LedgerCloseMeta` which contains the transactions per each ledger in the range. It reads each ledger's transactions using the `ingest.LedgerTransactionReader`, categorizes them as successful or failed, and tracks the operations associated with each transaction.
```

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/support/log"
+)
+
+var (
+	config = captiveCoreConfig()
+)
+
+func captiveCoreConfig() ledgerbackend.CaptiveCoreConfig {
+	archiveURLs := network.TestNetworkhistoryArchiveURLs
```
sreuland, line 32: Similar to [prior question](https://github.com/stellar/stellar-docs/pull/1349/files#r1984131836), can this use pubnet? The requested ledger range is very old and catch up time by the core instance should be minimal.

sreuland (approved):
left some comments for consideration, lgtm overall. I didn't verify running the example code.

docs/data/ingest-library/examples/transaction_statistics.mdx
karthikiyer56: I have no opinions one way or another on this.
i am simply inclined to accept your suggestions and make changes.
incoming in a bit

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,135 @@
...
+    defer cancel()
+
+    // We pass 123455 because given a checkpoint frequency of 64 ledgers (the
+    // default in `ConnectOptions`, above), 123455+1 mod 64 == 0. Incompatible
+    // sequence numbers will likely result in 404 errors.
+    var reader *ingest.CheckpointChangeReader
+    reader, err = ingest.NewCheckpointChangeReader(ctx, historyArchive, 123455)
+    panicIf(err)
+
+    accountsToTrustlineCountMap := make(map[xdr.AccountId]int32)
```
Shaptic, line 77: nitpicky, but maybe `trustlinesPerAccount` is a better variable name?

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+        info.TotalEntries++
+
+        switch entry.Type {
+        case xdr.LedgerEntryTypeClaimableBalance:
+            info.NumClaimableBalances++
+        case xdr.LedgerEntryTypeAccount:
+            info.NumAccounts++
+        case xdr.LedgerEntryTypeTrustline:
+            // The NewCheckpointChangeReader.read() function converts each ledgerEntry into a Change struct where Post is non empty
```
Shaptic, line 96: I don't think users of this lib should care about this level of detail (`xdr.LedgerEntryChange` -> `ingest.Change`). This comment is more confusing than elucidating imo.

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,138 @@
...
+	networkPassphrase := network.TestNetworkPassphrase
+	captiveCoreToml, err := ledgerbackend.NewCaptiveCoreToml(ledgerbackend.CaptiveCoreTomlParams{
+		NetworkPassphrase:  networkPassphrase,
+		HistoryArchiveURLs: archiveURLs,
+	})
+	panicIf(err)
+
+	return ledgerbackend.CaptiveCoreConfig{
+		// Change these based on your environment:
+		BinaryPath:         "/usr/local/bin/stellar-core", // Location of stellar-core binary compiled for your platform
```
Shaptic, line 42: You can use the same [`$PATH` method](https://stackoverflow.com/a/46236792) we use in the code here so it's not hardcoded for Linux

docs/data/ingest-library/examples/transaction_statistics.mdx
Shaptic: Suggestion^2 on top of @sreuland's:

```diff
- The captive core instance will emit a stream of `LedgerCloseMeta`
+ The Captive Core instance will emit a stream of ledger metadata.
```

docs/data/ingest-library/examples/README.mdx
tamirms, line 12: The captive-core terminology might be confusing for newcomers. I would rephrase as:

```suggestion
Some of the examples listed here might invoke the `stellar-core` binary. Refer to the "admin guide" for more details on how to compile `stellar-core` for your platform.
```

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		panic(fmt.Errorf("An error occurred, panicking: %s\n", err))
+	}
+}
+
+func getChangesFromLedger(passphrase string, ledger xdr.LedgerCloseMeta) []ingest.Change {
+	changeReader, err := ingest.NewLedgerChangeReaderFromLedgerCloseMeta(passphrase, ledger)
+	changes := make([]ingest.Change, 0)
+	defer changeReader.Close()
+	panicIf(err)
```
tamirms, line 58: if there is an error then `changeReader` then `panicIf(err)` will trigger a panic but the deferred function will [still be invoked](https://go.dev/blog/defer-panic-and-recover). It doesn't make sense to close the change reader if `err` is not nil because the change reader is only well formed if `err` is nil.

```suggestion
	panicIf(err)
	defer changeReader.Close()
```

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,229 @@
...
+func changeCausedBy(change ingest.Change) xdr.LedgerEntryChangeType {
+	if change.Pre == nil && change.Post != nil {
+		return xdr.LedgerEntryChangeTypeLedgerEntryCreated
+	} else if change.Pre != nil && change.Post == nil {
+		return xdr.LedgerEntryChangeTypeLedgerEntryRemoved
+	} else if change.Pre != nil && change.Post != nil {
+		return xdr.LedgerEntryChangeTypeLedgerEntryUpdated
+	}
+	// this should not happen
+	return xdr.LedgerEntryChangeType(-1)
```
tamirms, line 108: to be consistent with the rest of the code you could panic here

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,135 @@
+---
+title: Get ledger entries from History Archive
+sidebar_position: 20
+---
+
+This example connects to the Stellar History Archive using `historyarchive.Connect` and initializes a checkpoint reader with `ingest.NewCheckpointChangeReader` to read ledger entry changes at a given checkpoint. It iterates over the ledger changes, processing different entry types such as `Account`, `ClaimableBalance`, `Trustline`, and `Offer`, using the `ingest.Change` structure. The code tracks statistics in an `entriesInfo` struct, counting the total entries and specific types.
```
tamirms, line 6: "initializes a checkpoint reader" -> "initializes a checkpoint change reader"

"to read ledger entry changes at a given checkpoint" -> "to read ledger entry changes from a history archive snapshot occurring at a given checkpoint ledger"

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: So, I deliberated over this and decided that the best course of action is to leave this in as-is, with the comments.
I did the exercise with pubnet and it takes 5sh mins to run this on my machine before the user can see anything.
My concern here is that the UX for someone who just wants to copy-paste this go code as-is and run it to see a sample output. might be hindered if I make it point to pubnet in this example

karthikiyer56: > until a reset network is around at least 7 days old.

Isnt checkpoint ledger generatead every 64 ledgers?

karthikiyer56: never mind, I think you meant that ledger 123455 wont be available up until 7 days.

karthikiyer56: Ima just point to pubnet and leave a comment that says - "this example pooints to pubnet. This might take a while...."

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
tamirms, line 77: I don't think a `map[xdr.AccountId]int32` will work as you expect. The `AccountId` struct contains a pointer:

```go
type AccountId PublicKey

type PublicKey struct {
	Type    PublicKeyType
	Ed25519 *Uint256
}
```

This means that two account ids could store equivalent addresses but have differing pointers and therefore be stored as separate entries in your map.

karthikiyer56: yup, already reworking that one too

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,135 @@
...
+    )
+    panicIf(err)
+
+    // First, we need to establish a safe fallback in case of any problems
+    // during the history archive download+processing, so we'll set a 30-second
+    // timeout.
+    //
+    // NOTE: We're using the testnet here, whose archives are much smaller. For
+    //       the pubnet, a 30 *minute* timeout may be more appropriate.
+    ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
tamirms, line 67: I would get rid of the context timeout in case it is not high enough which would be frustrating in case the code is making progress iterating through the history archive snapshot instead being stuck indefinitely. a better approach would be to log progress and allow the user to ctrl-c if it looks like the reader is stuck

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+	fmt.Println("\nDone. Results:")
+	fmt.Printf("  - total transactions: %d\n", successfulTransactions+failedTransactions)
+	fmt.Printf("  - succeeded / failed: %d / %d\n", successfulTransactions, failedTransactions)
+	fmt.Printf("  - total operations:   %d\n", operationsInSuccessful+operationsInFailed)
+	fmt.Printf("  - succeeded / failed: %d / %d\n", operationsInSuccessful, operationsInFailed)
+}
+
+func main() {
+	statistics()
```
tamirms, line 120: I think it's better to inline `statistics()` into the main function. It seems odd to have a main function consist of a single line which calls another function

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
```diff
@@ -0,0 +1,135 @@
...
+    fmt.Println()
+    // Marshal the struct into JSON with indentation
+    jsonData, err := json.MarshalIndent(&info, "", "    ")
+    panicIf(err)
+    // Print the indented JSON
+    fmt.Println(string(jsonData))
+}
+
+func main() {
+    ledgerEntryStats()
```
tamirms, line 117: I think it's better to inline ledgerEntryStats() into the main function. It seems odd to have a main function consist of a single line which calls another function

docs/data/ingest-library/examples/README.mdx
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done.
removed the timeout logic, and the comment about pubnet, and made it actually point to pubnet.
Added a note saying "pubnet can take long..."

docs/data/ingest-library/examples/transaction_statistics.mdx
karthikiyer56: incorporated both the recommendations

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done.
removed the comment

docs/data/ingest-library/examples/README.mdx
karthikiyer56: done

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,229 @@
+---
+title: Get changes from a ledger range
+sidebar_position: 30
+---
+
+This code connects to a Stellar history archive using a `ledgerbackend.CaptiveCoreConfig` and retrieves ledger data between a specified range of ledger sequences. It uses the `ingest.LedgerChangeReader` to read changes from each ledger (`xdr.LedgerCloseMeta`) and categorizes the changes based on whether they represent created, updated, or deleted ledger entries. The `ingest.Change` structure is used to capture individual changes, and the code tracks various types of changes such as those related to fees, transactions, or operations.
```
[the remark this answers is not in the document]
karthikiyer56: reworded it and removed the mention of connecitng to stellar history archive

docs/data/ingest-library/examples/changes.mdx
karthikiyer56: done.
moved up and down

docs/data/ingest-library/examples/changes.mdx
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/transaction_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/transaction_statistics.mdx
karthikiyer56: I hear you, but I'd much rather not get clever here.
IMO, there is merit in making sure that the person using this code is cognizant of the place where their stellar-core binary is and change it accordingly.
tl;dr - I am not in favor of using the `exec.LookPath` to find out where "stellar-core" binary might actually live

docs/data/ingest-library/examples/changes.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+func main() {
+	startingLedger := uint32(123622) // Replace with your desired ledger number
+	endLedger := startingLedger + 5
+
+	// Get the ledger metadata
+	ledgers := getLedgers(startingLedger, endLedger)
+
+	for seq, ledgerMeta := range ledgers {
+		fmt.Printf("Processing ledger: %d\n", seq)
```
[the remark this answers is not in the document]
karthikiyer56: done

docs/data/ingest-library/examples/ledger_entry_statistics.mdx
karthikiyer56: done

docs/data/ingest-library/examples/transaction_statistics.mdx
```diff
@@ -0,0 +1,127 @@
...
+	for seq := startingSeq; seq <= startingSeq+ledgersToRead; seq++ {
+		fmt.Printf("Processed ledger %d...\r", seq)
+
+		var txReader *ingest.LedgerTransactionReader
+		var err error
+		txReader, err = ingest.NewLedgerTransactionReader(
+			ctx, backend, config.NetworkPassphrase, seq,
+		)
+		panicIf(err)
+		defer txReader.Close()
```
tamirms, line 82: it would be better style to avoid calling `defer` repeatedly in a loop because that will keep pushing more items to the defer stack. Then when `main()` exits all the `defer` calls from the for loop will have to be executed. A better option would be to ensure `txReader.Close()` is called at the end of each iteration.

karthikiyer56: addressed here: https://github.com/stellar/stellar-docs/pull/1379

PR #1353, opened 2025-03-06
Add openzepplin contract library description

janewang:
http://localhost:3000/docs/tools/developer-tools/openzepplin-contracts


briwylde08:
There are a few places where it's not spelled correctly. It's spelled OpenZeppelin with a second "e".

Issue #1309, opened 2025-01-30
RPC schema: Pagination declared twice


FrankSzendzielarz:
Pagination is declared twice in the openrpc spec, once in GeteventsParams and once in GettransactionsParams. Confuses codegen. Probably best to make that a ref

PR #1363, opened 2025-03-06
Fix pagination documentation for stellar-rpc

tamirms:
Fixes https://github.com/stellar/stellar-docs/issues/1309

This PR consolidates the documentation for pagination that is used for getEvents, getTransactions, and getLedgers. All those endpoints use a `cursor` field in the request / response in order to do pagination.

The `cursor` field used to be called `paging_token` and this PR updates the documentation to reference `cursor` instead of `paging_token`.

Also, the `cursor` field was missing in the schema of the `getEvents` and `getTransactions` responses. This PR includes the `cursor` fields in those responses.

### DIFF
--- a/openrpc/src/stellar-rpc/methods/getEvents.json
+++ b/openrpc/src/stellar-rpc/methods/getEvents.json
@@ -17,7 +17,7 @@
             "$ref": "#/components/contentDescriptors/EventFilters"
         },
         {
-            "$ref": "#/components/contentDescriptors/EventsPagination"
+            "$ref": "#/components/contentDescriptors/Pagination"
         },
         {
             "$ref": "#/components/contentDescriptors/Format"
@@ -39,7 +39,7 @@
                 },
                 "cursor": {
                     "type": "string",
-                    "$ref": "#/components/schemas/GetEventsCursor"
+                    "$ref": "#/components/schemas/Cursor"
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


openrpc/src/stellar-rpc/methods/getEvents.json
```diff
@@ -39,7 +39,7 @@
                 },
                 "cursor": {
                     "type": "string",
-                    "$ref": "#/components/schemas/GetEventsCursor"
+                    "$ref": "#/components/schemas/Cursor"
```
Shaptic, line 42: Each event itself also has a cursor - I think that needs to be updated, too?

Shaptic (commented):
Nice! You need to regenerate the "compiled" RPC spec, though:

```
yarn rpcspec:build
```

openrpc/src/stellar-rpc/methods/getEvents.json
tamirms: in the event schema the field is still called `pagingToken`:

https://github.com/stellar/stellar-rpc/blob/main/protocol/get_events.go#L29-L30

tamirms:
@Shaptic can you take another look? I updated the pr with the following changes:

- ran `yarn rpcspec:build`
- added a note to the `pagingToken` field within the event schema that it's deprecated in favor of the top level cursor
- fixed the link to the pagination page (previously it was hardcoded to the developers.stellar.org domain)

openrpc/src/stellar-rpc/methods/getEvents.json
Shaptic: Oh oof, gotcha. That one is still useful for mid-page eventing, though, so idk if deprecated is the word? Then again they should use `id` for that.

Shaptic (approved):

PR #1387, opened 2025-03-10
Fix invalid cli version in install guide

Ifropc:


fnando:
@Ifropc i think we should really fix the issue, instead of doing temporary workarounds. My proposal is to add a file to the release page, fetch that from this repo. This way, we can replace the file whenever we need, without having to message with anything else.

Ifropc:
There's an issue for docs validation here https://github.com/stellar/stellar-cli/issues/1908
We should remove tmp fixes before next version release, I agree

fnando (approved):

Issue #1394, opened 2025-03-11
Add documentation for initial backfill of hubble dataset


amishas157:
### What problem does your feature solve?
Fill out details in https://developers.stellar.org/docs/data/hubble/developer-guide/backfill/data-import

### What would you like to see?
Process documented to inital backfill for hubble dataset using data export / galexie

PR #1396, opened 2025-03-11
Add doc to setup inital backfill for hubble

amishas157:
Closes https://github.com/stellar/stellar-docs/issues/1394

This PR adds documentation for initial backfill

### DIFF
--- a/docs/data/hubble/developer-guide/backfill/data-import.mdx
+++ b/docs/data/hubble/developer-guide/backfill/data-import.mdx
@@ -44,7 +44,7 @@ COPY source_table_name
 
 # Use Galexie as a Source
 
-Galexie is a tool for extracting, processing, and exporting Stellar ledger metadata to external storage, creating a data lake of pre-processed ledger metadata. This is an upstream data source to Hubble, useful when you have a custom Stellar-ETL pipeline.
+[Galexie](../../../galexie/README.mdx) is a tool for extracting, processing, and exporting Stellar ledger metadata to external storage, creating a data lake of pre-processed ledger metadata. This is an upstream data source to Hubble, useful when you have a custom Stellar-ETL pipeline.
 
 **Steps**
 
--- a/docs/data/hubble/developer-guide/backfill/data-import.mdx
+++ b/docs/data/hubble/developer-guide/backfill/data-import.mdx
@@ -3,6 +3,51 @@ title: "Data Import"
 sidebar_position: 0
 ---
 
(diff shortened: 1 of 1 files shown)


docs/data/hubble/developer-guide/backfill/data-import.mdx
```diff
@@ -3,6 +3,51 @@ title: "Data Import"
...
+```sql
+CREATE [ OR REPLACE ] TABLE [ IF NOT EXISTS ] table_name
+COPY source_table_name
+...
+[OPTIONS(table_option_list)]
+```
+
+# Use Galexie as a Source
+
+Galexie is a tool for extracting, processing, and exporting Stellar ledger metadata to external storage, creating a data lake of pre-processed ledger metadata. This is an upstream data source to Hubble, useful when you have a custom Stellar-ETL pipeline.
```
chowbao, line 47: nit: add link to galexie docs

docs/data/hubble/developer-guide/backfill/data-import.mdx
```diff
@@ -3,6 +3,51 @@ title: "Data Import"
 sidebar_position: 0
 ---
 
-This document outlines methods to import data from Bigquery / Galexie for inital setup purpose.
+This document outlines methods to perform inital backfill when setting up hubble.
 
-Work in Progress
+# Use SDF ETL Data as a Source
```
chowbao, line 8: Hmm do we have a preference for what people should use out of the three options?

chowbao (approved):
LGTM 
The comments I left are both basically nits

docs/data/hubble/developer-guide/backfill/data-import.mdx
sydneynotthecity: My preference would be they run and operate their own Galexie 😄 

If people in the ecosystem are using SDF hosted infra as the source of record, that *technically* isn't using the blockchain as the source of record. Any holes we have in our dataset will be propagated into theirs. I see options 1 and 2 as more of a "hack" if you don't have the ability to run Galexie

amishas157: I would agree with having own Galexie as preferable option. It stands helpful when developers are building custom etl-pipeline. I think we should not list preferences in this doc as it can be use case dependent

sydneynotthecity: I'm ok with that. I do think we could implicitly rank our preferences though. So maybe Galexie moves to option #1? But I'll leave that up to you.

PR #1411, opened 2025-03-12
Collapse API providers tables

mollykarcher:


docs/data/apis/api-providers.mdx
```diff
@@ -0,0 +1,42 @@
+---
+sidebar_position: 60
+title: Providers
+---
+
+:::info
+
+On August 1, 2024, the SDF truncated historical data on its Horizon instances to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
```
leighmcculloch, line 8: Suggest removing the mix of third party and first party statements, and keep it all focused on the topic rather than on us:
```suggestion
On August 1, 2024, the SDF truncated historical data on its Horizon instances to one year. This update allows optimizes the performance of the SDF hosted Horizon and ensures a streamlined experience for all users of the publicly available API. Consider third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
```

leighmcculloch (approved):
One suggestion inline, but regardless looks good to me.

Needs updating `nginx/includes/redirects.conf` before merge so that the old URLs redirect to the new.

docs/data/apis/api-providers.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+### Publicly Accessible APIs
+
+| Provider | Futurenet RPC | Testnet RPC | Mainnet RPC | Futurenet Horizon | Testnet Horizon | Mainnet Horizon |
+| --- | --- | --- | --- | --- | --- | --- |
+| [Liquify](https://www.liquify.io/) | ✅ [Link](https://stellar.liquify.com/api=41EEWAH79Y5OCGI7/futurenet) | ✅ [Link](https://stellar.liquify.com/api=41EEWAH79Y5OCGI7/testnet) | ✅ [Link](https://stellar-mainnet.liquify.com/api=41EEWAH79Y5OCGI7/mainnet) | ❌ | ❌ | ❌ |
+| [Gateway](https://gateway.fm/) | ❌ | ✅ [Link](https://soroban-rpc.testnet.stellar.gateway.fm) | ✅ [Link](https://soroban-rpc.mainnet.stellar.gateway.fm) | ❌ | ❌ | ❌ |
+| [sorobanrpc.com](https://sorobanrpc.com/) | ❌ | ❌ | ✅ [Link](https://mainnet.sorobanrpc.com) | ❌ | ❌ | ❌ |
+| [Nodies](https://nodies.org) | ❌ | ✅ [Link](https://stellar-soroban-testnet-public.nodies.app) | ✅ [Link](https://stellar-soroban-public.nodies.app) | ❌ | ❌ | ❌ |
+| [SDF](http://www.stellar.org) | ✅ [Link](https://rpc-futurenet.stellar.org) | ✅ [Link](https://soroban-testnet.stellar.org) | ❌ | ✅ [Link](https://horizon-futurenet.stellar.org) | ✅ [Link](https://horizon-testnet.stellar.org) | ❌ |
```
chowbao, line 41: nit: Is there a way to make the links look nicer?
image

nit: Is there a way to make these two headers for the two tables the same?
image

docs/data/apis/api-providers.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+:::info
+
+On August 1, 2024, the SDF truncated historical data on its Horizon instances to one year. This update allows us to optimize performance and ensure a streamlined experience for all users. We encourage you to explore third-party ecosystem providers of Horizon, which may provide a longer history retention window as well as other features.
+
+:::
+
+Multiple infrastructure providers have made Stellar RPC and Horizon services available, and offer plans ranging from free to dedicated instances. These providers can be used for development, testing, and production.
+
+These providers allow access to the Testnet network and Mainnet network.
```
chowbao, line 14:
```suggestion
These providers allow access to the Futurenet network, Testnet network, and Mainnet network.
```

docs/data/apis/api-providers.mdx
mollykarcher: > nit: Is there a way to make the links look nicer?
 
I agree this looks kinda ugly but I wasn't sure if other options were better. I can make the emoji's themselves links, but in that case it's not really clear that they _are_ links. 

> nit: Is there a way to make these two headers for the two tables the same?

Unfortunately not, but the other alternative I tried was just Futurenet/Testnet/Mainnet columns and then named links inside the cells. This futzes around with the widths of the cells and makes them look more uneven. In short, it's also ugly but in a different way. I'm open to other ideas

mollykarcher: This is maybe the best better option to achieve both

image

PR #1417, opened 2025-03-13
Add URLs back to api providers table

mollykarcher:
We want the actual text of the URLs to appear in this table so that people can copy/paste them easily (ex. into Stellar Lab). Thread [here](https://stellarfoundation.slack.com/archives/C076K52SECF/p1741875960131389).

Just straight replacing the links with URLs makes the table wide and adds a scroll bar such that rightmost columns get hidden, like so:

image

Instead I've reformatted the table slightly to show the same info but alleviate that issue:

image

I'm open to other ideas opinions on what would be better.

### DIFF
--- a/docs/data/apis/api-providers.mdx
+++ b/docs/data/apis/api-providers.mdx
@@ -32,11 +32,16 @@ These providers allow access to the Futurenet, Testnet and Mainnet network.
 
 ### Publicly Accessible APIs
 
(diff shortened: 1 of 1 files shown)


chowbao (approved):

ElliotFriend:
that seems like a really good solution here!

janewang (approved):

PR #1439, opened 2025-03-20
cli: use artifacts instead of source code

Ifropc:
Previously we saw docs unexpectedly break with new cli release. 
This is first step to mitigate this: we now using release artifacts instead of source code to get the cookbook and full doc help.
Even if docs are broken, we don't need to re-release, we can simply re-upload artifacts.

### DIFF
--- a/scripts/stellar_cli.mjs
+++ b/scripts/stellar_cli.mjs
@@ -4,12 +4,18 @@ import { execSync } from "child_process";
 
 const repoUrl = "https://github.com/stellar/stellar-cli.git";
 const localRepoPath = "./stellar-cli-repo";
+const localArtifactPath = "./stellar-cli-artifacts"
 
 // Remove the existing repo if it exists
 if (fs.existsSync(localRepoPath)) {
   console.log("Removing existing repository...");
   fs.removeSync(localRepoPath);
 }
+if (fs.existsSync(localArtifactPath)) {
+  console.log("Removing existing artifacts...");
+  fs.removeSync(localArtifactPath);
+}
+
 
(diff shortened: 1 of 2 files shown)


scripts/stellar_cli.mjs
```diff
@@ -4,12 +4,18 @@ import { execSync } from "child_process";
 
 const repoUrl = "https://github.com/stellar/stellar-cli.git";
 const localRepoPath = "./stellar-cli-repo";
```
fnando, line 6:
```suggestion
```

scripts/stellar_cli.mjs
```diff
@@ -4,12 +4,18 @@ import { execSync } from "child_process";
...
+const localArtifactPath = "./stellar-cli-artifacts"
 
 // Remove the existing repo if it exists
```
fnando, line 9: We should remove this block as well.

fnando (changes requested):

scripts/stellar_cli.mjs
Ifropc: We are still getting latest version using git...

fnando: we can hit github's api instead, which I _think_ makes more sense if we're not using any of the repo.

```console
$ http https://api.github.com/repos/stellar/stellar-cli/tags | jq --raw-output '.[0].name'
v22.6.0
```

Ifropc: Yeah we can do that instead 👍🏼

Ifropc:
We decided not to to with this approach

PR #1446, opened 2025-03-25
change simulation state changes type to string

### LINKED ISSUE #1399: StateChanges Type enum in OpenRPC JSON specifies integer but server returns string
The spec here https://github.com/stellar/stellar-docs/blob/main/static/stellar-rpc.openrpc.json#L1942 defines the enum members as integers, unlike other enums such as these: https://github.com/stellar/stellar-docs/blob/9d3b63f80953dd101e268b4b393261b3a60ba9fa/static/stellar-rpc.openrpc.json#L163

An SDK user (Tarkus) reported this issue https://discordapp.com/channels/897514728459468821/1339636842026500239/1348704819098943548 where the RPC server is returning the string value Created instead of the corresponding integer. There is a mismatch between the spec and the RPC return value. Because the other enums in the spec are all string types, as far as I looked, then I am logging this as an anomaly in the spec. 

Other SDKs manually define these entities or just treat enums as strings, so others may not have noticed this discrepancy at execution time during serialisation, but the SDK under development in this case is automatically updated based on the specs, to make sure the SDK remains permanently useful. 

### What version are you using?

Main branch version

### What did you do?

SDK user is getting an enum value back from the server that cannot be parsed. Code generated RPC client based on RPC spec does not match returned values. Please see discord thread here: https://discordapp.com/channels/897514728459468821/1339636842026500239/1348704819098943548

RPC spec should match Stellar RPC response

### What did you see instead?

StateChangeType in SimulateTransactionResult is coming back as string type enum but is specified in rpc spec as integer type enum.

ElliotFriend:
This modifies the `simulateTransaction` response to include the `stateChanges[].type` as a string, like it is returned from the RPC server.

It appears [here](https://github.com/stellar/stellar-rpc/blob/49af4430cbcef8bb9995e4b266a36c619da7650f/protocol/simulate_transaction.go#L47C12-L47C12) in the `soroban-rpc` repo, that `LedgerEntryChangeType` is indeed an integer, but then there are some fancy-looking Golang things that I don't fully understand, but seems like it's what's responsible for it being returned as a string. @Shaptic, can you confirm that it's not a bug in the RPC server, but rather the spec, and that the server _is indeed_ designed to return a string for the state change type?

Fixes #1399

### DIFF
--- a/openrpc/src/stellar-rpc/methods/simulateTransaction.json
+++ b/openrpc/src/stellar-rpc/methods/simulateTransaction.json
@@ -85,8 +85,12 @@
                         "type": "object",
                         "properties": {
                             "type": {
-                                "type": "integer",
-                                "enum": [1, 2, 3],
+                                "type": "string",
+                                "enum": [
+                                    "created",
+                                    "updated",
+                                    "deleted"
+                                ],
                                 "description": "Indicates if the entry was created (1), updated (2), or deleted (3)"
                             },
                             "key": {
--- a/openrpc/src/stellar-rpc/methods/simulateTransaction.json
+++ b/openrpc/src/stellar-rpc/methods/simulateTransaction.json
@@ -91,7 +91,7 @@
                                     "updated",
                                     "deleted"
                                 ],
-                                "description": "Indicates if the entry was created (1), updated (2), or deleted (3)"
+                                "description": "Indicates if the entry was created, updated, or deleted"
                             },
                             "key": {
                                 "type": "string",
--- a/openrpc/src/anchor-platform/contentDescriptors/customer.json
+++ b/openrpc/src/anchor-platform/contentDescriptors/customer.json
@@ -2,7 +2,7 @@
   "customer_id": {
     "name": "customer_id",
     "description": "The SEP-12 ID of the customer",
-    "required": "false",
+    "required": false,
     "schema": {
       "type": "string"
     }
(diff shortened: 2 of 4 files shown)


Shaptic (approved):
@ElliotFriend your understanding is correct 
:clap:! [This line](https://github.com/stellar/stellar-rpc/blob/49af4430cbcef8bb9995e4b266a36c619da7650f/protocol/simulate_transaction.go#L97) is what defines the API schema, and you can see that `LedgerEntryChangeType` implements `MarshalJSON` by [encoding its string](https://github.com/stellar/stellar-rpc/blob/49af4430cbcef8bb9995e4b266a36c619da7650f/protocol/simulate_transaction.go#L55-L74) representation which is defined exactly as you have it in the schema. Thanks for looking into and fixing this!

I'll let you figure out the preview errors :laughing:

openrpc/src/stellar-rpc/methods/simulateTransaction.json
```diff
@@ -85,8 +85,12 @@
...
                             "type": {
-                                "type": "integer",
-                                "enum": [1, 2, 3],
+                                "type": "string",
+                                "enum": [
+                                    "created",
+                                    "updated",
+                                    "deleted"
+                                ],
                                 "description": "Indicates if the entry was created (1), updated (2), or deleted (3)"
```
Shaptic, line 94: You might wanna drop the integers from this, though!

ElliotFriend: good catch!

PR #1450, opened 2025-03-27
Check and warn on routes removal

leighmcculloch:
### What
Add a routes.txt file committed that stores a list of all routes.

Add a workflow that ensures the file stays up to date.

Add another workflow that checks that when the routes.txt is changing if there are any removed routes. If there are removed routes, it drops a reminder on the PR in a comment to make sure any redirects are added.

### Why
It's easy to miss adding redirects. We've all done it. The impact is real though. Folks search, get directed to a 404 page, and that experience isn't great.

### Example

When moving a route, the main build will fail with the following error prompting a developer to update the routes.txt by running `yarn build`:
Screenshot 2025-03-28 at 12 20 37 pm

If a developer runs `yarn build` locally and pushes the change, on the PR this diff will be obvious that routes are changing:
Screenshot 2025-03-28 at 12 23 32 pm

Any PR that modifies the routes and removes a route will get a reminder that will be posted to the PR to add redirects:
Screenshot 2025-03-28 at 12 31 11 pm


ElliotFriend (approved):
super helpful!!! Thank you a million times over!!!

PR #1460, opened 2025-04-01
Add doc validation workflow

Ifropc:
Adds github workflow for other repos to use for docs validation. (+ slightly re-organize existing main workflow)


.github/workflows/docs-validate.yml
leighmcculloch, line 1: The PR introduces a new docs-validate workflow, while also having a validate job in the main.yml. That's somewhat confusing.

Is refactoring the docs workflows necessary? I think we could run the existing docs workflow with just the new stellar-cli-ref input. Wdyt?

Ifropc: Yes we can do that. I didn't want to run extra stuff on cli side (especially if later main workflow will be extended for any reason), only the build.

leighmcculloch: Hmm, good point. I think the confusion wold be address by renaming the job / workflow.

.github/workflows/docs-validate.yml
```diff
@@ -0,0 +1,39 @@
...
+  workflow_call:
+    inputs:
+      stellar-cli-ref:
+        required: false
+        default: ''
+        type: string
+        description: "Stellar-cli ref to build against. If empty, will use latest cli version"
+
+jobs:
+  validate-docs:
```
leighmcculloch, line 13: This job does a build, could it be named `build`, and same with the file?

.github/workflows/main.yml
```diff
@@ -8,7 +8,15 @@ on:
...
-  build:
+  complete:
+    if: always()
+    needs: [ validate, build ]
+    runs-on: ubuntu-latest
+    steps:
+      - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+        run: exit 1
+
+  validate:
```
leighmcculloch, line 19: This job now just runs mdx checks. Can it be renamed `mdx-format`?

.github/workflows/main.yml
```diff
@@ -32,15 +40,13 @@ jobs:
...
-
-      - name: Build app
-        run: yarn build
-
       - name: Check if routes.txt file needs updating
         run: |
           if ! git diff --exit-code -- routes.txt; then
             echo 'The `routes.txt` file has changed. Please update the routes.txt file locally by running `yarn build` and push that change.'
             exit 1
           fi
```
leighmcculloch, line 48: There's no point checking the routes.txt here because it's not updated when not running build.

.github/workflows/docs-validate.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: yarn --prefer-offline
+        shell: bash
+
+      - name: Copy Files From Stellar CLI
+        run: yarn stellar-cli:build -- --cli-ref='${{ inputs.cli-ref }}'
+        shell: bash
+
+      - name: Build app
+        run: yarn build
+        shell: bash
```
leighmcculloch, line 37: The routes.txt check needs to happen after build.

.github/workflows/docs-validate.yml
leighmcculloch: I dropped some comments with suggestions to make it happen, but I'm still not sure this separation is meaningful. It'd be fine to keep it as one workflow imo.

Ifropc: I think it makes sense to have 1 workflow for build and a separate job for formatting check (because we don't need to run it in external repos), but I don't mind unifying it if you think it makes more sense

leighmcculloch (approved):
Looks good. Lets get an okay from @ElliotFriend too.

ElliotFriend (approved):
awesome!! thanks for taking care of this!!

PR #1501, opened 2025-04-22
Custom Token -> Contract Token

briwylde08:


janewang:
@briwylde08 I think we should add a major notice of this change in terminology at the top of the relevant pages. 

We should probably also add `custom tokens` in the glossary and indicate the term has changed to `contract tokens`

ElliotFriend (approved):
looks great to me!

PR #1520, opened 2025-04-29
[stellar-cli] Fix deprecated command usage.

fnando:


ElliotFriend (approved):
great catch! thanks for the fix

PR #1532, opened 2025-05-06
Update address conversion doc.

dmkozh:


docs/build/guides/conversions/address-conversions.mdx
```diff
@@ -59,44 +44,47 @@ use soroban_sdk::{Address, String, Env};
...
 // Example Stellar address
 const stellarAddress =
   "GCM5WPR4DDR24FSAX5LIEM4J7AI3KOWJYANSXEPKYXCSZOTAYXE75AFN";
-// Create an Address object
+// Create an Address object from string
 const address = new StellarSdk.Address(stellarAddress);
 // Convert the Address to xdr.ScVal
 const scVal = address.toScVal();
+// Convert scVal structure to the binary format
+const scValBuffer = scVal.toBuffer();
```
Shaptic, line 87:
```suggestion
const scValBuffer = scVal.toXDR("raw");
```

docs/build/guides/conversions/address-conversions.mdx
```diff
@@ -113,3 +101,28 @@ sc_address_xdr = address.to_xdr_sc_address()
 ```
 
 </CodeExample>
+
+## Conversions in Smart Contracts
+
+Smart contracts don't need to explicitly interact with the XDR types, as all the smart contract data types are automatically converted to XDR by the smart contract runtime. Every contract type, including `Address`, can be serialized to XDR bytes. This conversion is useful, for example, for performing hashing in smart contracts. It is also possible to convert the serialized XDR bytes back to contract types, which can be useful in some narrow use cases, such as custom authentication schemes.
+
+Note, that XDR conversions are and advanced feature and are not necessary for most of the Stellar smart contracts.
```
Shaptic, line 109: I was under the impression that XDR serialization required `std` :thinking: is that not true?

docs/build/guides/conversions/address-conversions.mdx
```diff
@@ -6,50 +6,35 @@ description: Convert an address to other types
...
 
-## Address to bytesN
+# String conversions
 
-Bytes are a more compact and efficient way to store data in terms of storage optimization and data transmission. In situations where you need to store or transmit an address in a fixed-size, such as for cryptographic operations or data serialization, you need to convert the address to a bytesN format
+The default string format for `Address` is so called 'string key' (or 'strkey'), a short base64-encoded string. For the default externally owned accounts strkey always starts with `G` and for all the contracts it starts with `C`. The exact details on the format can be found [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md), but they are not necessary to understand this section.
```
Shaptic, line 13: `strkey`s don't involve base64

```suggestion
The default string format for `Address` is a so called 'string key' (or 'strkey'), defined fully in [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md). For the default externally owned accounts strkey always starts with `G` and for all the contracts it starts with `C`.
```

Shaptic (approved):
Nice catches, definitely some weirdly-written docs here originally.

docs/build/guides/conversions/address-conversions.mdx
dmkozh: Fixed.

docs/build/guides/conversions/address-conversions.mdx
dmkozh: Nope, we have host functions for that. TBH I'm not sure if this section is useful, I just left it for the sake of consistency (because address<->strkey conversions are unique and make sense as a separate section). But maybe it's worth removing it and document the XDR conversions generically elsewhere...

docs/build/guides/conversions/address-conversions.mdx
dmkozh: Yeah, it's base32 actually, but I guess it's not very relevant.

docs/build/guides/conversions/address-conversions.mdx
Shaptic: Gotcha. I do think it's useful, though - I don't think it's covered well elsewhere (could be wrong, in which case a cross reference is better) -- more is better either way.

dmkozh: Yeah, I'm not saying this information is not useful at all, I'm rather questioning whether this belongs to a separate (not address-focused section). But for now it's probably better than nothing.

PR #1537, opened 2025-05-07
fix: correct import path for hello_world contract client

kcoolio:
Fixed the import path to point to the actual location of the TypeScript file at packages/hello_world/src/index.ts, resolving the module not found error.

### DIFF
--- a/docs/build/smart-contracts/getting-started/hello-world-frontend.mdx
+++ b/docs/build/smart-contracts/getting-started/hello-world-frontend.mdx
@@ -99,7 +99,7 @@ The default Astro project consists of a page (`pages/index.astro`) and a welcome
 
 ```ts title="src/pages/index.astro"
 ---
-import * as Client from './packages/hello_world';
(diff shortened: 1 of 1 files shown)


ElliotFriend:
@kcoolio Thanks for the contribution!

I gave the tutorial another run this morning, and for me it worked as advertised in the docs. Maybe you forgot to run the `npm run build` command in the `packages/hello_world` directory?

We don't _actually_ want to be importing the hello world client from the typescript file. The exported functionality is in `packages/hello_world/dist/index.js`, but that file only shows up after the build script is run.

@carstenjacobsen did you come across anything similar to this while you were writing the tutorial?

kcoolio:
That is right, I may have missed running the ´npm run build'in the packages.hello_world. Followed the steps again and it worked. Thanks a lot for the quick support.

ElliotFriend:
@kcoolio Awesome! Thanks for giving it a second look, and double-thanks for getting back to us!!

PR #1538, opened 2025-05-07
[stellar-cli] Add note about subcommand plugins.

fnando:


docs/tools/cli/plugins.mdx
```diff
@@ -75,6 +75,8 @@ If you're using GitHub to host your plugin's repository, consider adding a `stel
 
 :::
 
+When you run a command and no built-in command is found, we then fallback to searching for a plugin. This happens even for subcommands. For instance, let's say you're creating a new plugin to generate Ruby bindings for your contract; you can have something like `stellar-contract-bindings-ruby`, and users could execute `stellar contract bindings ruby` rather than the dasherized version.
```
leighmcculloch, line 78:
```suggestion
When you run a command and no built-in command is found, the Stellar CLI falls back to searching for a plugin. This happens even for subcommands. For instance, let's say you're creating a new plugin to generate Ruby bindings for your contract; you can have something like `stellar-contract-bindings-ruby`, and users could execute `stellar contract bindings ruby` rather than the dasherized version.
```

leighmcculloch (approved):

docs/tools/cli/plugins.mdx
leighmcculloch: Do plugins show up in the --help? E.g. `stellar contract bindings --help`. If they do, that's one of the big sells for this because plugins exist in context with other plugins of that type and that are relevant.

fnando: Not currently, but maybe we can extend all `--help` to include plugin lookups.

Issue #1499, opened 2025-04-21
Add SAC errors to documentation


briwylde08:
Add the SAC errors to this page:  https://developers.stellar.org/docs/tokens/stellar-asset-contract

Discord request: https://discord.com/channels/897514728459468821/1331985629290168351/1363987511881367604

JFWooten4:
This was pretty cool. Was there an internal process followed here?

re 1044 💜

briwylde08:
I don't think there was anything too special here! But... it did follow a pretty nice flow.

-Request comes in through the docs channel in Discord
-SDF creates an issue
-SDF assigns issue
-PR created and discussions/questions had in GH instead of internally for visibility

One thing I could have done is respond in Discord again with the completed PR so we can put a bow on it! @JFWooten4

JFWooten4:
When assigning the issue, was it prompted to Elliot because he's sort of the only person who's (publicly) worked on these sorts of docs through direct commits.[^1] More to the point, there presumably exists an internal communication system for SDF teams. If we used such a normal communication system; then what exchange occurred in the process of picking Elliot, and could we expand it to a community member?

[^1]: _See, e.g.,_ past work done by coauthors like Tyler which get bulkily wrapped into a PR by Elliot or yourself in the past. I also wonder to what extent code may be internally written and then communicated through a well-known public account like Garand's. Everyone fires off such incredible work on all cylinders, that some public commits seem like they could come from an unattributed group.

PR #1559, opened 2025-05-28
add contract errors for SACs

ElliotFriend:
I've copied the contract errors from the [built-in contracts crate](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/builtin_contracts/contract_error.rs).

Questions for specific people:

@briwylde08 Do the notes make sense? I'm sure there are ways they could be reworded or cleaned up, especially the ones that are like "never used i think" lol

@dmkozh I searched in that repo for occurrences of `ContractError::ErrorNameHere` and tried my best to make sense of what I was seeing. Did I capture correctly what each one is doing? What's the best way to handle the "never used" errors?

refs: #1499

### DIFF
--- a/docs/tokens/stellar-asset-contract.mdx
+++ b/docs/tokens/stellar-asset-contract.mdx
@@ -109,6 +109,57 @@ Priviliged mutators require authorization from a specific privileged identity, k
 
 This interface can be found in the [SDK]. It extends the common [SEP-41 Token Interface].
 
+## Contract Errors
+
+All built-in smart contracts on the Stellar network share the same error types, outlined below.
+
+```rust
+#[derive(Debug, FromPrimitive, PartialEq, Eq)]
+pub(crate) enum ContractError {
+    // Indicates something has gone wrong with an account's trustline/balance,
+    // or some unexpected input.
+    InternalError = 1,
+    // Indicates an impossible function has been invoked, such as clawback for
+    // an asset that does not have clawback enabled. Or, an operation has been
+    // called affecting the issuer's trustline.
+    OperationNotSupportedError = 2,
+    // Indicates the SAC has already been initialized.
+    AlreadyInitializedError = 3,
+
+    // Unused, from what I can tell.
+    UnauthorizedError = 4,
+    // Indicates a problem with a transaction's signature(s), such as too many
+    // signatures, too low signature weight(s), improper signature order, etc.
+    AuthenticationError = 5,
+    // An account which would be modified by this transaction does not exist on
+    // the network.
+    AccountMissingError = 6,
+    // Unused, from what I can tell.
+    AccountIsNotClassic = 7,
+
+    // Indicates an amount less-than zero was provided for a transfer amount.
+    NegativeAmountError = 8,
+    // Indicates an insufficient spender's available allowance amount. Also used
+    // to indicate a problem with expiration ledger when creating an allowance.
+    AllowanceError = 9,
+    // Indicates too low of a balance to spend the requested amount, or a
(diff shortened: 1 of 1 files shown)


docs/tokens/stellar-asset-contract.mdx
```diff
@@ -109,6 +109,55 @@ Priviliged mutators require authorization from a specific privileged identity, k
...
+## Contract Errors
+
+All built-in smart contracts on the Stellar network share the same error types, outlined below.
+
+```rust
+#[derive(Debug, FromPrimitive, PartialEq, Eq)]
+pub(crate) enum ContractError {
+    // Indicates something has gone wrong with an account's trustline/balance,
+    // or some unexpected input.
+    InternalError = 1,
```
dmkozh, line 122: This error is an unfortunate artifact of the early Soroban development and it shouldn't be a contract error at all, but an internal host error. But since it exists, the proper documentation would be 'Indicates an internal error in protocol implementation, such as invalid ledger state. This may not happen in the real networks, but might appear when using malformed test data (such as malformed ledger snapshots).'.

docs/tokens/stellar-asset-contract.mdx
```diff
@@ -109,6 +109,55 @@ Priviliged mutators require authorization from a specific privileged identity, k
...
+    // Indicates an impossible function has been invoked, such as clawback for
+    // an asset that does not have clawback enabled. Or, an operation has been
+    // called affecting the issuer's trustline.
+    OperationNotSupportedError = 2,
+    // Indicates the SAC has already been initialized.
```
dmkozh: It's worth adding that this error may only occur during the SAC instantiation.

docs/tokens/stellar-asset-contract.mdx
```diff
@@ -109,6 +109,55 @@ Priviliged mutators require authorization from a specific privileged identity, k
...
+    AlreadyInitializedError = 3,
+    // Appears to be unused.
+    UnauthorizedError = 4,
```
dmkozh: Yep, this is unused (unfortunately). The name is meaningless, so maybe it's worth commenting this line out, something like `// Unused = 4, - this error code is not used by SAC

docs/tokens/stellar-asset-contract.mdx
```diff
@@ -109,6 +109,55 @@ Priviliged mutators require authorization from a specific privileged identity, k
...
+    // Indicates a problem with a transaction's signature(s), such as too many
+    // signatures, too low signature weight(s), improper signature order, etc.
+    AuthenticationError = 5,
```
dmkozh: This is not a SAC error (sorry, this is really embarrassing), so we should give it the same treatment as code 4

docs/tokens/stellar-asset-contract.mdx
```diff
@@ -109,6 +109,55 @@ Priviliged mutators require authorization from a specific privileged identity, k
...
+    // An account that would be modified by this transaction does not exist on
+    // the network.
+    AccountMissingError = 6,
+    // Appears to be unused.
+    AccountIsNotClassic = 7,
```
dmkozh: Yep, this is not used, so we should give it the same treatment as to the other unused codes.

dmkozh (commented):
Thanks, this uncovers a bunch of really embarrassing issues with this enum (the worst one is that we're reusing this enum for the account contract errors). I don't think we can do much about that besides explicitly indicating the codes unused by SAC.

ElliotFriend:
Thanks so much, @dmkozh! I've pushed those changes now 🙏🏻

dmkozh (approved):

PR #1560, opened 2025-05-29
Fix RPC Provider Redirect

briwylde08:


ElliotFriend (approved):
Great catch!!! 🎉 

Thanks for the fix!

PR #1563, opened 2025-05-30
adjust structure of payment guide to include RPC and SAC details

fazzatti:
Update the Send and receive payments guide with the following adjustments:
- Modify the page structure and description to include mentions about the option of using RPC, Custom tokens and the Stellar Asset Contract.
- Include snippets and detailed examples to:
  - Send classic payments through RPC
  - Send SAC payments
  - Receive SAC payments

Not included: 
- Receive a classic payment through RPC: This should be added in the future once Classic payments emit soroban events.

### DIFF
--- a/docs/build/guides/basics/send-and-receive-payments.mdx
+++ b/docs/build/guides/basics/send-and-receive-payments.mdx
@@ -71,84 +71,68 @@ try {
   throw error;
 }
 
-// The next step is to parametrize and build the transaction object
-let transaction: StellarSdk.Transaction;
-try {
-  // Using the source account we just loaded we begin to assemble the transaction.
-  // We set the fee to the base fee, which is 100 stroops (0.00001 XLM).
-  // We also set the network passphrase to TESTNET.
-  transaction = new StellarSdk.TransactionBuilder(sourceAccount, {
-    fee: StellarSdk.BASE_FEE,
-    networkPassphrase: StellarSdk.Networks.TESTNET,
-  })
-    // We then add a payment operation to the transaction oject.
-    // This operation will send 10 XLM to the destination account.
-    // Obs.: Not specifying a explicit source account here means that the
-    // operation will use the source account of the whole transaction, which we specified above.
-    .addOperation(
-      StellarSdk.Operation.payment({
-        destination: destinationId,
-        asset: StellarSdk.Asset.native(),
-        amount: "10",
-      })
-    )
-    // We include an optional memo which oftentimes is used to identify the transaction
-    // when working with pooled accounts or to facilitate reconciliation.
-    .addMemo(StellarSdk.Memo.id("1234567890"))
-    // Finally, we set a timeout for the transaction.
-    // This means that the transaction will not be valid anymore after 180 seconds.
-    .setTimeout(180)
-    .build();
-} catch (error) {
-  console.error("Error building transaction:", error);
-  throw error;
+// The next step is to parametrize and build the transaction object:
+// Using the source account we just loaded we begin to assemble the transaction.
+// We set the fee to the base fee, which is 100 stroops (0.00001 XLM).
+// We also set the network passphrase to TESTNET.
+const transaction = new StellarSdk.TransactionBuilder(sourceAccount, {
+  fee: StellarSdk.BASE_FEE,
+  networkPassphrase: StellarSdk.Networks.TESTNET,
+})
+  // We then add a payment operation to the transaction oject.
+  // This operation will send 10 XLM to the destination account.
+  // Obs.: Not specifying a explicit source account here means that the
+  // operation will use the source account of the whole transaction, which we specified above.
+  .addOperation(
+    StellarSdk.Operation.payment({
+      destination: destinationId,
+      asset: StellarSdk.Asset.native(),
+      amount: "10",
+    })
+  )
+  // We include an optional memo which oftentimes is used to identify the transaction
+  // when working with pooled accounts or to facilitate reconciliation.
+  .addMemo(StellarSdk.Memo.id("1234567890"))
+  // Finally, we set a timeout for the transaction.
+  // This means that the transaction will not be valid anymore after 180 seconds.
+  .setTimeout(180)
+  .build();
+
+
+// We sign the transaction with the source account's secret key.
+transaction.sign(sourceKeys);
+
+// Now we can send the transaction to the network.
+// The sendTransaction method immediately returns a reply with the transaction hash
+// and the status "PENDING". This means the transaction was received and is being processed.
+const sendTransactionResponse = await rpc.sendTransaction(transaction);
+
+// Here we check the status of the transaction as there are
+// a possible outcomes after sending a transaction that would have
+// to be handled accordingly, such as "DUPLICATE" or "TRY_AGAIN_LATER".
+if (sendTransactionResponse.status !== "PENDING") {
+  throw new Error(
+    `Failed to send transaction, status: ${sendTransactionResponse.status}`
+  );
 }
 
-try {
-  // We sign the transaction with the source account's secret key.
-  transaction.sign(sourceKeys);
-} catch (error) {
-  console.error("Error signing transaction:", error);
-  throw error;
-}
+// Here we poll the transaction status to await for its final result.
+// We can use the transaction hash to poll the transaction status later.
+const finalStatus = await rpc.pollTransaction(sendTransactionResponse.hash, {
+  sleepStrategy: (_iter: number) => 500,
+  attempts: 5,
+});
 
-let sendTransactionResponse: StellarSdk.rpc.Api.SendTransactionResponse;
-try {
-  // Now we can send the transaction to the network.
-  // The sendTransaction method immediately returns a reply with the transaction hash
-  // and the status "PENDING". This means the transaction was received and is being processed.
-  sendTransactionResponse = await rpc.sendTransaction(transaction);
-
-  if (sendTransactionResponse.status !== "PENDING") {
-    throw sendTransactionResponse;
-  }
-} catch (error) {
-  console.error("Error sending transaction:", error);
-  throw error;
-}
-
-try {
-  // Here we poll the transaction status to await for its final result.
-  // We can use the transaction hash to poll the transaction status later.
-  const finalStatus = await rpc.pollTransaction(sendTransactionResponse.hash, {
-    sleepStrategy: (_iter: number) => 500,
-    attempts: 5,
-  });
-
-  // The pollTransaction method will return the final status of the transaction
-  // after the specified number of attempts or when the transaction is finalized.
-  // We then check the final status of the transaction and handle it accordingly.
-  switch (finalStatus.status) {
-    case StellarSdk.rpc.Api.GetTransactionStatus.FAILED:
-    case StellarSdk.rpc.Api.GetTransactionStatus.NOT_FOUND:
-      throw new Error(`Transaction failed with status: ${finalStatus.status}`);
-    case StellarSdk.rpc.Api.GetTransactionStatus.SUCCESS:
-      console.log("Success! Results:", finalStatus);
-      break;
-  }
-} catch (error) {
-  console.error("Error polling transaction status:", error);
-  throw error;
+// The pollTransaction method will return the final status of the transaction
+// after the specified number of attempts or when the transaction is finalized.
+// We then check the final status of the transaction and handle it accordingly.
+switch (finalStatus.status) {
+  case StellarSdk.rpc.Api.GetTransactionStatus.FAILED:
+  case StellarSdk.rpc.Api.GetTransactionStatus.NOT_FOUND:
+    throw new Error(`Transaction failed with status: ${finalStatus.status}`);
+  case StellarSdk.rpc.Api.GetTransactionStatus.SUCCESS:
+    console.log("Success! Results:", finalStatus);
+    break;
 }
 
 ```
@@ -201,46 +185,36 @@ try {
   throw error;
 }
 
-// The next step is to parametrize and build the transaction object
-let transaction: StellarSdk.Transaction;
-try {
-  // Using the source account we just loaded we begin to assemble the transaction.
-  // We set the fee to the base fee, which is 100 stroops (0.00001 XLM).
-  // We also set the network passphrase to TESTNET.
-  transaction = new StellarSdk.TransactionBuilder(sourceAccount, {
+// The next step is to parametrize and build the transaction object:
+// Using the source account we just loaded we begin to assemble the transaction.
+// We set the fee to the base fee, which is 100 stroops (0.00001 XLM).
+// We also set the network passphrase to TESTNET.
+const transaction = new StellarSdk.TransactionBuilder(sourceAccount, {
     fee: StellarSdk.BASE_FEE,
     networkPassphrase: StellarSdk.Networks.TESTNET,
   })
-    // We then add a payment operation to the transaction oject.
-    // This operation will send 10 XLM to the destination account.
-    // Obs.: Not specifying a explicit source account here means that the
-    // operation will use the source account of the whole transaction, which we specified above.
-    .addOperation(
-      StellarSdk.Operation.payment({
-        destination: destinationId,
-        asset: StellarSdk.Asset.native(),
-        amount: "10",
-      })
-    )
-    // We include an optional memo which oftentimes is used to identify the transaction
-    // when working with pooled accounts or to facilitate reconciliation.
-    .addMemo(StellarSdk.Memo.id("1234567890"))
-    // Finally, we set a timeout for the transaction.
-    // This means that the transaction will not be valid anymore after 180 seconds.
-    .setTimeout(180)
-    .build();
-} catch (error) {
-  console.error("Error building transaction:", error);
-  throw error;
-}
+  // We then add a payment operation to the transaction oject.
+  // This operation will send 10 XLM to the destination account.
+  // Obs.: Not specifying a explicit source account here means that the
+  // operation will use the source account of the whole transaction, which we specified above.
+  .addOperation(
+    StellarSdk.Operation.payment({
+      destination: destinationId,
+      asset: StellarSdk.Asset.native(),
+      amount: "10",
+    })
+  )
+  // We include an optional memo which oftentimes is used to identify the transaction
+  // when working with pooled accounts or to facilitate reconciliation.
+  .addMemo(StellarSdk.Memo.id("1234567890"))
+  // Finally, we set a timeout for the transaction.
+  // This means that the transaction will not be valid anymore after 180 seconds.
+  .setTimeout(180)
+  .build();
 
-try {
-  // We sign the transaction with the source account's secret key.
-  transaction.sign(sourceKeys);
-} catch (error) {
-  console.error("Error signing transaction:", error);
-  throw error;
-}
+
+// We sign the transaction with the source account's secret key.
+transaction.sign(sourceKeys);
 
 // Now we can send the transaction to the network.
 // The sendTransaction method returns a promise that resolves with the transaction result.


docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
 
 Every transaction also incurs a small fee. Like the minimum balance on accounts, this fee deters spam and prevents people from overloading the system. This [base fee](../../../learn/fundamentals/fees-resource-limits-metering.mdx) is very small — 100 stroops per operation where a stroop equals 1 \* 10 ^-7 XLM — and it's charged for each operation in a transaction. A transaction with two operations, for instance, would cost 200 stroops.
```
JakeUrban: I'm not sure why we have this section about operations and transactions here. We have a more in-depth page on this in the learn section: https://developers.stellar.org/docs/learn/fundamentals/transactions

Maybe we should remove this paragraph and the text above it, and reference the learn section similar to how we reference the "create an account" guide?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
...
 
+With the recent addition of Soroban and smart contracts, Stellar also supports a different way of executing transactions. These use a specific type of operation (`invokeHostFunction`) to invoke a smart contract function, and this contract carries the logic that executes custom functionality and/or payments. By default, all classic assets on Stellar can be represented by a default smart contract called a [Stellar Asset Contract (SAC)](../../../tokens/stellar-asset-contract.mdx), which provides a standardized interface for asset operations through contract logic.
```
JakeUrban: If we remove the background information on transactions and operations, this section also seems a bit out of place.

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
...
+
+For the purpose of this guide, the payment examples will be segregated between **Classic Payments** (using built-in operations like the traditional `payment` operation) and **SAC Payments** (which make use of the default smart contract and execute payments through contract logic).
+
+Additionally, there are two main options to interact with the Stellar network: the [Horizon API](../../../data/apis/horizon/README.mdx) and [Stellar RPC](../../../data/apis/rpc/README.mdx). Each payment example will be broken down to show how to perform such payments with either option.
+
```
JakeUrban: This is the core context I think readers need before diving into the examples, but I think we can expand on it.

```suggestion
There are two ways to make payments on Stellar. One approach is to use Stellar's payment-related operations, and the other is to invoke a function on the token's contract, and the approach you should take depends on the use case.

If you want to make a payment of a Stellar asset between two Stellar accounts, use the payment operations. Transaction fees are cheaper when using them compared to those invoking the Stellar asset contract (or SAC).

If you want to make a payment of a Stellar asset between a Stellar account and a contract address, or between two contract addresses, you must use the SAC. Stellar's payment-related operations cannot have contract addresses as their source or destination.

Finally, if you want to make a payment of a contract token (not a Stellar asset), you must use the token's contract. Stellar's payment operations can only be used to transfer Stellar assets.

To learn more about the differences between Stellar assets and contract tokens, see the [Tokens](https://developers.stellar.org/docs/tokens) overview.

Additionally, there are two main options to interact with the Stellar network: the [Horizon API](../../../data/apis/horizon/README.mdx) and [Stellar RPC](../../../data/apis/rpc/README.mdx). Generally speaking, you should use RPC unless you need features that only Horizon provides. If your use case is limited to making payments, RPC should be sufficient. Regardless, each payment example will be broken down to show how to perform such payments with either option.

```

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
...
 <Alert>
 
 In the following code samples, proper error checking is omitted for brevity. However, you should _always_ validate your results, as there are many ways that requests can fail. You should refer to the guide on [Error Handling](../../../data/apis/horizon/api-reference/errors/error-handling.mdx) for tips on error management strategies.
 
 </Alert>
 
-## Send a Payment
+## Classic Payments
```
JakeUrban: Lets try and find a different heading title that doesn't use "Classic". Maybe, "Using the Payment Operation"?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
...
+rpc
+  .getAccount(destinationId)
+  .catch(function (error) {
+    if (error instanceof StellarSdk.NotFoundError) {
+      throw new Error("The destination account does not exist!");
+    } else return error;
+  })
+  .then(function () {
+    return rpc.getAccount(sourceKeys.publicKey());
+  })
```
JakeUrban: I've noticed most of our JS examples use this syntax as opposed to async/await or try/catch. I think the latter is more readable, can we switch?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -17,15 +17,107 @@ Transactions are atomic, meaning that if any operation in a transaction fails, t
...
+      networkPassphrase: StellarSdk.Networks.TESTNET,
+    })
+      .addOperation(
+        StellarSdk.Operation.payment({
+          destination: destinationId,
+          asset: StellarSdk.Asset.native(),
+          amount: "10",
+        })
+      )
+      .addMemo(StellarSdk.Memo.text("Test Transaction"))
```
JakeUrban: Lets use an integer memo here, since most of the time users use integer memos when depositing to exchanges.

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -472,14 +595,20 @@ server.submit_transaction(transaction)
...
+That's why you should always save a built transaction (or transaction encoded in XDR format) in a variable or a database and resubmit it if you don't know its status. If the transaction has already been successfully applied to the ledger, both Horizon and RPC will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction's status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.
 
-## Receive a Payment
+### Receive a Payment
 
 You don’t actually need to do anything to receive payments into a Stellar account: if a payer makes a successful transaction to send assets to you, those assets will automatically be added to your account.
 
 However, you may want to keep an eye out for incoming payments. A simple program that watches the network for payments and prints each one might look like:
 
+{/* Using RPC will be introduced later once we have events launched for classic asset effects */}
```
JakeUrban: Can we add some text to explain this is coming with Protocol 23?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -795,8 +924,371 @@ payments_next = payments.next()
 
 </CodeExample>
 
+## Stellar Asset Contract (SAC) Payments
+
+When sending payments using Stellar Asset Contracts, you invoke smart contract functions instead of using classic payment operations. SAC payments utilize the `invokeHostFunction` operation to call contract methods like `transfer`, which provides the same payment functionality but through contract logic. This approach offers additional flexibility and programmability while maintaining compatibility with Stellar's asset system.
```
JakeUrban:
```suggestion
When sending payments using Stellar Asset Contracts, you invoke smart contract functions instead of using payment operations. SAC payments utilize the `invokeHostFunction` operation to call contract methods like `transfer`, which provides the same payment functionality but through contract logic. This approach offers additional flexibility and programmability while maintaining compatibility with Stellar's asset system.
```

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -795,8 +924,371 @@ payments_next = payments.next()
...
+
+In the example below, we're using SAC as a smart contract interface for the native asset, so the end result is the same as if a native payment operation had been executed. However, the same process can be applied for any smart contract transfer involving standardized custom tokens that follow the [custom token interface](../../../tokens/token-interface.mdx). This makes SAC payments a unified approach for handling both native assets and custom tokens through a consistent contract-based interface.
```
JakeUrban:
```suggestion
In the example below, we're using SAC as a smart contract interface for the native asset, so the end result is the same as if a native payment operation had been executed. However, the same process can be applied for any smart contract transfer involving standardized contract tokens that follow the [contract token interface](../../../tokens/token-interface.mdx). This makes SAC payments a unified approach for handling both native assets and contract tokens through a consistent contract-based interface.
```

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -795,8 +924,371 @@ payments_next = payments.next()
...
+  publicKey: sourceKeys.publicKey(),
+  // Custom transaction signing function
+  signTransaction: async (txXdr) => {
+    const tx = await TransactionBuilder.fromXDR(txXdr, Networks.TESTNET);
+    tx.sign(sourceKeys);
+    return {
+      signedTxXdr: tx.toXDR(),
+      signerAddress: sourceKeys.publicKey(),
+    };
+  },
```
JakeUrban: This seems like the simplest and most common approach to signing transactions, is there a default provided by the client we can leverage? If not, maybe we should file an issue on the JS SDK?

JakeUrban (commented):
The examples look great. I added some additional context and recommended removing other pieces of content better placed elsewhere in the docs. I also think we need to switch to the more imperative coding style for readability.

docs/build/guides/basics/send-and-receive-payments.mdx
fazzatti: I'm so glad to hear that! I agree 100%!
I'll make these adjustments right away!

docs/build/guides/basics/send-and-receive-payments.mdx
fazzatti: Great point! I also miss having something ready to use for such common cases.
 
I see two paths in my mind depending on what we should consider scope for the SDK versus the bindings.

1. Add a ready-to-use option to the JS SDK, but also with a bit more heavy lifting for the users such as handling common auth entries cases, etc. I did something similar for the Stellar Plus signing steps. You would simply provide a list of signers and the SDK would run a few steps:
- Verify auth entries and sign with the proper account if provided (throw if missing a required signer)
- Verify classic ops and envelope and sign with the proper account if provider (throw if missing a required signer)
- Adjusting the necessary when fee bumping a soroban transaction

These are common enough that we could cover most average user while, more advance use cases, would still be tackled with the custom approach.


2. As a quick approach, simply make the bindings client a bit more flexible to optionally use the provided account as simple signer instead of requiring the function.

I'll open the issue 👍

fazzatti: Ok looking further I just notice this function is part of the underlying client from JS-SDK indeed. I mistakenly thought this was generated by the bindings as an extended client.

So it seems like definitely something we could tackle on the SDK directly!

fazzatti: https://github.com/stellar/js-stellar-sdk/issues/1179

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,221 @@ description: Learn to send payments and watch for received payments on the Stell
 import { CodeExample } from "@site/src/components/CodeExample";
 import { Alert } from "@site/src/components/Alert";
 
-Most of the time, you’ll be sending money to someone else who has their own account. For this tutorial, however, you'll need a second account to transact with. So before proceeding, follow the steps outlined in [Create an Account](./create-account.mdx) to make _two_ accounts: one for sending and one for receiving.
+Most of the time, you’ll be sending money to someone else who has their own account. For this tutorial, however, you'll need a second account to transact with. So before proceeding, follow the steps outlined in [Create an Account](./create-account.mdx) to make _two_ accounts: one for sending and one for receiving. We'll also be using both native operations as well as Smart Contract invocations to perform payments. For more details about the different types of transactions, see the [Operations and Transactions](../../../learn/fundamentals/transactions) documentation.
```
JakeUrban:
```suggestion
Most of the time, you’ll be sending money to someone else who has their own account. For this tutorial, however, you'll need a second account to transact with. So before proceeding, follow the steps outlined in [Create an Account](./create-account.mdx) to make _two_ accounts: one for sending and one for receiving. We'll also be using both payment operations as well as smart contract invocations to perform payments. For more details about the different types of transactions, see the [Operations and Transactions](../../../learn/fundamentals/transactions) documentation.
```

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,221 @@ description: Learn to send payments and watch for received payments on the Stell
...
 
-Every transaction also incurs a small fee. Like the minimum balance on accounts, this fee deters spam and prevents people from overloading the system. This [base fee](../../../learn/fundamentals/fees-resource-limits-metering.mdx) is very small — 100 stroops per operation where a stroop equals 1 \* 10 ^-7 XLM — and it's charged for each operation in a transaction. A transaction with two operations, for instance, would cost 200 stroops.
+If you want to make a payment of a Stellar asset between a Stellar account and a contract address, or between two contract addresses, you must use the SAC. Stellar's payment-related operations cannot have contract addresses as their source or destination.
+
+Finally, if you want to make a payment of a contract token (not a Stellar asset), you must use the token's contract. Stellar's payment operations can only be used to transfer Stellar assets.
+
+To learn more about the differences between Stellar assets and contract tokens, see the [Tokens](https://developers.stellar.org/docs/tokens) overview.
```
JakeUrban: I recognize I wrote this, but can we replace the external link with the internal link?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,221 @@ description: Learn to send payments and watch for received payments on the Stell
...
+);
+const destinationId =
+  "GA2C5RFPE6GCKMY3US5PAB6UZLKIGSPIUKSLRB6Q723BM2OARMDUYEJ5";
+
+// First, check to make sure that the destination account exists.
+await rpc.getAccount(destinationId).catch((error) => {
+  if (error instanceof StellarSdk.NotFoundError) {
+    throw new Error("The destination account does not exist!");
+  } else throw error;
+});
```
JakeUrban: Can we use `try/catch` here for the `NotFoundError`? Same for the other instances of `.catch`

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,221 @@ description: Learn to send payments and watch for received payments on the Stell
...
+      switch (finalStatus.status) {
+        case StellarSdk.rpc.Api.GetTransactionStatus.FAILED:
+        case StellarSdk.rpc.Api.GetTransactionStatus.NOT_FOUND:
+          throw new Error(
+            `Transaction failed with status: ${finalStatus.status}`
+          );
+        case StellarSdk.rpc.Api.GetTransactionStatus.SUCCESS:
+          return finalStatus;
+      }
+    });
```
JakeUrban: Can we remove the `.then` and use `await`?

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,221 @@ description: Learn to send payments and watch for received payments on the Stell
...
-  })
-  .catch(function (error) {
-    console.error("Something went wrong!", error);
-    // If the result is unknown (no response body, timeout etc.) we simply resubmit
-    // already built transaction:
-    // server.submitTransaction(transaction);
+    } else throw error;
   });
+
+try {
```
JakeUrban, line 181: Putting large pieces of code inside a `try/catch` is bad practice, can we keep them scoped to specific function calls so we can check for the possible error types? Looking at the code block, its not clear to me where errors could be thrown.

JakeUrban (commented):
Looks like we're still using the promise-chain approach in some cases (`.then() or `.catch()`)

docs/build/guides/basics/send-and-receive-payments.mdx
ElliotFriend: While we're here, let's also make that last link go to `../../../learn/fundamentals/transactions/README.mdx`

fazzatti:
> Looks like we're still using the promise-chain approach in some cases (`.then()` or `.catch()`)

I refactored the snippets with this in mind. While I agree with standardizing the approach and making it very explicit, I think from a knowledge management perspective we're getting to a point in which we might be getting to a limit to how much we should add to a single article.

If we plan on expanding it more, we might begin considering either bringing these snippets to a repo as suggested examples and just referring to them in the article or, breakdown this article into shorter, atomic KBAs and linking them into a higher-level guide.

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,251 @@ description: Learn to send payments and watch for received payments on the Stell
...
+  sourceAccount = await rpc.getAccount(sourceKeys.publicKey());
+} catch (error) {
+  console.error("Error checking source account:", error);
+  throw error;
+}
+
+// The next step is to parametrize and build the transaction object
+let transaction: StellarSdk.Transaction;
+try {
```
JakeUrban: I think we can remove this try/catch, because all the errors would be unexpected rather than expected. For example, a 404 not found from Horizon when looking up the account is an expected error, hence the try/catch.

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,251 @@ description: Learn to send payments and watch for received payments on the Stell
...
+    // Finally, we set a timeout for the transaction.
+    // This means that the transaction will not be valid anymore after 180 seconds.
+    .setTimeout(180)
+    .build();
+} catch (error) {
+  console.error("Error building transaction:", error);
+  throw error;
+}
+
+try {
```
JakeUrban: Same here re: removal of try/catch

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,251 @@ description: Learn to send payments and watch for received payments on the Stell
...
+  // We sign the transaction with the source account's secret key.
+  transaction.sign(sourceKeys);
+} catch (error) {
+  console.error("Error signing transaction:", error);
+  throw error;
+}
+
+let sendTransactionResponse: StellarSdk.rpc.Api.SendTransactionResponse;
+try {
```
JakeUrban: Will `sendTransaction` throw an error if the status is `ERROR`? Or will it just return the JSON response normally? I don't think we need a `try/catch` unless it throws the error rather than returning it.

docs/build/guides/basics/send-and-receive-payments.mdx
```diff
@@ -7,90 +7,251 @@ description: Learn to send payments and watch for received payments on the Stell
...
+  }
+} catch (error) {
+  console.error("Error sending transaction:", error);
+  throw error;
+}
+
+try {
+  // Here we poll the transaction status to await for its final result.
+  // We can use the transaction hash to poll the transaction status later.
+  const finalStatus = await rpc.pollTransaction(sendTransactionResponse.hash, {
```
JakeUrban: same here re: removal of try/catch

JakeUrban (commented):
This is looking great. Just one more ask: lets remove unnecessary try/catchs. If we don't know the error that could be thrown, we shouldn't have one.

fazzatti:
> This is looking great. Just one more ask: lets remove unnecessary try/catchs. If we don't know the error that could be thrown, we shouldn't have one.

Much better now!! I misunderstood the other comment and though you wanted to wrap each individual step one by one.

PR #1595, opened 2025-07-03
stellar-rpc: Update getEvents schema

urvisavla:
Update the getEvents schema to reflect the following changes introduced in stellar-rpc 23.0:

- Support for the new "**" wildcard - https://github.com/stellar/stellar-rpc/pull/419

- Removal of diagnostic events - https://github.com/stellar/stellar-rpc/issues/402

### DIFF
--- a/openrpc/src/stellar-rpc/schemas/Event.json
+++ b/openrpc/src/stellar-rpc/schemas/Event.json
@@ -46,7 +46,7 @@
         "title": "type",
         "description": "The type of event emission.",
         "type": "string",
-        "enum": [ "contract", "diagnostic", "system" ]
+        "enum": [ "contract", "system" ]
     },
     "EventId": {
         "description": "Unique identifier for this event.\n\n- The event's unique id field is based on a [`toid` from Horizon](https://github.com/stellar/go/blob/master/toid/main.go) as used in Horizon's /effects endpoint.\n\n- https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/effect.go#L58\n\n- Specifically, it is a string containing:\n\n- bigint(32 bit ledger sequence + 20 bit txn number + 12 bit operation) + `<hyphen>` + number for the event within the operation.\n\n- For example: `1234-1`",
--- a/openrpc/src/stellar-rpc/schemas/EventFilters.json
+++ b/openrpc/src/stellar-rpc/schemas/EventFilters.json
@@ -24,7 +24,7 @@
     "EventFilterType": {
         "title": "type",
         "type": "string",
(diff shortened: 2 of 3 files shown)


openrpc/src/stellar-rpc/schemas/EventFilters.json
```diff
@@ -24,7 +24,7 @@
     "EventFilterType": {
         "title": "type",
         "type": "string",
-        "description": "A comma separated list of event types (system, contract, or diagnostic) used to filter events. If omitted, all event types are included."
+        "description": "A comma separated list of event types (system or contract) used to filter events. If omitted, all event types are included."
```
Shaptic: Wait I don't think you can do comma separation in this field... if you want `contract` and `system`, you need two separate filters. Can you amend this accordingly?

openrpc/src/stellar-rpc/schemas/Event.json
```diff
@@ -46,7 +46,7 @@
         "title": "type",
         "description": "The type of event emission.",
         "type": "string",
-        "enum": [ "contract", "diagnostic", "system" ]
+        "enum": [ "contract", "system" ]
     },
     "EventId": {
         "description": "Unique identifier for this event.\n\n- The event's unique id field is based on a [`toid` from Horizon](https://github.com/stellar/go/blob/master/toid/main.go) as used in Horizon's /effects endpoint.\n\n- https://github.com/stellar/go/blob/master/services/horizon/internal/db2/history/effect.go#L58\n\n- Specifically, it is a string containing:\n\n- bigint(32 bit ledger sequence + 20 bit txn number + 12 bit operation) + `<hyphen>` + number for the event within the operation.\n\n- For example: `1234-1`",
```
Shaptic: Can you expand here (since we're here) that the first part of the `EventId` is zero-padded (`"0"`) to 19 characters and the second part is zero-padded to 10 characters? (see [`protocol/cursor.go#L42`](https://github.com/stellar/stellar-rpc/blob/d855b99e76bbd8189ebb2011f0f8390f6ef6d9bf/protocol/cursor.go#L42))

Shaptic (approved):
Approving so you don't need a re-review but left some :memo: below :+1:

openrpc/src/stellar-rpc/schemas/Event.json
urvisavla: done

openrpc/src/stellar-rpc/schemas/EventFilters.json
urvisavla: done.

PR #1609, opened 2025-07-18
Add p23/testnet section to software versions

sreuland:
Started a new section for Protocol 23 versions for Testnet as it was just updated on 7/17 and people are likely looking for single place to reference the versions across the full stack.

This is starting the block and have filled in a few known versions, but will need input from some app owners to complete before this merges.

### DIFF
--- a/docs/networks/software-versions.mdx
+++ b/docs/networks/software-versions.mdx
@@ -36,8 +36,8 @@ Release candidates are software releases that are also released to the [Testnet]
 New features in Protocol 23:
 
 - [Release Notes](https://github.com/stellar/stellar-core/releases/tag/v23.0.0rc4)
-- Unified Events, [CAP-67](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md)
-- State Archival, [CAP-62](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md) and [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md)
+- Unified Events: [CAP-67](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md)
+- State Archival: [CAP-62](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0062.md) and [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md)
 
 ## Protocol 22 (Mainnet, December 5, 2024)
 


docs/networks/software-versions.mdx
```diff
@@ -12,6 +12,49 @@ Release candidates are software releases that are also released to the [Testnet]
 
 [testnet]: ./README.mdx
 
+## Protocol 23 (Testnet only, Jul 17, 2025)
+
+### Software
+
+| Software                   | Version                             |
+| -------------------------- | ----------------------------------- |
+| XDR                        | `todo`                              |
```
dmkozh: XDR: `v23.0`
Rust XDR: `23.0.0-rc.2`
Soroban Environment: `23.0.0-rc.2`
Stellar Core: `23.0.0rc4` (note: we're using the tags in this doc; Horizon needs to be updated as well)
Soroban Rust SDK: `23.0.0-rc.2.2`

sreuland: yes, the sw tags are the pattern, but it seems like the deb pkgs are really important as well, we should probably include both here?

docs/networks/software-versions.mdx
```diff
@@ -12,6 +12,49 @@ Release candidates are software releases that are also released to the [Testnet]
...
+| Rust XDR                   | `todo`                              |
+| Soroban Environment        | `todo`                              |
+| Stellar Core               | `stellar-core_23.0.0-2587.rc4`      |
+| Soroban Rust SDK           | `todo`                              |
+| Stellar CLI                | `todo`                              |
```
jeesunikim: Stellar CLI: [23.0.0](https://github.com/stellar/stellar-cli/releases/tag/v23.0.0)

docs/networks/software-versions.mdx
sreuland: ok, I'm adding a 3rd column here for 'Package' so can accommodate both versions.

docs/networks/software-versions.mdx
```diff
@@ -12,6 +12,48 @@ Release candidates are software releases that are also released to the [Testnet]
 
 [testnet]: ./README.mdx
 
+## Protocol 23 (Testnet only, Jul 17, 2025)
+
+### Software
+
+| Software | Version | Package |
+| --- | --- | --- |
+| XDR | `v23.0` | `?` |
```
dmkozh: I think ? is misleading here. There are no packages for libraries, only versions. We should something like 'not applicable' or `--` to signify that there is no package to be expected here.

sreuland: Yes, for XDR that is referring to the spec. For the other components, there are rust crates published, so, I plugged the rest in.

janewang:
@sreuland Thanks for adding tooling stuff!

PR #1629, opened 2025-07-25
chore: Update OpenZeppelin Contracts

stevedylandev:
This PR makes the following changes
- Updated `openzepplin-contracts.mdx` with new Wizard embed and updated content
- Updates position of OpenZeppelin in sidebar and navigation
- Added `fungible-token.mdx` to Example Contracts
- Added `non-fungible-token.mdx` to Example Contracts


stevedylandev:
Two notes:
- Please let me know if we need to adjust anything regarding OpenZeppelin's position in the sidebar and navigation. I did my best to make it a bit more visible but also not breaking patterns, totally open to changing it
- Right now the main file for OpenZeppelin's page is spelled `openzepplin-contracts.mdx`; I would like to update it but wanted to know if we could add a redirect to the updated file name with the correct spelling

ElliotFriend (approved):
This is lookin' really great! Thanks for contributing it!!

I made a couple tiny nitpicks to (hopefully lol) make our build process happier. Once the preview builds, I think this one's ready to go.

Thanks again!

stevedylandev:
> I suspect the strange action error might be due to @stevedylandev being outside the Stellar gh org? I think we'll be fine, but it's an interesting phenomenon... 🤔 

Yeah I think it's a combo of a removed/updated a route + I'm not part of the org

PR #1636, opened 2025-07-29
Add path payment example

briwylde08:


docs/learn/encyclopedia/transactions-specialized/path-payments.mdx
```diff
@@ -3,6 +3,8 @@ title: Path Payments
 sidebar_position: 60
 ---
 
+import { CodeExample } from "@site/src/components/CodeExample";
+
```
ElliotFriend, line 7: Since we're just creating an example in JS for now, let's remove the CodeExample components. We'll probably add them back in down the line when we include Python or something, but for now we don't really need it.

Issue #1513, opened 2025-04-28
Currently supported Rust versions in the docs


briwylde08:
Disc link: https://discord.com/channels/897514728459468821/966788672164855829/1365388443403944158

- Note that you are trying with a --wasm-hash arg. To use that you need to previously install the contract using stellar contract install ... command. If your are trying to make a deployment by using the .wasm file you need to use --wasm instead of --wasm-hash option
- I switched to Rust v1.81.0 and changed --wasm-hash to --wasm and it worked.

PR #1642, opened 2025-07-29
Replace old Wasm target with new

ElliotFriend:
This is a fairly blanket find-n-replace of `wasm32-unknown-unknown` to `wasm32v1-none`. The old target is still in place for where it makes sense, such as the setup page where it's telling you which target to install, depending on your version of Rust.

Refs: #1513


briwylde08 (approved):
NOICE!

PR #1644, opened 2025-07-30
updating --source to --source-account and close #1632

### LINKED ISSUE #1632: Update to `--source-account`
### What did you do?

Went to docs to learn how to deploy on stellar

Working guide

### What did you see instead?

Outdated use of `source` instead of `--source-account`:

https://developers.stellar.org/docs/build/smart-contracts/getting-started/deploy-to-testnet

https://developers.stellar.org/docs/build/smart-contracts/getting-started/deploy-increment-contract

Pfed-prog:


Pfed-prog:
> Thanks for the PR! I'm a little torn on the use of `--source-account` over `--source`.
> 
> On the one hand, `--source` is just an alias of the longer `--source-account`, and i don't think it's going away anytime soon. It's shorter and faster to use/type, too.
> 
> On the other hand, though, `--source-account` _is_ the "official" CLI parameter, and there's something to be said for using the actual name in the examples. It's also used more often in the CLI manual and `--help` outputs on the CLI.
> 
> I'm on the fence. @leighmcculloch Do you have any strong feelings one way or the other?

Yes, I have followed the tutorials and installed software. I was slightly confused on why `source` is preferred. Because there are no instructions on how to create an alias.

leighmcculloch:
We should favour the original / long form option names in examples to reduce ambiguity, rather than the short form options. For most options with aliases the alias is usually a single character, but I guess in this case the short form is still a word.

> there are no instructions on how to create an alias.

@ElliotFriend was referring to the fact that `--source` is an alias for `--source-account` in the stellar-cli. For example, the help output for the `stellar contract invoke` command says `[aliases: source]` under the `--source-account` option which means that `--source-account`, `--source`, or `-s` can be used to reference the option.
```
  -s, --source-account <SOURCE_ACCOUNT>
          Account that where transaction originates from. Alias `source`. Can be an identity (--source alice), a public key (--source GDKW...), a muxed account (--source MDA…), a secret key (--source SC36…), or a seed phrase (--source "kite urban…"). If `--build-only` or `--sim-only` flags were NOT provided, this key will also be used to sign the final transaction. In that case, trying to sign with public key will fail

          [env: STELLAR_ACCOUNT=me]
          [aliases: source]
```

Given how generic and ambiguous the term "source" is, but how specific and meaning there is with the term "source account" in the Stellar context, I think we should use `--source-account` in the docs. Unless the length is truly prohibitive from a presentation pov.

leighmcculloch:
cc @fnando @janewang for more input.

PR #1655, opened 2025-08-04
Update security docs

zbraiterman:
Add Open Worldwide Application Security Project (OWASP) references to the page about web security.


briwylde08:
Hi there @zbraiterman! Thank you so much for your contribution here. We typically try to keep the resources we list in the docs Stellar-specific and/or Web3-specific. Can you provide some more context on these resources so we can ensure they're as helpful as possible for our builders?

briwylde08:
@zbraiterman awesome, thank you! Will take a look.

PR #1656, opened 2025-08-05
Add stellar-cli plugin list page.

fnando:


docs/tools/cli/plugins.mdx
ElliotFriend, line 1: While we're touching this page, let's update the CLI command on line 30. `stellar plugins --list` is now `stellar plugin ls`, if i'm not mistaken.

scripts/stellar_cli_plugins.mjs
```diff
@@ -0,0 +1,62 @@
...
+];
+
+function exportMDX(data) {
+  const pluginsContent = data.items.reduce((buffer, item) => {
+    if (!allowedPlugins.includes(item.full_name)) {
+      return buffer;
+    }
+
+    const plugin = `
+### [${item.full_name}](${item.html_url})
```
ElliotFriend, line 17:
```suggestion
    const plugin = `### [${item.full_name}](${item.html_url})
```

Don't think it's really a _problem_, by any stretch, but this newline at the beginning of the `plugin` string gives us two blank lines after the plugins-list frontmatter. I don't know that the formatter/linter would be upset about it, but let's just make it more consistent with the rest of the documents.

PR #1681, opened 2025-08-19
Build: Upload English source files to crowdin through a cron-scheduled Github action

ElliotFriend:
This removes the crowdin tokens, configurations, and commands from the Makefile and Dockerfile, as well as from the `build:production` script in `package.json`.

Also enabling the scheduled cron job on the github action.


briwylde08 (approved):
Nice!

PR #1688, opened 2025-08-21
Update goldsky to include better info

janewang:


ElliotFriend (approved):
looks great! thank you!

Issue #1586, opened 2025-06-24
Add documentation for Galexie-backed RPC "archive node" feature


Shaptic:
At a minimum we should:
- [x] Add context on how to operate/configure it to "Admin Guide"
- [x] Highlight caveats with different types of ledger metadata across different sources
- [x] Blog announcing / highlighting the above

urvisavla:
@Shaptic 
I've added a [note](https://github.com/stellar/stellar-docs/pull/1695) regarding potential differences in metadata but I'm not sure how useful it is without explicitly highlighting what those differences are and providing instructions on how to configure the RPC to match the data lake.  Actually I'm not sure if these flags can actually be modified through the RPC configuration.

PR #1695, opened 2025-08-26
stellar-rpc: add a note regarding difference in metadata

urvisavla:
addresses https://github.com/stellar/stellar-docs/issues/1586

### DIFF
--- a/docs/data/apis/rpc/admin-guide/data-lake-integration.mdx
+++ b/docs/data/apis/rpc/admin-guide/data-lake-integration.mdx
@@ -11,7 +11,11 @@ RPC version 23.0 introduces data lake integration for the `getLedgers` endpoint,
 
 The process involves setting up a ledger data lake and then configuring your RPC node to use it.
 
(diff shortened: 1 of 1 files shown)


docs/data/apis/rpc/admin-guide/data-lake-integration.mdx
```diff
@@ -11,6 +11,8 @@ RPC version 23.0 introduces data lake integration for the `getLedgers` endpoint,
 
 The process involves setting up a ledger data lake and then configuring your RPC node to use it.
 
+> Note: The metadata of a ledger returned by `getLedgers` may vary depending on its source. When a ledger is retrieved from the RPC's local datastore, its metadata is subject to your RPC configuration. However, ledgers fetched from the data lake are typically stored with all metadata included.
```
Shaptic: I think this is perfect for the purposes of just highlighting possible differences. I would just suggest using the special box syntax on our site for it, I think it's something like

```
::: info

your content here

:::
```

Shaptic (approved):
One suggestion just for aesthetics, but otherwise this LGTM :+1:

docs/data/apis/rpc/admin-guide/data-lake-integration.mdx
urvisavla: done.

PR #1705, opened 2025-08-27
Update Core version to v23.0.1

dmkozh:


briwylde08:
Hi @dmkozh - are we okay to close this since we've updated the software versions to P23 on Mainnet?

PR #1715, opened 2025-09-03
Fix(docs): Align contract call parameter with implementation on cross-contract-call.mdx

masch:
The parameter for calling another contract is contract, but the documentation incorrectly listed it as contract_id.

This PR updates the documentation to use the correct parameter name.


masch:
> Also, hope you don't mind, but i've pushed a change that does the same thing for the *nix shell example.

Excellent. I look forward to us merging this soon.

masch:
:rocket:

PR #1761, opened 2025-09-10
Update api-providers.mdx

johncanneto:
Added Quasar (Overcat's pro RPC services) and a new column for RPC Historical Access (Both Quasar and OBSRVR now offer this)

### DIFF
--- a/config/theme/navbar.ts
+++ b/config/theme/navbar.ts
@@ -330,9 +330,19 @@ const data: NavbarItem = {
       activeBasePath: 'docs/data/apis/rpc'
     },
     {
+      to: '/docs/data/apis/rpc/providers',
+      label: 'RPC Providers',
+      activeBasePath: 'docs/data/apis/rpc/providers'
+    },
+    {
       to: '/docs/data/apis/horizon',
       label: 'Horizon',
       activeBasePath: 'docs/data/apis/horizon'
+    },
+    {
+      to: '/docs/data/apis/horizon/providers',
+      label: 'Horizon Providers',
+      activeBasePath: 'docs/data/apis/horizon/providers'
     },
     {
       to: '/docs/data/apis/migrate-from-horizon-to-rpc',


docs/data/apis/api-providers.mdx
```diff
@@ -13,24 +13,27 @@ Multiple infrastructure providers have made Stellar RPC and Horizon services ava
...
+| [QuickNode\*](https://www.quicknode.com/docs/stellar) | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| [NowNodes](https://nownodes.io/nodes/stellar-xlm) | ✅ | ✅ | ✅ | ❌ | ✅ | ❌ |
+| [Gateway](https://gateway.fm/public-rpc/) | ❌ | ✅ | ✅ | ❌ | ✅ | ❌ |
+| [Ankr](https://www.ankr.com/rpc/advanced-api/) | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| [Infstones](https://infstones.com/) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ |
+| [Obsrvr](https://www.withObsrvr.com/) | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| [GetBlock](https://getblock.io/nodes/xlm/) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ |
+| [Nodies](https://nodies.org) | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| [OnFinality](https://onfinality.io/networks/stellar) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ |
+| [Quasar](https://quasar.lightsail.network/) | ❌ | ❌ | ❌ | ❌ | ✅ | ✅ |
```
mollykarcher: I personally would remove the separate column for this in the table, and have it be another footnote option like the other two things below. If we add a column for this, why wouldn't we add a "Horizon Historical Access" column?

docs/data/apis/api-providers.mdx
```diff
@@ -13,24 +13,27 @@ Multiple infrastructure providers have made Stellar RPC and Horizon services ava
...
 
 \*_Blockdaemon, Validation Cloud, and Quicknode combine Horizon and RPC in the same endpoint._
 
 †_Blockdaemon and Validation Cloud provide full historical data for Horizon._
 
+**_RPC Historical Access is a new option for those looking to retrieve full ledger history. Currently only the [getLedgers](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getLedgers) RPC method supports this feature. You can choose one of the providers above, or create your own getLedgers archive by following [these steps](https://developers.stellar.org/docs/data/apis/rpc/admin-guide/data-lake-integration)._
+
```
mollykarcher: This isn't rendering how I think you want it to in the preview (image attached). I would also choose a symbol other than a * since that one is already used, and * / ** on a single row will cause a lot of confusion in the above chart. I would also make the wording here more consistent with the above line, something like:

> Obsrvr and Quasar provide full historical data for Horizon.

image

docs/data/apis/api-providers.mdx
```diff
@@ -13,24 +13,27 @@ Multiple infrastructure providers have made Stellar RPC and Horizon services ava
...
+| [GetBlock](https://getblock.io/nodes/xlm/) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ |
+| [Nodies](https://nodies.org) | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ |
+| [OnFinality](https://onfinality.io/networks/stellar) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ |
+| [Quasar](https://quasar.lightsail.network/) | ❌ | ❌ | ❌ | ❌ | ✅ | ✅ |
 
 \*_Blockdaemon, Validation Cloud, and Quicknode combine Horizon and RPC in the same endpoint._
 
 †_Blockdaemon and Validation Cloud provide full historical data for Horizon._
 
+**_RPC Historical Access is a new option for those looking to retrieve full ledger history. Currently only the [getLedgers](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getLedgers) RPC method supports this feature. You can choose one of the providers above, or create your own getLedgers archive by following [these steps](https://developers.stellar.org/docs/data/apis/rpc/admin-guide/data-lake-integration)._
```
janewang: @johncanneto This is not in italics. You may have to escape the **
See line 31:
```
\**_RPC........
```

docs/data/apis/api-providers.mdx
janewang: +1 on Molly's comment

Thank you so much for updating this page @johncanneto!! 🙏

johncanneto: Thanks for the feedback @mollykarcher and @janewang! My goal here is to Seperate/highlight a) the products we've built (horizon, RPC, getLedgers historical proxy, etc) and list which features each of these providers support. Molly and I discussed and we are going to split this table up by Stellar api products (horizon and RPC) and i'll add more detail on the features we support for each, as well as which providers provide what.

docs/data/apis/api-providers.mdx
johncanneto: yup - thanks. will go back fix this with the work mentioned in my comment above.

johncanneto:
Added content to each page that @mollykarcher spun up for me. I think this is ready to go, but i want to ensure that once i commit it, the original providers page will go away. True @mollykarcher?

docs/data/apis/rpc/rpc-providers.mdx
```diff
@@ -0,0 +1,47 @@
+---
+sidebar_position: 60
+title: Providers
+---
+:::info
```
mollykarcher: Please review the preview yourself because there are a ton of obvious visual issues here:
image

- This top line shouldn't be in an "info" block
- The rows in the table still have both *† symbols despite the second not being relevant to this new page
- I think it's duplicative to have *both* a footnote symbol (^) and a separate column to represent the archive functionality
- The "Dedicated Notes" footnote isn't italicized and that symbol (`\`) isn't used in table, so I'm confused about how you're trying to visually represent this
- I would consider sorting the providers in the table by the number of green checks

docs/data/apis/horizon/horizon-providers.mdx
```diff
@@ -0,0 +1,27 @@
+---
+sidebar_position: 60
+title: Providers
+---
+:::info
```
mollykarcher, line 6: - The Horizon warning block from the current page was not carried over to this page
- No one supports Futurenet so we should probably remove that column from the table
- The rows in the table still have both *† symbols but the footnotes for both are not present
- Again, I think it's duplicative to have both a column _and_ a symbol/footnote to denote the same functionality
- Don't we want a "dedicated nodes" column for this page as well?

johncanneto: thanks - will review all of these again

PR #1799, opened 2025-09-19
Add core config parameter to galexie config admin guide

chowbao:
Add core config configuration to galexie configuring admin guide


docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
```diff
@@ -44,3 +44,42 @@ sidebar_position: 20
...
+       The default core.cfg used by Galexie will enable the following core parameters:
+
+       - BACKFILL_RESTORE_META
+       - ENABLE_SOROBAN_DIAGNOSTIC_EVENTS
+       - EMIT_SOROBAN_TRANSACTION_META_EXT_V1
+       - EMIT_LEDGER_CLOSE_META_EXT_V1
+       - EMIT_CLASSIC_EVENTS
+       - BACKFILL_STELLAR_ASSET_EVENTS
+
+       If you wish to disable any of these parameters you can specify a custom core.cfg file to use that will override the default core parameters used with the Stellar Network specified in the `network` parameter.
```
mollykarcher: This feels very duplicative with the bottom section you added as well. I might just reference here that you have to supply a custom core config with the toml change, and then leave the details to the section below. You can then add the detail about Galexie defaults in that section

chowbao: Updated in a55c8a24639f71943bcec8de50b9c90b9df1f9fe

How does that look?

docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
```diff
@@ -44,3 +44,33 @@ sidebar_position: 20
         # Number of files per partition/directory
         files_per_partition = 64000
        ```
+
+     - **Core Config (Optional)**
```
urvisavla: Having two separate sections ("Core Config" and  "Use a Custom Core Config") for configuring stellar-core feels unnecessary. I think they can be combined into a single section.

chowbao: Updated in 8d22d17523bb921800492331e78c4ff28d0d20c9

Reduced to a single core config section

docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
```diff
@@ -52,3 +52,31 @@ ledgers_per_file = 1
...
+
+You can specify a custom `core.cfg` file in the Galexie `config.toml` to use that will override the default core parameters used with the Stellar Network specified in the `network` parameter.
+
+Copy an existing `core.cfg` file such as the `captive-core-pubnet.cfg` provided [here](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/configs/captive-core-pubnet.cfg) and add the following to your `config.toml`
+
+```toml
+captive_core_toml_path = "my-captive-core.cfg"
+```
+
+- Be careful about changing the quorum sets. Most likely quorum sets won't be needed for any mainnet, testnet, or futurenet usage with Galexie
```
urvisavla: Is that the case? I thought that the captive-core uses the quorum set for history archives.

chowbao: Correct. But my interpretation is that if you change the quorum set outside of the tier 1 validators that the history archives can also be off.

Like if you list one of the ded [whalestack](https://radar.withobsrvr.com/nodes/GADLA6BJK6VK33EM2IDQM37L5KGVCY5MSHSHVJA4SCNGNUIEOTCR6J5T?center=0&no-scroll=1) validators in the quorum set you have a chance to use their bad/behind history archive which would break Galexie (maybe not break but not have the expected output

urvisavla: > Be careful about changing the quorum sets.

Instead, how about we just point to this link on quorum set configuration: https://developers.stellar.org/docs/validators/admin-guide/configuring#choosing-your-quorum-set

chowbao: Oh whoops I reread it. It's a typo. I meant you won't need to change the quoroum sets 🤦

chowbao: Yeah sounds good

docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
```diff
@@ -52,3 +52,31 @@ ledgers_per_file = 1
...
+- **BACKFILL_RESTORE_META -** ledger metadata will be populated with LedgerEntryChange RESTORE type for protocol versions prior to 23
+- **ENABLE_SOROBAN_DIAGNOSTIC_EVENTS -** additional diagnostic Soroban events that are not part of the protocol will be generated while applying Soroban transactions
+- **EMIT_SOROBAN_TRANSACTION_META_EXT_V1 -** Soroban extension V1 data will be emitted
+- **EMIT_LEDGER_CLOSE_META_EXT_V1 -** ledger metadata extension V1 data will be emitted
+- **EMIT_CLASSIC_EVENTS -** classic events will be enabled and emitted for protocol `>= 23`
+- **BACKFILL_STELLAR_ASSET_EVENTS -** classic events will be enabled and emitted for protocol `<= 22`. This parameter also requires `EMIT_CLASSIC_EVENTS` to be enabled
+
+:::warning
+
+Note that any core parameters enabled with the default Galexie core config will be overwritten by the custom core.cfg provided. If the parameters are not provided in the custom core.cfg they will be set to `false`. Please explicitly set any core parameters that you wish to be enabled.
```
urvisavla: Wording is slightly confusing. How about something like 
"When you provide a custom core configuration file, it completely replaces the Galexie's default core configuration. This means you must explicitly set any parameters you want to enable, including the ones that are enabled by default in Galexie. Any parameter not listed in your custom config will be set to `false`."

docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
chowbao: Updated in 754ab364b0d5dc0f7c69b062ca6c530f11bef16d

chowbao: Updated in 754ab364b0d5dc0f7c69b062ca6c530f11bef16d

docs/data/indexers/build-your-own/galexie/admin_guide/configuring.mdx
chowbao: Yeah I like that

updated in 1e96c4879ae1c5269d44503b0346b0c6a5433824

PR #1802, opened 2025-09-19
Add Rust example to create test accounts

vinamogit:
Following previous example on invoking contracts, here is another example for the Rust SDK.

### DIFF
--- a/docs/build/guides/transactions/create-account.mdx
+++ b/docs/build/guides/transactions/create-account.mdx
@@ -704,13 +704,11 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let ledger_key = xdr::LedgerKey::Trustline(xdr::LedgerKeyTrustLine { account_id, asset });
     let response = server.get_ledger_entries(vec![ledger_key]).await?;
     let entries = response.entries.unwrap_or_default();
-    if !entries.is_empty() {
-        if let xdr::LedgerEntryData::Trustline(trustline) = entries[0].to_data() {
-            println!("USDC trustline balance (raw): {}", trustline.balance);
-        }
-    } else {
-        println!("No USDC trustline");
-    }
+        let usdc_balance = match entries.first().map(|e| e.to_data()) {
+        Some(xdr::LedgerEntryData::Trustline(t)) => t.balance,
+        _ => 0,
+    };
+    println!("USDC trustline balance (raw): {}", usdc_balance);
 
     Ok(())
 }


docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,138 @@ public final class main {
...
+            println!("No USDC trustline");
+        }
+    }
+
+    Ok(())
+}
+
+async fn wait_success(server: &Server, hash: &str, status: SendTransactionStatus) -> bool {
+    if status != SendTransactionStatus::Error {
+        loop {
```
willemneal: This should be a while loop.

docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,138 @@ public final class main {
...
+            if let Ok(tx_result) = response {
+                match tx_result.status {
+                    TransactionStatus::Success => {
+                        println!("Transaction successful!");
+                        if let Some(ledger) = tx_result.ledger {
+                            println!("Confirmed in ledger: {}", ledger);
+                        }
+                        return true;
+                    }
+                    TransactionStatus::NotFound => {
```
willemneal: If the transaction isn't found because for example it never existed this would loop for ever.

docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,138 @@ public final class main {
...
+                        println!(
+                            "Waiting for transaction confirmation... Latest ledger: {}",
+                            tx_result.latest_ledger
+                        );
+                        tokio::time::sleep(Duration::from_secs(1)).await;
```
willemneal: Should use exponential backoff which helps prevent rate limiting from the server: https://en.wikipedia.org/wiki/Exponential_backoff

docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,138 @@ public final class main {
...
+            }
+        } else {
+            println!("No USDC trustline");
+        }
+    }
+
+    Ok(())
+}
+
+async fn wait_success(server: &Server, hash: &str, status: SendTransactionStatus) -> bool {
```
willemneal: It seems strange to have this not be part of the upstream API. And it's usually better to provide a result type so that the caller can inspect the reason for the error.

ElliotFriend: @rahul-soshte is the maintainer of the upstream SDK, if i'm not mistaken. This might be something he could take into consideration there

vinamogit: Thanks for the review.
I will take your proposed changes and check other SDK for inclusion in the API.

vinamogit: Should be ok with the wait_transaction method now

docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
```diff
@@ -430,67 +425,24 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            // On success we can extract the returned value
+            let (_meta, ret_val) = tx_result.to_result_meta().expect("No result meta");
+            ret_val
+                .expect("None returned value")
+                .try_into()
+                .expect("Return value is not u32")
+        }
+        _ => {
+            return Err("Failed to create contract".into());
+        }
```
willemneal, line 439: I think there is room for improvement with this API. Especially around the result parsing. I would advise either using an [extension trait](http://xion.io/post/code/rust-extension-traits.html) or to create your own TransactionResult which wraps it.

docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,102 @@ public final class main {
...
+    let asset = Asset::new("USDC", Some(TESTNET_USDC_ISSUER))?.into();
+    let ledger_key = xdr::LedgerKey::Trustline(xdr::LedgerKeyTrustLine { account_id, asset });
+    let response = server.get_ledger_entries(vec![ledger_key]).await?;
+    if let Some(entries) = response.entries {
+        if !entries.is_empty() {
+            if let xdr::LedgerEntryData::Trustline(trustline) = entries[0].to_data() {
+                println!("USDC trustline balance (raw): {}", trustline.balance);
+            }
+        } else {
+            println!("No USDC trustline");
```
willemneal: Shouldn't this be an error? And the else of the outer block?

docs/build/guides/transactions/create-account.mdx
```diff
@@ -625,6 +624,102 @@ public final class main {
...
+    //
+    // Polling for transaction completion
+    //
+    if server
+        .wait_transaction(&hash, Duration::from_secs(15))
+        .await
+        .is_err()
+    {
+        return Err("Failed to create account".into());
+    }
```
willemneal: You can choose to ignore this nit and I'm not sure if this is formatter correctly.
```suggestion
    server
      .wait_transaction(&hash, Duration::from_secs(15))
      .await
      .map_err(|_| Err("Failed to create account"))?;
```

docs/build/guides/transactions/create-account.mdx
vinamogit: I think the API should not give an option for the entries, when there is no errors the entries are never None

willemneal: I'm not saying it should be the API just in this example if there was no trustline it would be an error right?

docs/build/guides/transactions/invoke-contract-tx-sdk.mdx
rahul-soshte: Okay thanks @willemneal I have taken a note of it, and made a corresponding issue [here ](https://github.com/rahul-soshte/rs-soroban-client/issues/28)a nd, will look at it soon outside the scope of this PR.

docs/build/guides/transactions/create-account.mdx
vinamogit: I see, this is mapped to other examples, I simplified the balance handling to make it cleared (defaulting to zero if no trustline defined)

PR #1812, opened 2025-09-22
fix history ledgers displaying its table incorrectly

jeesunikim:
**before:**

before

**after:**

after


briwylde08 (approved):

janewang (approved):
Thank you @jeesunikim!

Issue #1729, opened 2025-09-05
404 on Friendbot link from the JS SDK documentation


ElliotFriend:
The JS SDK documentation tries to link to a friendbot page in the docs for `rpc.Server.request_airdrop()` [(source)](https://stellar.github.io/js-stellar-sdk/module-rpc.Server.html#requestAirdrop).

The link is pointing to `https://developers.stellar.org/docs/learn/networks#friendbot`, which i think was moved into the `/fundamentals` directory, perhaps 🤔

Two things should be done:

- [x] fix the relevant link in the JS SDK (stellar/js-stellar-sdk#1229)
- [x] create an nginx redirect to make sure the URL is reachable for anyone else trying to reach that page. (#1816)
  - [x] (perhaps?) if there's already an attempted redirect (which i suspect there could be), we should figure out why it's not working (Edit: Looks like there isn't)

PR #1816, opened 2025-09-23
add a catch-all `/docs/learn/networks` redirect

ElliotFriend:
The 404 coming from the JS SDK will be redirected. I'll still need to test if the `#friendbot` remains at the end of the URL after the redirect has taken place.

Refs: #1729


ElliotFriend:
Tested the redirect (Chrome; macOS), and it maintains the `#friendbot` throughout the redirect

Issue #870, opened 2024-08-05
📒 Draw account summary codeblocks


JFWooten4:
### What problem does your feature solve?
The current examples run through the balances of 3 accounts.

### What would you like to see?
Rather than [plaintext](https://github.com/stellar/stellar-docs/blob/71435158a2fcdbb824eae4c6445eb5219982c64a/docs/learn/encyclopedia/transactions-specialized/clawbacks.mdx?plain=1#L215), draw a visual for each account's holdings.

### What alternatives are there?
The existing plaintext makes things harder to follow than necessary.

briwylde08:
Adding to our idea spreadsheet versus keeping in docs issues.

briwylde08:
Moved to Discussions

Pfed-prog:
I could take this issue. I want to draw images in figma.

JFWooten4:
How is this going, Pavel? 💜

JFWooten4:
Thanks a ton for coming around to help fix this! I know I can be weird but looks like the file was recently moved [here](https://github.com/stellar/stellar-docs/blob/main/docs/build/guides/transactions/clawbacks.mdx). There've been a lot of link path changes over the years, and I appreciate the repository management help coauthoring merge conflicts when things get tough.

I remember one of the first things I worked on was mired by changes. To the extent it's relevant, you can see the edit history of https://github.com/trezor/trezor-firmware/issues/1691#issue-939190410 where, within two days, a bunch of links were changed. But I know we can get this one across the finish line, and I am here to help 🙂

How can I support your efforts?[^1] Is there a draft PR set up for this with the Figma link? Your other three PRs were awesome, and I really appreciate your efforts.


[^1]: I haven't been able to get TikZ pictures to show up in Markdown either, lol. But I have had some success in this repo adding PNG snapshots. What approach have you been working with from the wireframe diagram outputs?

JFWooten4:
Awesome work from both in the PR adding an image exemplifying the first example! 🎊 There were two iterations of drawings, and I see now how this also implicates some of the changes from @karthikiyer56.

I apologize for not updating this after Karthik's update to the RPC endpoints, as their changes significantly affected how logged outputs get displayed.[^1] Part of my thinking here was that we could remove the old A, B, and C balance summary at the end of each subsection to reduce terse code viewing.[^2] I envisioned it alongside the first implementation by Pavel, which was not implemented, and here's a snippet from that I marked up:

[...]

PR #1828, opened 2025-09-27
added graphs #870

Pfed-prog:


docs/build/guides/transactions/clawbacks.mdx
```diff
@@ -4,7 +4,9 @@ description: Use clawbacks to burn a specific amount of a clawback-enabled asset
 sidebar_position: 30
 ---
 
+```js
 import { CodeExample } from "@site/src/components/CodeExample";
```
briwylde08, line 7: Is there a reason we're putting this in backticks? It shows up in the documentation with those in there. (screenshot below)
image

briwylde08:
@Pfed-prog This is great! I really like the idea of displaying the clawback flow in images. I wonder if the bar graphs are the best way to do this? Did you explore using a mermaid chart instead? The graphs are alright, but they're quite large, and it's not particularly user-friendly to look at (at least to me).

JFWooten4:
Mermaid charts are definitely the standard, and there are some examples for inspiration in the [ledgers](https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/ledgers) and [restoration](https://developers.stellar.org/docs/build/guides/archival/restore-contract-js) pages. They have the added benefit of rendering locally with text as searchable fields, although I haven't worked with building them yet. I agree that the bar charts can be a little eye-wandering despite their great raw facts, especially when the X axis changes order in its small font.

docs/build/guides/transactions/clawbacks.mdx
Pfed-prog: I am not sure, I removed backticks. I was thinking that all code needs to be in backticks.

briwylde08 (approved):
Put the first table in a collapsible menu since it's so big, but very helpful imagery! Thank you.

Pfed-prog:
@JFWooten4 Both Images were AI.

PR #1834, opened 2025-09-29
Move CLI Guides to Tools-CLI

briwylde08:


ElliotFriend:
I'll throw out there an option to rename the paths to `/docs/tools/cli/cookbook/<filename>.mdx`? since we're moving them out of the context of the "how-to guides" already, we may as well name them what we're already calling them in the CLI repo? 🤷🏻‍♂️

ElliotFriend:
For now, I've drafted stellar/stellar-cli#2231 in the stellar CLI repo. I think it would work best to create a temporary in this repo _first_ that will fix any broken links at download/build time. Then, once we have this change merged in, we can proceed with changing the CLI files.

PR #1870, opened 2025-10-01
Update configuration in RPC data lake integration guide

mollykarcher:
See https://github.com/stellar/stellar-rpc/issues/516


docs/data/apis/rpc/admin-guide/data-lake-integration.mdx
```diff
@@ -56,13 +59,19 @@ A. GCS Configuration Example
   ledgers_per_file = 1
   files_per_partition = 64000
 
-# Enable fetching historical ledgers from the datastore when not available locally
-SERVE_LEDGERS_FROM_DATASTORE = true
+[buffered_storage_backend_config]
```
Shaptic, line 62: There's actually a default one set if this isn't passed in, hence why it was omitted in the original doc.

mollykarcher: I may be doing something weird, but if I remove it I get an error at runtime when querying old ledgers 🤔

PR #1875, opened 2025-10-03
Add Rust example to upload wasm bytecode

vinamogit:


docs/build/guides/transactions/upload-wasm-bytecode.mdx
```diff
@@ -258,6 +258,100 @@ async def main():
...
+    // Read the wasm file
+    let wasm_bytes = std::fs::read(wasm_file_path)?;
+
+    // Build the xdr::HostFunction::UploadContractWasm operation
+    let upload_wasm_op = Operation::new()
+        .upload_wasm(&wasm_bytes, None)
+        .expect("Cannot create upload_wasm operation");
+
+    // Replace Your_Secret_Key with your actual secret key
+    let source_keypair = Keypair::from_secret("Your_Secret_Key")?;
```
willemneal, line 306: It is not advised to put a secret key as plaintext in a source file. It would be better to read it from a file or an ENV.

docs/build/guides/transactions/install-wasm-bytecode.mdx
```diff
@@ -258,6 +258,100 @@ async def main():
...
+
+    // Replace Your_Secret_Key with your actual secret key
+    let source_keypair = Keypair::from_secret("Your_Secret_Key")?;
+    let source_public_key = &source_keypair.public_key();
+
+    // Get account information from server
+    let account_data = server.get_account(source_public_key).await?;
+
+    // Build the Account to use in the transaction
+    let source_account = Rc::new(RefCell::new(Account::new(
```
willemneal: Rc and RefCell is pretty advanced for new rust users. The Account type seems simple enough. Can it just be cloned? Is internal mutability required? Can the transaction builder handle the sequence number?

docs/build/guides/transactions/upload-wasm-bytecode.mdx
vinamogit: Followed the same pattern than for other code examples, to comply with the rest of the doc:
> Replace `"Your_Secret_Key"` with your actual secret key.

docs/build/guides/transactions/install-wasm-bytecode.mdx
vinamogit: That's right, this is for the sequence number.
There was a discussion I think with @rahul-soshte about maybe removing this need.

docs/build/guides/transactions/upload-wasm-bytecode.mdx
ElliotFriend: yeah, in a lot of our simple examples, plaintext secret keys are present (or the placeholders like here). We _should_ discourage that, probably more overtly in the docs. But, seems out-of-scope for this particular PR, imo

docs/build/guides/transactions/install-wasm-bytecode.mdx
rahul-soshte: hi @willemneal I don't think cloning the Account is the right way. Since I think we will run into the same sequence number problem. For example, if I try to use the library in such a manner,

```
let account = Account::new(&public_key, "100").unwrap();

// First transaction - uses sequence 101 (100 + 1)
let tx1 = TransactionBuilder::new(account, Networks::testnet(), None)
    .fee(100u32)
    .build();  // Internally clones and increments to 101

// Second transaction - ALSO uses sequence 101 (100 + 1)
let tx2 = TransactionBuilder::new(account, Networks::testnet(), None)
    .fee(100u32)
    .build();  // Internally clones and increments to 101

```

both transactions use sequence 101. That will cause an error for the 2nd transaction.

rahul-soshte: @willemneal 

I have tried fixing the code by using a Mutable Reference instead of using the Rc RefCell logic
https://github.com/rahul-soshte/rs-soroban-client/pull/26

The examples work properly with this new PR. But still exploring more pros and cons if this is the most right fit for the library.

@vinamogit has some good ideas to improve the API I mean which can implemented as well.

rahul-soshte:
@willemneal @ElliotFriend @briwylde08 Everything has been resolved, please do review the PR again to check for any issues.

docs/build/guides/transactions/install-wasm-bytecode.mdx
```diff
@@ -258,6 +258,97 @@ async def main():
 asyncio.run(main())
 ```
 
+</TabItem>
+<TabItem value="rust" label="Rust">
+
+This guide will walk you through installing the Wasm of the contract using the Rust SDK: [rs-soroban-client](https://github.com/rahul-soshte/rs-soroban-client).
```
willemneal:
```suggestion
This guide will walk you through uploading the Wasm of the contract using the Rust SDK: [rs-soroban-client](https://github.com/rahul-soshte/rs-soroban-client).
```

willemneal:
The file is still named install instead of upload.

ElliotFriend (approved):

Issue #1884, opened 2025-10-06
Sample scripts in Python/Java/PHP not working


carstenjacobsen:
Page: https://developers.stellar.org/docs/build/guides/dapps/working-with-contract-specs

**Section 1**: Invoking a contract function with no parameters
* Python: Example transaction fails
* Java: update code to SDK v2.1.0
* PHP OK

**Section 3**: Getting responses from contracts
* Python: Change v3 to v4
* Java: Change v3 to v4
* PHP OK

PR #1888, opened 2025-10-07
Updating example scripts #1884

carstenjacobsen:
Updating the example scripts in Python, Java and PHP

### DIFF
--- a/docs/build/guides/dapps/working-with-contract-specs.mdx
+++ b/docs/build/guides/dapps/working-with-contract-specs.mdx
@@ -327,7 +327,7 @@ def send_transaction() -> soroban_rpc.SendTransactionResponse:
     root_keypair = Keypair.from_secret(
         "SXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
     )
-    root_account = server.load_account("GBSBL6FBPX5UHKL4AZCPUU6PXKUBYMKRUN3L4YQ4V2CCWSE7YMN2HYPB")
+    root_account = Account(account=root_keypair.public_key, sequence=1)
     contract_id = "CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE"
     transaction = (
         TransactionBuilder(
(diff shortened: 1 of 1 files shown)


docs/build/guides/dapps/working-with-contract-specs.mdx
```diff
@@ -327,16 +327,21 @@ def send_transaction() -> soroban_rpc.SendTransactionResponse:
     root_keypair = Keypair.from_secret(
         "SXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
     )
-    root_account = server.load_account("GBSBL6FBPX5UHKL4AZCPUU6PXKUBYMKRUN3L4YQ4V2CCWSE7YMN2HYPB")
+    root_account = Account(account=root_keypair.public_key, sequence=1)
```
ElliotFriend, line 330: Is there a reason to create our own account `Account()` object here? the sequence will be greater than `1`, so this transaction wouldn't actually be valid. typically the `server.load_account()` function is the way i've done it so you're sure to have the correct sequence number for the next transaction

docs/build/guides/dapps/working-with-contract-specs.mdx
```diff
@@ -327,16 +327,21 @@ def send_transaction() -> soroban_rpc.SendTransactionResponse:
...
             network_passphrase=Network.TESTNET_NETWORK_PASSPHRASE,
             base_fee=100,
         )
-        .append_invoke_contract_function_op(contract_id,"increment")
-         # mark this transaction as valid only for the next 30 seconds
+        .append_invoke_contract_function_op(
+            contract_id=contract_id,
+            function_name="increment",
+            parameters=[],
+            source=root_keypair.public_key,
```
ElliotFriend, line 342: we can omit the `source=...` account in the operation, since it can inherit from the transaction. not sure about `parameters`, though. does the sdk require us to have the empty list now?

ElliotFriend (commented):
left a few comments/notes on the python example. for java, i'll believe pretty much whatever you say 🤣

docs/build/guides/dapps/working-with-contract-specs.mdx
ElliotFriend: Actually, it occurs to me, i don't think the transaction in this example gets simulated. We might be able to use a dummy sequence number like this, if we simulate it, after all. Might need some testing.

PR #1900, opened 2025-10-08
Add stellar-cli plugin list page.

fnando:


scripts/stellar_cli_plugins.mjs
```diff
@@ -0,0 +1,61 @@
...
+
+    return buffer + plugin;
+  }, "");
+
+  const modifiedContent = `---
+title: Plugins List
+description: See a list of published Stellar CLI plugin
+sidebar_position: 30
+---
+
```
ElliotFriend, line 29: Maybe we could add a brief description here about which plugins we're including? something like "Below is a vetted list of community developed and maintained plugins for the Stellar CLI."

scripts/stellar_cli_plugins.mjs
```diff
@@ -0,0 +1,61 @@
...
+  const pluginsContent = data.items.reduce((buffer, item) => {
+    if (!allowedPlugins.includes(item.full_name)) {
+      return buffer;
+    }
+
+    const plugin = `### [${item.full_name}](${item.html_url})
+
+${item.description || ""}
+
+[${item.html_url}](${item.html_url})
```
ElliotFriend, line 18: This gives us something like the following

```md
### [OpenZeppelin/stellar-upgrader-cli](https://github.com/OpenZeppelin/stellar-upgrader-cli)

CLI that help developers to upgrade stellar contracts

[https://github.com/OpenZeppelin/stellar-upgrader-cli](https://github.com/OpenZeppelin/stellar-upgrader-cli)
```

Maybe we could clean up that last line, and use `[${item.full_name}](${item.html_url})`?

scripts/stellar_cli_plugins.mjs
```diff
@@ -0,0 +1,61 @@
+import fs from "fs-extra";
+import https from "https";
+
+const allowedPlugins = [
```
ElliotFriend: Shouldn't we include `fnando/stellar-hello-plugin` in the allowed list? 🤣

scripts/stellar_cli_plugins.mjs
fnando: I think it's useful to show the url, but maybe I can use `<url>` instead. I don't know why I haven't done it in the first place. 🤷‍♂️

scripts/stellar_cli_plugins.mjs
fnando: thinking about it, there's no reason why we shouldn't list all the plugins that were found. We can have a comment like "this is a list of all plugins made available by the community, so please review with care before using them". Wdyt?

scripts/stellar_cli_plugins.mjs
fnando: Went a different route and added "This is a list of all plugins made available by the community, so please review with care before using them."

I changed the behavior to list everything by default.

scripts/stellar_cli_plugins.mjs
```diff
@@ -0,0 +1,61 @@
...
+  const pluginsContent = data.items.reduce((buffer, item) => {
+    if (excludePlugins.includes(item.full_name)) {
+      return buffer;
+    }
+
+    const plugin = `### [${item.full_name}](${item.html_url})
+
+${item.description || ""}
+
+<${item.html_url}>
```
ElliotFriend: strange enough, this `<https://whatever.com>` is exactly what the markdown parser doesn't like lol We'd be better off keeping it like `[https://whatever.com](https://whatever.com)`

PR #1937, opened 2025-10-21
SDP-1887: Remove migration doc

JiahuiWho:
Remove the obsolete 1.x to 2.x Migration Guide

Changes include:
1. Dropped the migration guide (and its ES translation) from the docs/navigation.
2. Added the missing environment-variable details to `configuring-sdp.mdx` (EN/ES)
3. Highlighted in the tenant provisioning docs (EN and ES) that choosing `DISTRIBUTION_ACCOUNT.STELLAR.ENV` in multi-tenant setups removes fund segregation.

### DIFF
--- a/nginx/includes/redirects.conf
+++ b/nginx/includes/redirects.conf
@@ -63,6 +63,7 @@ rewrite ^/docs/anchoring-assets(/|/overview)?$ "/docs/learn/fundamentals/anchors
 rewrite ^/docs/anchoring-assets(.*)$ "/docs/learn/fundamentals/anchors" permanent;
 rewrite ^/docs/category/use-the-stellar-disbursement-platform$ "/platforms/stellar-disbursement-platform/admin-guide" permanent;
 rewrite ^/docs/stellar-disbursement-platform(.*)$ "/platforms/stellar-disbursement-platform/admin-guide$1" permanent;
+rewrite ^/platforms/stellar-disbursement-platform/admin-guide/single-tenant-to-multi-tenant-migration/?$ "/platforms/stellar-disbursement-platform/admin-guide/configuring-sdp" permanent;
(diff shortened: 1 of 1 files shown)


i18n/es/docusaurus-plugin-content-docs-sdp/current/admin-guide/configuring-sdp.mdx
ElliotFriend, line 1: As a general rule, we don't really need to mess with the files in the `i18n/es` directory. There's an automated process in place that's intended to upload the english changes, and download the translated files (currently once a day, though we're working on perhaps fine-tuning that).

That said, it should be fine to modify them directly in the github repo, but these changes _may_ be overwritten once the AI translation is run.

routes.txt
```diff
@@ -1224,7 +1224,6 @@
 /platforms/stellar-disbursement-platform/admin-guide/making-your-wallet-sdp-ready
 /platforms/stellar-disbursement-platform/admin-guide/overview
 /platforms/stellar-disbursement-platform/admin-guide/secure-operation-manual
-/platforms/stellar-disbursement-platform/admin-guide/single-tenant-to-multi-tenant-migration
```
ElliotFriend, line 1227: Is there a page that we should redirect any requests to? Adding one can help anybody who's bookmarked this page, and helps with SEO if the page has been indexed (i think? i'm not much of a marketing guy lol)

A redirect can be added to `/nginx/includes/redirects.conf`. If you need help taking care of that, I'd be more than happy to lend a hand!

JiahuiWho: Thanks for the nudge! I added a redirect from the old migration guide URL to /platforms/stellar-disbursement-platform/admin-guide/configuring-sdp, since that’s the page where we moved the content. That way anyone with the old link (or search engines) lands on the current configuration guide.

PR #1965, opened 2025-11-03
Update Galexie repo links

urvisavla:
We recently moved Stellar Galexie from the monorepo into its own repo (https://github.com/stellar/go/issues/5673). I'm updating the links here to reflect that change.


sreuland (approved):
lgtm!

PR #1970, opened 2025-11-04
[Lab] add a doc on upload and deploy contract

jeesunikim:
[Upload and Deploy Contract](https://developers-pr1970.previews.kube001.services.stellar-ops.com/docs/tools/lab/smart-contracts/upload-deploy-contract)

### DIFF
--- a/routes.txt
+++ b/routes.txt
@@ -580,6 +580,7 @@
 /docs/tools/lab/smart-contracts
 /docs/tools/lab/smart-contracts/contract-explorer
 /docs/tools/lab/smart-contracts/smart-contract-list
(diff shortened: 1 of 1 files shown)


janewang:
Thank you!

static/assets/lab/lab-already-uploaded.gif
quietbits, line 1: Animation is nice, but the files are very heavy. Users on mobile devices might not want to use up data for this (often it's not free). Or maybe we could use a video instead of a GIF, so users can play it if they want.

jeesunikim: Great suggestion. I added as a to do item for the next Docapocalyse!

PR #1994, opened 2025-11-12
docs: remove dated reference to local .stellar directory

cjonas9:
A .stellar configuration folder used to be made for each project via the CLI, but now it's in a global config in one's OS's config path (e.g. ~/.config/stellar). The Hello World frontend guide still references the per-project config folder despite it not existing.


cjonas9:
> This is a great catch! Thanks for the contribution!!
> 
> If you can, I'd love to see one more small change to go along with this. On the "Deploy to Testnet" page in this tutorial, we have a note about what the `--alias` flag does in the deploy command. Can you change it to make the alias directory correct? It should be `~/.config/stellar/contract-ids`

Happy to help! Great catch there too -- added to this PR.

PR #2053, opened 2025-11-20
[DocPoc] `Admin Guide` --> `Overview` Refresh

lijamie98:
- **Merged content**: Moved introductory content from `admin-guide/overview.mdx` to the main `README.mdx` to consolidate the platform overview in one place
- **Refreshed overview**: Updated `admin-guide/overview.mdx` to serve as an introduction/summary specifically for the Admin Guide section, including:
  - Overview of what the Admin Guide covers
  - Navigation to key sections (Getting Started, Architecture, Components, Events, etc.)
  - Quick links to important topics (Security, RPC Interface, Event System, Observers)

### DIFF
--- a/platforms/anchor-platform/README.mdx
+++ b/platforms/anchor-platform/README.mdx
@@ -34,11 +34,11 @@ The Anchor Platform implements the following Stellar Ecosystem Proposals:
 
 ## Documentation Links
 
-- **[Architecture](./architecture.mdx)** — System architecture and component overview
-- **[Getting Started](./getting-started.mdx)** — Initial setup and deployment instructions
-- **[Event Handling](./events/README.mdx)** — Event delivery, webhooks, and integration patterns
-- **[SEP Guides](../sep-guide/README.mdx)** — Implementation guides for Stellar Ecosystem Proposals
-- **[API Reference](../api-reference/README.mdx)** — Complete API documentation and reference
+- **[Architecture](./admin-guide/architecture.mdx)** — System architecture and component overview
+- **[Getting Started](./admin-guide/getting-started.mdx)** — Initial setup and deployment instructions
+- **[Event Handling](./admin-guide/events/README.mdx)** — Event delivery, webhooks, and integration patterns
+- **[SEP Guides](./sep-guide/README.mdx)** — Implementation guides for Stellar Ecosystem Proposals
+- **[API Reference](./api-reference/README.mdx)** — Complete API documentation and reference
 
 ## Additional Resources
 


platforms/anchor-platform/admin-guide/overview.mdx
```diff
@@ -3,37 +3,28 @@ title: "Overview"
...
+## Getting Started
+
+If you're new to the Anchor Platform, start with the [Getting Started guide](./getting-started.mdx) to set up your first deployment. For a high-level understanding of the system architecture, see the [Architecture documentation](./architecture.mdx).
+
+## Key Topics
+
+- **Security** — Learn about API keys, JWT authentication, and security best practices
+- **RPC Interface** — Understand the RPC request/response patterns and error handling
+- **Event System** — Configure and integrate with the event delivery system
+- **Observers** — Monitor and track transaction statuses and platform events
```
philipliu: Some of the wording is different than the `Key Features` in `platforms/anchor-platform`. For example, we call it `Event System` here but call it `Transaction processing` there.

philipliu: Do we need a section for the Callback API?

platforms/anchor-platform/admin-guide/overview.mdx
```diff
@@ -3,37 +3,28 @@ title: "Overview"
...
-[sep24-ap]: ../sep-guide/sep24/README.mdx
-[sep31-ap]: ../sep-guide/sep31/README.mdx
-[sep45-ap]: ../sep-guide/sep45/README.mdx
+# Admin Guide Overview
+
+This Admin Guide provides comprehensive documentation for setting up, configuring, and managing the Anchor Platform. Whether you're getting started with your first deployment or looking to optimize an existing installation, this guide covers everything you need to know.
+
+## What You'll Learn
+
+The Admin Guide covers:
```
philipliu: What's the difference between this section and `Key Topics` below?

platforms/anchor-platform/admin-guide/overview.mdx
```diff
@@ -3,37 +3,28 @@ title: "Overview"
...
-[anchor-platform-github]: https://github.com/stellar/java-stellar-anchor-sdk
-[sep1-ap]: ../sep-guide/sep1/README.mdx
-[sep6-ap]: ../sep-guide/sep6/README.mdx
-[sep10-ap]: ../sep-guide/sep10/README.mdx
-[sep24-ap]: ../sep-guide/sep24/README.mdx
-[sep31-ap]: ../sep-guide/sep31/README.mdx
-[sep45-ap]: ../sep-guide/sep45/README.mdx
+# Admin Guide Overview
+
+This Admin Guide provides comprehensive documentation for setting up, configuring, and managing the Anchor Platform. Whether you're getting started with your first deployment or looking to optimize an existing installation, this guide covers everything you need to know.
```
philipliu: This seems a bit wordy and I don't think we will cover any optimization topics. What about something like:

"This admin guide walks you through installing, configuring, and running the Anchor Platform in production."

platforms/anchor-platform/README.mdx
```diff
@@ -7,8 +7,14 @@ description: "Learn how the Anchor Platform (AP) simplifies the process of build
 
 # Anchor Platform
 
+The Anchor Platform is the easiest and fastest way to deploy [anchor services](/docs/learn/fundamentals/anchors) compatible with [Stellar Ecosystem Proposals (SEPs)](https://github.com/stellar/stellar-protocol/tree/master/ecosystem).
```
philipliu: I would avoid phrasing like "easiest" or "fastest" especially given the issues we saw with some partners integrating AP. I think we can remove this section and move the doc links into the paragraph below.

platforms/anchor-platform/README.mdx
```diff
@@ -7,8 +7,14 @@ description: "Learn how the Anchor Platform (AP) simplifies the process of build
...
+
 The Anchor Platform provides a complete set of tools and APIs for building on and off-ramp services on the Stellar network. With standardized interfaces and full implementations of key Stellar Ecosystem Proposals (SEPs), it simplifies integration with Stellar-based wallets and exchanges, enabling you to focus on your core business logic rather than protocol implementation details.
 
+The goal of the Anchor Platform is to handle all Stellar-specific functionality and requirements for running an anchor, allowing businesses to focus on the core business logic necessary to provide these services.
+
+The Anchor Platform accomplishes this by implementing the ecosystem's standardized APIs (SEPs) for wallets, exchanges, and other applications to consume, while offering a set of backend APIs for businesses to provide information specific to them, such as transaction fees, exchange rates, and off-chain transaction statuses.
```
philipliu: We're repeating a few things here that are already in the previous paragraph like:

- Focus on core business logic
- Implements SEPs

And the second paragraph is the same as the first one but in more detail. I would figure out a way to combine these two paragraphs with the previous one.

platforms/anchor-platform/admin-guide/overview.mdx
lijamie98: Simplified to avoid inconsistency in the future

lijamie98: I have removed this section and replace with the document links

platforms/anchor-platform/admin-guide/overview.mdx
lijamie98: This is removed to avoid future inconsistency.

platforms/anchor-platform/README.mdx
lijamie98: This paragraph is removed. The doc links is also moved to this doc from the Admin Guide

platforms/anchor-platform/README.mdx
lijamie98: I removed several paragraphs to keep the document focused.

PR #2062, opened 2025-11-24
docs: Add Homebrew installation instructions for Linux

kaankacar:
## Problem
The current documentation assumes `brew` is available on Linux or that the user knows how to install it. An audit of the "Getting Started" guide identified this as a blocker for users on Linux, where commands like `brew install stellar-cli` fail with "command not found".

## Solution
This PR adds explicit instructions for installing Homebrew on Linux within the `setup.mdx` file.

**Changes:**
- Added the standard Homebrew installation command for Linux.
- Added the command to configure the shell environment.

## Verification
- Verified that the commands are the standard installation steps from `brew.sh`.
- The changes are restricted to the Linux tab in the `Install the Stellar CLI` section.

### DIFF
--- a/docs/build/smart-contracts/getting-started/setup.mdx
+++ b/docs/build/smart-contracts/getting-started/setup.mdx
@@ -119,7 +119,7 @@ If you don't have Homebrew installed, you can install it with the following comm
 Then, follow the instructions to add Homebrew to your PATH:
 
 ```bash
-eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+eval "$(~/.linuxbrew/bin/brew shellenv)"
 ```
 
 Once Homebrew is installed, install the Stellar CLI:


docs/build/smart-contracts/getting-started/setup.mdx
```diff
@@ -108,7 +108,21 @@ Install with cargo from source:
...
+If you don't have Homebrew installed, you can install it with the following command:
+
+```bash
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+```
+
+Then, follow the instructions to add Homebrew to your PATH:
+
+```bash
+eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
```
ElliotFriend: this might need some more work. `/home/linuxbrew` likely won't exist on a linux system, and (I'm assuming) is a placeholder here. `eval "$(~/.linuxbrew/bin/brew shellenv)"` might be a better fit

ElliotFriend:
Interesting! hadn't seen that lol. i even looked at homebrew's linux documentation, but somehow missed that part 🤦🏻‍♂️ now i'm second-guessing myself 🤣 

I've never used homebrew on a linux system, so i'm glad to learn something new about it! let's actually revert that change lol. it does appear that the default, best-supported option is to have the installer create and install to `/home/linuxbrew`, so let's go with that. Thanks!

PR #2078, opened 2025-12-01
Add network IDs to networks page

leighmcculloch:
### What
Document the Network IDs for each network alongside their network passphrases for Mainnet, Testnet, and Futurenet. Remove duplicate passphrase listing and reference the existing section instead.

### Why
Network IDs are used in transaction signing, contract address generation, and hash uniqueness across networks but are not well documented. They show up in a variety of places, like xdr, and auth signing, and are are concept exposed in the contract sdk. They should be discussed.

### DIFF
--- a/docs/networks/README.mdx
+++ b/docs/networks/README.mdx
@@ -19,7 +19,7 @@ Stellar has three networks: the public network (Mainnet, also called Pubnet or t
 | --- | --- | --- | --- |
 | **Purpose** | Production network | Stable testing environment | Bleeding-edge feature testing |
 | **Network Passphrase** | `Public Global Stellar Network ; September 2015` | `Test SDF Network ; September 2015` | `Test SDF Future Network ; October 2022` |
-| **Network ID** | <code style={{overflow-wrap:'anywhere'}}>7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979</code> | <code style={{overflow-wrap:'anywhere'}}>cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472</code> | <code style={{overflow-wrap:'anywhere'}}>a3a1c6a78286713e29be0e9785670fa838d13917cd8eaeb4a3579ff1debc7fd5</code> |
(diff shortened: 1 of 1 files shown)


docs/networks/README.mdx
```diff
@@ -19,6 +19,7 @@ Stellar has three networks: the public network (Mainnet, also called Pubnet or t
 | --- | --- | --- | --- |
 | **Purpose** | Production network | Stable testing environment | Bleeding-edge feature testing |
 | **Network Passphrase** | `Public Global Stellar Network ; September 2015` | `Test SDF Network ; September 2015` | `Test SDF Future Network ; October 2022` |
+| **Network ID** | `7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979` | `cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472` | `a3a1c6a78286713e29be0e9785670fa838d13917cd8eaeb4a3579ff1debc7fd5` |
```
leighmcculloch: The long hashes unfortunately make the table a little wide. @ElliotFriend it might be reasonable to introduce a custom style here to force these hashes to line break mid-word. But if that not super trivial I don't think it's of huge importance, unless you think it is.

docs/networks/README.mdx
```diff
@@ -97,6 +98,18 @@ Passphrases serve two main purposes: (1) used as the seed for the root account (
 
 Many SDKs have the passphrases hardcoded for Stellar's networks. If you’re running a private network, you’ll have to manually pass in a passphrase to be used whenever transaction hashes are generated. All of Stellar’s official SDKs allow you to use a network with a custom passphrase.
 
+## Network IDs
+
+Each Stellar network also has a network ID, which is the SHA-256 hash of the network passphrase. The network ID is used in transaction signing, contract address generation, and other things and ensures that hashes and ids generated are different on each network.
```
leighmcculloch:
```suggestion
Each Stellar network also has a network ID, which is the SHA-256 hash of the network passphrase. The network ID is used in transaction signing and contract address generation and ensures that hashes and IDs generated are different on each network.
```

docs/networks/README.mdx
ElliotFriend: Probably the simplest way to do it might look like this:

```markdown
| Feature | Mainnet | ... |
| --- | --- | --- |
| **Network ID** | <code style="overflow-wrap: anywhere;">7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979</code> | ... |
```

This is certainly more cumbersome. Though, it does prevent the `overflow-wrap` from also affecting other rows/columns in the table, which might be the result of creating some kind of custom component with styles, etc. for the whole table (i'm unaware of any way to apply css styles on a per-cell/row/column basis through markdown)

leighmcculloch: Done, with some tweaks, thanks.

Issue #2077, opened 2025-11-30
getEvents topics parameter documentation is ambiguous about array structure


leighmcculloch:
### What version are you using?
Documentation at developers.stellar.org (getEvents RPC method)

### What did you do?
Tried to implement a transfer event filter for the getEvents RPC endpoint following the documentation.

### What did you expect to see?
Clear documentation that `topics` is an **array of arrays**, where:
- The outer array contains multiple filter options (OR logic)
- Each inner array is a complete segment matcher (e.g., `[topic1, topic2, topic3, ...]`)

### What did you see instead?
The documentation shows `topics` as `array[array]` with a description of `SegmentMatcher`, but doesn't clearly communicate that each inner array represents a complete filter with all 4 positions. It's not clear what a "SegmentMatcher" is.

<img width="857" height="274" alt="Image" src="https://github.com/user-attachments/assets/aafe8141-55f8-4d45-9e4a-4a6f9987cce5" />

Multiple AI agents get confused about the description, and I do too. It's super easy to think the topic list should look like this, even after reading the docs screenshotted above:
```json
"topics": [
  ["AAAADwAAAAh0cmFuc2Zlcg=="],
  ["GABC..."],
  ["*"],
  ["*"]
]
```

When the correct structure is:
```json
"topics": [
  ["AAAADwAAAAh0cmFuc2Zlcg==", "GABC...", "*", ...]
]
```

### Suggested improvement
Add a clear explanation like:
> The `topics` parameter is an array of topic filters. Each filter is itself an array of zero or more `SegmentMatcher` elements. A `SegmentMatcher` may be one of the following: [current list of segment matchers as shown]

Also, change the `array[array]` to be more descriptive, something like `array[array[SegmentMatcher]]`.

Maybe there are other improvements possible too? We'd benefit from some brainstorming on how to better communicate getEvents filters.

cc @stellar/platform-eng

leighmcculloch:
Also for consideration:
- https://github.com/stellar/stellar-rpc/issues/552

PR #2084, opened 2025-12-03
Improve `getEvents` documentation surrounding filters

Shaptic:
Closes https://github.com/stellar/stellar-docs/issues/2077.

This clarifies the format of a topic filter and introduces an example using the new `**` topic syntax. Even though the filters were [already defined](https://github.com/stellar/stellar-docs/blob/main/openrpc/src/stellar-rpc/schemas/EventFilters.json#L44-L52), they weren't being rendered in the generated docs. Unfortunately another suggestion isn't possible:

> Also, change the `array[array]` to be more descriptive, something like `array[array[SegmentMatcher]]`.

Because `type`s must be OpenRPC primitive types.

### DIFF
--- a/docs/platforms/anchor-platform/api-reference/platform/rpc/anchor-platform.openrpc.json
+++ b/docs/platforms/anchor-platform/api-reference/platform/rpc/anchor-platform.openrpc.json
@@ -1271,8 +1271,8 @@
               "description": "The domain of the client.",
               "type": "string"
             },
-            "client_name": {
-              "description": "The domain of the client.",
+            "request_client_ip_address": {
+              "description": "The IP address of the client requesting the transaction.",
               "type": "string"
             },
             "customers": {
@@ -1987,8 +1987,8 @@
                     "description": "The domain of the client.",
                     "type": "string"
                   },
-                  "client_name": {
-                    "description": "The domain of the client.",
+                  "request_client_ip_address": {
+                    "description": "The IP address of the client requesting the transaction.",
                     "type": "string"
                   },
                   "customers": {
--- a/openrpc/src/anchor-platform/schemas/platform_transaction.json
+++ b/openrpc/src/anchor-platform/schemas/platform_transaction.json
@@ -92,10 +92,6 @@
         "description": "The domain of the client.",
         "type": "string"
       },
-      "client_name": {
-        "description": "The domain of the client.",
-        "type": "string"
-      },
       "request_client_ip_address": {
         "description": "The IP address of the client requesting the transaction.",
         "type": "string"


openrpc/src/stellar-rpc/schemas/EventFilters.json
```diff
@@ -30,7 +30,7 @@
     "TopicFilters": {
         "title": "topics",
         "type": "array",
-        "description": "List of topic filters. If omitted, query for all events. If multiple filters are specified, events will be included if they match any of the filters. Maximum 5 filters are allowed per request.",
+        "description": "A list of topic filters. Each filter is itself an array of one to four `SegmentMatcher` elements (see below). If omitted, query for all events. If multiple filters are specified, events will be included if they match any of the filters.",
```
leighmcculloch, line 33: I thought the maximum number of filters was 5, so we could do a double star as the 5th one, otherwise it isn't possible to match on events that have 4 or more, only 3 or more or exactly 4.

openrpc/src/stellar-rpc/examplePairingObjects/EventsPairs.json
```diff
@@ -11,6 +11,18 @@
             "$ref": "#/components/examples/TwoNativeTransferEvents"
         }
     },
+    "WildcardTransferEventsExample": {
```
leighmcculloch: Could we use the wildcard on the first example that is displayed? It's the native asset streaming, and it would be very reasonable to use `**` on it. Folks consuming tokens should be encouraged to stream for `transfer, *, *, **` so that their streams work for both stellar assets and tokens implementing sep-41. It would also mean that a reader doesn't have to go and know to chase a wildcard example which is part of the issue today.

leighmcculloch (commented):
Thanks for taking this. One ask and one question.

openrpc/src/stellar-rpc/schemas/EventFilters.json
Shaptic: `**` as the 5th actually [doesn't count](https://github.com/stellar/go-stellar-sdk/blob/master/protocols/rpc/get_events.go#L242-L248) towards the limit :o updated to clarify this

openrpc/src/stellar-rpc/examplePairingObjects/EventsPairs.json
Shaptic: sure yeah! updated

Shaptic: oh I misread this, I thought you meant move the wildcard example to be the first one displayed, now updated

urvisavla:
>Even though the filters were [already defined](https://github.com/stellar/stellar-docs/blob/main/openrpc/src/stellar-rpc/schemas/EventFilters.json#L44-L52), they weren't being rendered in the generated docs. 

I previously highlighted this exact [issue](https://stellarfoundation.slack.com/archives/C049E67BFSP/p1751578225475119) when working on the getevents documentation but it wasn't addressed at the time.

leighmcculloch (approved):

PR #2092, opened 2025-12-04
Add guide for supporting payments from and to C accounts

brunomlr:
Introduces documentation explaining how to send and receive payments to and from C-Accounts on the Stellar network. Covers differences in protocol, handling memos, using Horizon and RPC APIs, and provides code example references for developers integrating C-Account support.

### DIFF
--- a/docs/build/guides/transactions/send-and-receive-c-accounts.mdx
+++ b/docs/build/guides/transactions/send-and-receive-c-accounts.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Send to and receive payments from C Accounts"
-description: "Learn to send payments to and receive payments from C Accounts on the Stellar network."
+title: "Send to and receive payments from Contract Accounts"
+description: "Learn to send payments to and receive payments from Contract Accounts on the Stellar network."
 sidebar_position: 7
 ---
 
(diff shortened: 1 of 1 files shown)


docs/build/guides/transactions/send-and-receive-c-accounts.mdx
```diff
@@ -0,0 +1,207 @@
...
+### Receiving payments from C-accounts that require a memo
+
+Entities such as Exchanges and Asset Issuers usually use omnibus accounts, using a shared receiving address and a unique identifier memo like:
+
+- **Account:** `GBLVHX33XGOBDOXK7ERDL34NVH6WW7VTT2OBAHPJ7G3D423HBG5NOMY7`
+- **Memo ID:** `123456789`
+
+To enable support for deposits coming from C addresses, nothing needs to change in the information displayed to customers or the deposit accounts used by the exchange. Wallets will still use the G account and memo provided to send a payment to the exchange's G address including the memo with the unique ID.
+
+That's the main difference between receiving from customers using G or C accounts: instead of looking at the memo field to identify the individual account, you'll look at the muxed ID. Everything else stays the same.
```
tomerweller: We should probably introduce the term "muxed ID" somewhere before using it

tomerweller (commented):
Thanks. This is awesome. 

On muxed/memo ids I think we do a good job in explaining how to receive payments, but wallets might need some guidance on how to send. 

If the wallet is presented with an address plus a memo id, what is it expected to do? Send a payment and add a memo id? make a payment to a muxedaccount? invoke a sac transfer with a muxedaccount? How does that change for contract account senders vs. classic senders?  

Similar, but what is a client expected to do when presented with an M address?

briwylde08 (approved):

PR #2097, opened 2025-12-05
Feature/google translate integration

kaankacar:
This PR adds a Google Translate widget to the navbar, allowing users to translate the documentation into various languages.

### DIFF
--- a/.github/workflows/crowdin-download-workflow.yml
+++ b/.github/workflows/crowdin-download-workflow.yml
@@ -8,8 +8,8 @@ on:
         description: "Branch to run this workflow on"
         required: true
         default: "main"
-  schedule:
-    - cron: "20 14,18,22 * * 1,2,3,4,5" # Runs at 20 minutes past beginning/middle/end of day Mon-Fri (UTC-6) (an hour after the upload action)
+  # schedule:
+  #   - cron: "20 14,18,22 * * 1,2,3,4,5" # Runs at 20 minutes past beginning/middle/end of day Mon-Fri (UTC-6) (an hour after the upload action)
 
 permissions:
   contents: write
--- a/.github/workflows/crowdin-upload-workflow.yml
+++ b/.github/workflows/crowdin-upload-workflow.yml
@@ -3,8 +3,8 @@ name: Crowdin Upload Action
 on:
   workflow_call:
   workflow_dispatch:
-  schedule:
-    - cron: "20 13,17,21 * * 1,2,3,4,5" # Runs at 20 minutes past beginning/middle/end of day Mon-Fri (UTC-6)
+  # schedule:
+  #   - cron: "20 13,17,21 * * 1,2,3,4,5" # Runs at 20 minutes past beginning/middle/end of day Mon-Fri (UTC-6)
 
 jobs:
   crowdin-upload:


docusaurus.config.ts
```diff
@@ -31,7 +31,27 @@ const config: Config = {
     defaultLocale: DEFAULT_LOCALE,
     locales: ["en", "es"],
   },
+  scripts: [
+    'https://buttons.github.io/buttons.js',
+    'https://use.fontawesome.com/221fd444f5.js',
```
ElliotFriend: i don't think we need these scripts

docusaurus.config.ts
```diff
@@ -31,7 +31,27 @@ const config: Config = {
...
+    'https://translate.google.com/translate_a/element.js?cb=googleTranslateElementInit',
+  ],
   plugins: [
+    () => ({
```
ElliotFriend: i'd love if we could have this as a defined plugin in its own file. i'm thinking similar to how we have this [`route-export-plugin` defined](https://github.com/stellar/stellar-docs/blob/main/src/route-export-plugin.js)

docusaurus.config.ts
```diff
@@ -155,6 +181,12 @@ const config: Config = {
...
           'aria-label': "GitHub",
         },
+        {
+          type: 'html',
+          position: 'right',
+          value:
+            '<div id="google_translate_element" style="display:inline-block"></div><a class="navbar__link false" href="#" onclick="new google.translate.TranslateElement({pageLanguage: \'en\', includedLanguages: \'af,sq,am,en,fa,ar,ps,ja,zh-CN,hy,az,eu,be,bn,bs,bg,ca,ceb,ny,zh-TW,co,hr,cs,da,nl,eo,et,tl,fi,fr,fy,gl,ka,de,el,gu,ht,ha,haw,iw,hi,hmn,hu,is,ig,id,ga,it,jw,kn,kk,km,ko,ku,ky,lo,la,lv,lt,lb,mk,mg,ms,ml,mt,mi,mr,mn,my,ne,no,pl,pt,pa,ro,ru,sm,gd,sr,st,sn,sd,si,sk,sl,so,es,su,sw,sv,tg,ta,te,th,tr,uk,ur,uz,vi,cy,xh,yi,yo,zu\'}, \'google_translate_element\'); return false;"><svg viewBox="0 0 24 24" width="20" height="20" aria-hidden="true" style="vertical-align: text-bottom"><path fill="currentColor" d="M12.87 15.07l-2.54-2.51.03-.03c1.74-1.94 2.98-4.17 3.71-6.53H17V4h-7V2H8v2H1v1.99h11.17C11.5 7.92 10.44 9.75 9 11.35 8.07 10.32 7.3 9.19 6.69 8h-2c.73 1.63 1.73 3.17 2.98 4.56l-5.09 5.02L4 19l5-5 3.11 3.11.76-2.04zM18.5 10h-2L12 22h2l1.12-3h4.75L21 22h2l-4.5-12zm-2.62 7l1.62-4.33L19.12 17h-3.24z"></path></svg></a>',
```
ElliotFriend: I wonder if we could turn this into a custom React component, too? (I don't know if that kind of thing would be supported in the docusaurus navbar, though). Or, we could move this part into a `GOOGLE_TRANSLATE_ELEMENT` string in `/config/constants.ts` so it doesn't feels "cluttered" in the config file?

src/css/custom.scss
```diff
@@ -231,8 +222,21 @@ select[data-testid="example-pairing-select"] {
...
   line-height: 1;
   display: inline-block;
+}
+
+/* Fix for Google Translate making headers invisible */
+html.translated-ltr .hash-link,
+html.translated-rtl .hash-link {
+  opacity: 1 !important;
+  visibility: visible !important;
+}
```
ElliotFriend: i think this may not be making it all the way through the `docusaurus build` process. i'm still seeing some invisible headers

Image

ElliotFriend (commented):
I think this is really promising! Thank you!!! 🎉 

I've added a couple notes throughout the PR changes.

Wondering if we might be able to adjust the spacing on these elements to make things a little nicer?

Image

Also, I'm seeing some strange behavior where the button doesn't appear to work in the following scenarios:

- On mobile, the button in the "hamburger menu" doesn't seem to do anything
- If the translation button is clicked and then "dismissed" (either closed outright, or if you navigate to some page without selecting a language), the button won't do anything if it's clicked again (unless you refresh the page)

Some of these comments may just be "it's weird" kinds of things that we'll have to just live with and/or fix in the future.

PR #2101, opened 2025-12-05
Migrate example go code to use go-stellar-sdk

sreuland:
Migrate the go code examples in docs to reference the now renamed `github.com/stellar/go-stellar-sdk` repo and go module. Changes any existing references of Go sdk from `github.com/stellar/go` to `github.com/stellar/go-stellar-sdk`

also bumped the ingest pipeline example to use the aws public blockchain bucket.

Closes https://github.com/stellar/go-stellar-sdk/issues/5688

PR #2122, opened 2025-12-11
Replace go -> go-stellar-sdk

mollykarcher:


sreuland (approved):
whoa, I missed these on [earlier pass](https://github.com/stellar/stellar-docs/pull/2101), I clearly didn't have grep filters correct, thanks for cleanup here!

PR #2055, opened 2025-11-20
feat: more approachable & in-depth indexer intro

chadoh:
Clean up the Data section's sidebar and add MUCH more info & context to the "Data -> Indexers" page.

While mostly adding info, this removes the small Goldsky tutorial+screenshots that had been on this page. I'm not sure our documentation is the right place to walk through a partner's solution in this way. We already link out to Goldsky's documentation, which is quite good, containing more & better screenshots than what we had.  Additionally, this sort of tutorial-via-screenshots content did not fit as well within the reworked Indexers homepage, which is a bit higher-level. I wasn't sure it was worth keeping this content anywhere, and decided to remove it for now.

PR #2146, opened 2025-12-19
Horizon ain't deprecated yet

chadoh:
"deprecated", to my engineering-trained mind, means "there is a better alternative available that you should now switch to."

There is ***no*** viable alternative to Horizon today. We therefore cannot declare it to be deprecated.

When I read that it is _already_ deprecated, it makes me think I can avoid using it in a project. If I cannot realistically avoid using it, at least not without great cost in both engineering research time and infrastructure bills, then the assertion that it is already deprecated will cause me distress.

Once we have good Portfolio API-style indexing services in place, with freemium models appropriate for the whole gamut of hackathon-startups-to-enterprises, then we can consider Horizon to be deprecated. Until then, there's just no other reasonable way to get a whole host of necessary data, such as token balances and circulating supply.

This updates the "Horizon is deprecated" warning to say that it _will_ be deprecated in the future, deep-linking to the "Portfolio APIs" section of the [newly-revamped indexers page][1].

  [1]: https://github.com/stellar/stellar-docs/pull/2055

### DIFF
--- a/docs/learn/fundamentals/stellar-stack.mdx
+++ b/docs/learn/fundamentals/stellar-stack.mdx
@@ -37,7 +37,7 @@ SDF does not provide a publicly available RPC endpoint for Mainnet. Developers s
 
 :::warning
 
-Horizon is considered deprecated in favor of Stellar RPC. While it will continue to receive updates to maintain compatiblity with upcoming protocol releases, it won't receive significant new feature development.
+Horizon is nearing end-of-life and will eventually be deprecated in favor of Stellar RPC and [Portfolio APIs](../../data/indexers/README.mdx#portfolio-apis). While it will continue to receive updates to maintain compatibility with upcoming protocol releases, it won't receive new feature development.
 
 :::
 


JFWooten4:
LOL um thank you uWu

docs/learn/fundamentals/stellar-stack.mdx
```diff
@@ -37,7 +37,7 @@ SDF does not provide a publicly available RPC endpoint for Mainnet. Developers s
 
 :::warning
 
-Horizon is considered deprecated in favor of Stellar RPC. While it will continue to receive updates to maintain compatiblity with upcoming protocol releases, it won't receive significant new feature development.
+Horizon is nearing end-of-life and will eventually be deprecated in favor of Stellar RPC and [Portfolio APIs](../../data/indexers/README.mdx#portfolio-apis). While it will continue to receive updates to maintain compatiblity with upcoming protocol releases, it won't receive significant new feature development.
```
johncanneto: i would just remove the word significant in the following statement - " it won't receive significant new feature development." there will be no new feature development and i want to be clear on that.

JFWooten4: What about people who keep running it? Anyone could propose minor bug fixes if they need indexing and use it. Our entire tech stack @blocktransfer relies on running Horizon atm.

chadoh: To my mind, "bug fixes" do not count as "feature development." Saying "no new feature development" does not imply no bug fixes or no accepted PRs.

JFWooten4: Other community members may need to add very basic functionality[^ss] over time as the protocol develops

[^ss]: Consider a protocol upgrade which allows contracts to invocate the DEX. Upon calling, I understand the resulting txMeta would require advanced parsing and indexing not present in the current implementation. There are also backwards-compatible upgrades from CAPs like event emmits which I understand require adoption in this 'legacy' system.

briwylde08 (approved):

PR #2174, opened 2026-01-07
Add asset comparison

janewang:


docs/tokens/README.mdx
```diff
@@ -11,78 +11,239 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
-## Issuing assets on Stellar
+- **Built-in Smart Contract (Stellar Asset Contract)**  
+  A [Stellar Asset Contract (SAC)][sac] can optionally be deployed to a contract address for the asset, implementing the SEP-41 token interface so the asset can be used in smart contracts.
 
-Stellar has first-class support for asset tokenization — issuing an asset can be done using a [built-in transaction](./quickstart.mdx) without the development of a smart contract.
+### How they work
 
-Stellar’s transactions are fast and cost-effective, making the network great for remittances and micropayments. It also has built-in features for compliance, asset management, and auditing. If you are looking to perform transfers of value, issuing assets on Stellar has all the needed capabilities.
+- For protocol-level behavior and trustline semantics, see [Asset Design Considerations](./control-asset-access.mdx).
+- Implemented at the protocol level, the **Stellar Asset Contract (SAC)** provides a smart contract interface that enables Stellar assets to interoperate seamlessly with Soroban smart contracts.
```
briwylde08: Do we want to say "Stellar smart contracts" instead of "Soroban smart contracts" here? Or define "Soroban" early on. I guess since we're using Soroban a lot in this page, defining it early on might be the best route as new users coming to this page might not know what Soroban is.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,239 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- Transfers between accounts and contracts using SAC result in the same underlying trustline debit/credit.
```
briwylde08: I got caught up on this a bit. Wondering if this makes more sense: "Transfers between accounts and contracts using SAC resolve to the same trustline balance updates the Stellar protocol has always used."

Feel free to take it or leave it.

docs/tokens/README.mdx
janewang: Yes, we say "Stellar", thank you!

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+to help issuers select the most appropriate model for their use case.
 
-Several factors can help you determine whether to issue an asset on Stellar or create a contract token with a smart contract for your project.
+## 1. Stellar Classic Assets (with Built-in [Stellar Asset Contract][sac])
 
-However:
+### What this category encompasses
 
-### TL;DR
+All assets issued on the Stellar ledger (by a `G…` address) are **Classic Stellar Assets**. These assets can be represented in two forms:
```
leighmcculloch: > issued on the Stellar ledger

The post is using "on the Stellar ledger" as a distinction for assets issued by a G address, but that language is confusing because all types of tokens are issued on the Stellar ledger.

I think the original language doesn't need changing and is more accurate that states that there are assets "issued by Stellar accounts (`G...` addresses)" and assets issued by Wasm contracts. We can then separate those latter assets into two categories, those that implement SEP-41 and those that implement SEP-57.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
 
-If possible, we recommend issuing a Stellar asset and using the SAC to interact with that asset in smart contracts or to send to contract addresses. More on why below.
+- **Native Classic Asset**  
+  Held in trustlines and transferred via native payment operations.
```
leighmcculloch: > Native Classic Asset Held in trustlines

The native asset is not held in trust lines. The word native has special meaning, it refers to the native asset that exists at network creation, that every account can hold without the use of a trustline. i.e. the Lumen on mainnet.

> native payment operations

I think the term native here is being used to describe the Payment operation and PathPayment operations, but it's confusing because native has special meaning.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- Stellar Classic Assets (with built-in Stellar Asset Contract),
+- [SEP-41][sep-41] Contract Tokens, and
+- Regulated [SEP-57][sep-57] Tokens (Tokens for Regulated EXchanges / T-REX)
 
-1. Assets issued by Stellar accounts (`G...` addresses) and their built-in [Stellar Asset Contract (SAC)][sac] implementation, and
-2. [Contract tokens][ti] issued by a deployed WASM contract (`C...` addresses).
+to help issuers select the most appropriate model for their use case.
 
-Several factors can help you determine whether to issue an asset on Stellar or create a contract token with a smart contract for your project.
+## 1. Stellar Classic Assets (with Built-in [Stellar Asset Contract][sac])
```
leighmcculloch: This page previously didn't require the term "Classic" and did a good job talking about both without needing that term. I think at some point, sometime ago now, we had decided to not use the term. Why are we bringing it back now?
```suggestion
## 1. Assets Issued by `G...` Accounts (with Built-in [Stellar Asset Contract][sac])
```

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
-Stellar’s transactions are fast and cost-effective, making the network great for remittances and micropayments. It also has built-in features for compliance, asset management, and auditing. If you are looking to perform transfers of value, issuing assets on Stellar has all the needed capabilities.
+- For protocol-level behavior and trustline semantics, see [Asset Design Considerations](./control-asset-access.mdx).
+- Implemented at the protocol level, the **Stellar Asset Contract (SAC)** provides a smart contract interface that enables Stellar assets to interoperate seamlessly with Stellar smart contracts.
+- Transfers between accounts and contracts using SAC resolve to the same trustline balance updates the Stellar protocol has always used.
 
-Stellar assets:
+### Strengths
 
-- Are compatible with Stellar ecosystem products (such as Stellar wallets) and other ecosystem products (such as exchanges).
+- Has the benefits of native Stellar assets (fast, low cost), making these great for remittances and micropayments.
```
leighmcculloch: Unclear what the intent of the use of the word native here, given its special meaning.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- Can be used in Stellar smart contracts via SAC with a [SEP-41][sep-41] interface.
+- Maintains issuer controls and trustline semantics even when used from contracts.
+- Wide ecosystem wallet and indexer support. Classic assets are compatible with Stellar ecosystem products (such as Stellar wallets) and other ecosystem products (such as exchanges).
```
leighmcculloch: For example for how to avoid the word classic, and telling the reader what we mean instead:
```suggestion
- Wide ecosystem wallet and indexer support. Assets issued by `G...` accounts are compatible with Stellar ecosystem products (such as Stellar wallets) and other ecosystem products (such as exchanges).
```

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
 - Benefit from [anchors](../learn/fundamentals/anchors.mdx), the bridges between the Stellar network and traditional financial systems. Explore the global [Stellar anchor directory](https://anchors.stellar.org/) for further details.
 - Give the issuer granular control over asset management with features that allow the issuer to [name the asset](./control-asset-access.mdx#naming-an-asset), [determine access control](./control-asset-access.mdx#controlling-access-to-an-asset-with-flags), [limit asset supply](./control-asset-access.mdx#limiting-the-supply-of-an-asset), [publish asset information](./publishing-asset-info.mdx), and [ensure compliance](./anatomy-of-an-asset.mdx#compliance).
+- Native operations on Classic Assets emit standardized events through the Stellar Asset Contract, allowing off-chain systems and indexers to observe asset activity in a uniform way. Event semantics are defined at the protocol level according to [CAP-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md).
```
leighmcculloch: Instead of linking to CAP-67 and presenting that it emits standardized events, I think it might be helpful to state that Payment operations (not native) emit the same events that SEP-41 defines. The Stellar Asset Contract is a SEP-41 token.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+### When to use
+
+- Simple payment rails with optional smart contract integration.
+- Trustline-governed asset flows that also need Soroban composability
+
+## 2. SEP-41 Contract Tokens
+
+### What these are
+
+Fully customizable contract tokens native to Soroban that implement the **[SEP-41][sep-41] token standard**, with their own storage and logic. [Contract tokens][ti] issued by a deployed Wasm contract (`C...` addresses).
```
leighmcculloch: > native to Soroban

I think we can omit this.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+  - Vesting
+  - Hooks
+  - Custom transfer logic
+  - Fee structures
+- Contract tokens can be held by both `G…` accounts and `C…` accounts.
+- Balances are always stored in contract data.
+
+### Strengths
+
+- Fully programmable token logic beyond what Classic Assets + SAC can express.
```
leighmcculloch:
```suggestion
- Fully programmable token logic beyond what the Stellar Asset Contract expresses.
```

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- [SEP-41][sep-41] standard ensures broad tooling compatibility in Soroban.
+- Can be integrated into existing Soroban applications and DeFi protocols.
+- Some ecosystem wallet support.
+
+### Tradeoffs
+
+- Higher execution costs compared to native Classic Asset operations.
+- Less widespread ecosystem support compared to Classic Assets.
```
leighmcculloch: Should we be more specific to what the gaps in support are as this is relatively generic statement. Is it primarily exchanges that do not support now?

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- When extending SEP-41 with custom functionality.
 
-The Stellar Asset Contract (SAC) is compiled into the protocol layer and allows smart contracts to interact with assets issued on Stellar. An instance of the SAC can be deployed for every Stellar asset by anyone who wants to interact with the asset from a contract. The SAC has access to all account balances (for XLM) and trustline balances (for all other assets) as well as smart contract token balances.
+## 3. ERC-3643 / SEP-57 (T-REX) Tokens
 
-Read more about the SAC [here][sac].
+### What these are
 
-Learn how to deploy a Stellar Asset Contract for an asset in [this How-To Guide](../tools/cli/cookbook/deploy-stellar-asset-contract.mdx).
+A **permissioned token standard** for regulated, compliance-aware assets that extends [SEP-41][sep-41] with identity and rules enforcement. T-REX tokens are issued by a deployed Wasm contract (`C...` addresses) and implement compliance logic directly in the contract suite.
```
leighmcculloch:
```suggestion
A permissioned token standard for regulated, compliance-aware assets that extends [SEP-41][sep-41] with identity and rules enforcement. T-REX tokens are issued by a deployed Wasm contract (`C...` addresses) and implement compliance logic directly in the contract suite.
```

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+| **Source Code** | [Built-in Stellar Asset Contract](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host/src/builtin_contracts) | [OpenZeppelin Fungible Token reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/fungible) | [OpenZeppelin T-REX (RWA token) reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/rwa) |
+| **Relevant SEPs** | [SEP-0001 (Stellar Info File)][sep-1] <br /><br /> [SEP-0014 (Dynamic Asset Metadata)][sep-14] <br /><br />[SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] <br /><br /> [SEP-0057 (T-REX / Token for Regulated EXchanges)][sep-57] |
+
+_\* By assigning a smart contract (`C...` address) as the owner of a Stellar Asset Contract, balances and transfer rules can be fully managed within the contract. This removes the need for trustlines, as balances are hold in the contract storage._
+
+## Decision Guide
+
+Choose the right token model based on the use case:
+
+- **Classic Stellar Assets (with SAC)**:
```
leighmcculloch:
```suggestion
- **Assets Issued by `G...` Accounts (with SAC)**:
```

docs/tokens/README.mdx
leighmcculloch: Something not discussed is customisation through the admin which is a reasonably important feature to consider because it does provide a level of customisation that is meaningful.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- SEP-41 Contract Tokens are ideal for DeFi and advanced tokenomics requiring full programmability.
+- SEP-57 ERC-3643 Tokens are designed for regulated Real World Assets, offering on-chain compliance and identity with institutional-grade controls.
+
+---
+
+## More Info
+
+### Stellar Asset Contract (SAC)
+
+The Stellar Asset Contract (SAC) is a protocol-level contract that enables smart contracts to interact with assets issued on Stellar.
```
leighmcculloch:
```suggestion
The Stellar Asset Contract (SAC) is a contract built-in to the Stellar protocol that implements SEP-41 for assets issued by Stellar accounts (`G...`), and enables smart contracts to interact with assets issued on Stellar.
```

docs/tokens/README.mdx
janewang: Okay removing "native"

docs/tokens/README.mdx
janewang: Updated.

docs/tokens/README.mdx
janewang: Removing

docs/tokens/README.mdx
janewang: Updated

docs/tokens/README.mdx
janewang: Updated

docs/tokens/README.mdx
janewang: Thanks

docs/tokens/README.mdx
janewang: Thanks

docs/tokens/README.mdx
janewang: I can add exchanges. That is indeed a painpoint.

docs/tokens/README.mdx
```diff
@@ -11,78 +11,242 @@ The term "custom token" has been deprecated in favor of "contract token". View t
...
+- Balances are always stored in contract data entries.
+- Commonly used when custom logic is required:
+  - Mint/burn rules
+  - Vesting
+  - Hooks
+  - Custom transfer logic
+  - Fee structures
+- Contract tokens can be held by both `G…` accounts and `C…` accounts.
+- Balances are always stored in contract data.
+
```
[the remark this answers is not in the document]
janewang: Fixed.

docs/tokens/README.mdx
janewang: Removing the word "Native"

docs/tokens/README.mdx
janewang: Ok make sense

docs/tokens/README.mdx
janewang: This is a great call out. I think it's buried such that it is almost tribal knowledge. Adding..

docs/tokens/README.mdx
janewang: Make sense on the rephrasing.

Also globally removed the word `classic`.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+The term "custom token" has been deprecated in favor of "contract token". View the conversation in the [Stellar Developer Discord](https://discord.com/channels/897514728459468821/966788672164855829/1359276952971640953).
+
+:::
+
+This brief compares the three primary tokenization models available on the Stellar network:
+
+- Stellar Assets (with built-in Stellar Asset Contract),
+- [SEP-41][sep-41] Contract Tokens, and
+- Regulated [SEP-57][sep-57] Tokens (Tokens for Regulated EXchanges / T-REX)
```
leighmcculloch:
```suggestion
- [SEP-57][sep-57] T-REX Tokens for Regulated EXchanges
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Ideal For** | Payments, simple assets (fiat based stablecoins), Soroban integration | DeFi and custom tokenomics | Institutional RWAs with compliance |
+| **Source Code** | [Built-in Stellar Asset Contract](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host/src/builtin_contracts) | [OpenZeppelin Fungible Token reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/fungible) | [OpenZeppelin T-REX (RWA token) reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/rwa) |
+| **Relevant SEPs** | [SEP-0001 (Stellar Info File)][sep-1] <br /><br /> [SEP-0014 (Dynamic Asset Metadata)][sep-14] <br /><br />[SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] <br /><br /> [SEP-0057 (T-REX / Token for Regulated EXchanges)][sep-57] |
+
+_\* By assigning a smart contract (`C...` address) as the owner of a Stellar Asset Contract, balances and transfer rules can be fully managed within the contract. This removes the need for trustlines, as balances are hold in the contract storage._
```
ElliotFriend: I don't think this is true, as written. At least, it the way it reads to me feels inaccurate. This sounds like a `G...` account holding an asset wouldn't need a trustline if the asset's SAC has a `C...` address admin.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+  - Full programmability for innovative use cases
+  - Integration with advanced Soroban applications
+
+- **SEP-57 / ERC-3643 (T-REX) Tokens**:
+  - Onchain compliance and identity verification
+  - Institutional-grade access controls
+  - Regulatory requirements for Real World Assets (RWAs)
+  - Interoperability with Ethereum ERC-3643
+
+::::note Summary: TLDR
```
ElliotFriend:
```suggestion
:::note[Summary: TLDR]
```

docusaurus admonitions titles use `[]` nowadays ([source](https://docusaurus.io/docs/markdown-features/admonitions#specifying-title))

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+- Stellar Assets are best for simple payments and fiat-backed assets, with optional smart contract interoperability via the Stellar Asset Contract.
+- SEP-41 Contract Tokens are ideal for DeFi and advanced tokenomics requiring full programmability.
+- SEP-57 ERC-3643 Tokens are designed for regulated Real World Assets, offering on-chain compliance and identity with institutional-grade controls.
+
+::::
```
ElliotFriend:
```suggestion
:::
```

three colons for these adomintions

docs/tokens/anatomy-of-an-asset.mdx
janewang: This is what @leighmcculloch had suggested

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- [SEP-41][sep-41] Contract Tokens, and
+- Regulated [SEP-57][sep-57] Tokens (Tokens for Regulated EXchanges / T-REX)
+
+to help issuers select the most appropriate model for their use case.
+
+## Tokenization Model Comparison
+
+| Dimension | Stellar Asset (with SAC) | SEP-41 Contract Token | ERC-3643 (T-REX) |
+| --- | --- | --- | --- |
+| **Platform** | Stellar Ledger + Soroban | Soroban on Stellar | Soroban on Stellar |
```
leighmcculloch: Wdyt?
```suggestion
| **Platform** | Stellar Operations + Soroban | Soroban on Stellar | Soroban on Stellar |
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Token Model** | Trustlines + native ops + SEP-41 via SAC | SEP-41 interface implemented by a smart contract, fully extensible | SEP-41 with extensible compliance logic |
```
leighmcculloch:
```suggestion
| **Token Model** | Trustlines + Operations + SEP-41 via SAC | SEP-41 interface implemented by a smart contract, fully extensible | SEP-41 with extensible compliance logic |
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Programmability** | ❌ Protocol-defined / SAC interface not updatable | ✅ Fully customizable | ✅ Fully customizable + compliance rules |
```
leighmcculloch: I think we can shorten this to:
```suggestion
| **Programmability** | ❌ Protocol-defined | ✅ Fully customizable | ✅ Fully customizable + compliance rules |
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Interaction Method** | Native ops (accounts) and contract calls (via SAC) | Contract calls | Contract calls |
+| **Admin Control** | Protocol flags + optional admin via SAC | Custom contract logic | Built-in compliance and rule enforcement |
```
leighmcculloch:
```suggestion
| **Admin Control** | Issuer flags + optional admin via SAC | Custom contract logic | Built-in compliance and rule enforcement |
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Identity Model** | Off-chain identity + optional contract logic | Programmable identity | Programmable identity with on-chain enforcement |
```
leighmcculloch: The off-chain identity piece isn't required if the admin is changed to a contract. So this is less true I think and not really a differentiator. In all cases the identity model can be programmable after initial setup.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Cost & Speed** | Very low (native ops) / Moderate (via SAC) | Moderate | Higher |
+| **Ecosystem** | Stellar payments, DEX, and Soroban | Stellar DeFi & dApps / Interop with other L1s | Institutional DeFi / Interop with other L1s |
+| **Trustline Required** | ✅ Yes\* | ❌ No | ❌ No |
+| **Ledger Storage** | Blockchain ledger (trustlines)\* | Smart contract storage | Smart contract storage |
```
leighmcculloch: The ledger storage is the same, I think the differentiator isn't ledger vs contract storage, because they are essentially the same both are on the same ledger, but instead the costs for the storage. And for rent and archives we should link to docs about them.
```suggestion
| **Ledger Storage Fees/Expiry** | 0.5xlm per trustline, no expiry\* | Rent and Archives | Rent and Archives |
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+## 1. Stellar Assets (with Built-in [Stellar Asset Contract][sac])
+
+### What this category encompasses
+
+All assets issued by Stellar accounts (G…` addresses) are **Stellar Assets**. These assets can be represented in two forms:
+
+- **Stellar Asset** Held in trustlines and transferred via payment operations.
+
+- **Built-in Smart Contract (Stellar Asset Contract)** A [Stellar Asset Contract (SAC)][sac] can optionally be deployed to a contract address for the asset, implementing the SEP-41 token interface so the asset can be used in smart contracts.
```
leighmcculloch:
```suggestion
- **Built-in Smart Contract (Stellar Asset Contract)** A [Stellar Asset Contract (SAC)][sac] can be deployed to a contract address for the asset, implementing the SEP-41 token interface so the asset can be used in smart contracts. Anyone can deploy this contract to enable interacting with the asset in smart contracts.
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- Native operations on Stellar Assets emit standardized events through the Stellar Asset Contract, allowing off-chain systems and indexers to observe asset activity in a uniform way. Event semantics are defined at the protocol level according to [CAP-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md).
+- Operations on Stellar Assets emit the same standardized events defined by SEP-41, because the Stellar Asset Contract implements the SEP-41 token interface, Payment operations on Stellar Assets surface identical transfer events, allowing off-chain systems and indexers to observe asset activity through a uniform way.
+- By default, the admin of a Stellar Asset Contract (SAC) is the asset issuer (a `G...` address). However, the admin can be changed to a smart contract. Assigning a contract as the SAC admin enables custom, onchain policy for administrative actions for privileged operations.
+
+### Tradeoffs
+
+- Accounts must have an active trustline to receive, hold, and transact.
```
leighmcculloch, line 69: This should also be listed under strengths as 'accounts can choose which assets they receive'

leighmcculloch (approved):
Some feedback inline.

docs/tokens/anatomy-of-an-asset.mdx
leighmcculloch: The admin can manage minting. In protocol 26 G accounts in contract land can also manage their own trust lines.

docs/tokens/anatomy-of-an-asset.mdx
janewang: @leighmcculloch Should we delete this comparison for Identity Model?

docs/tokens/anatomy-of-an-asset.mdx
janewang: @leighmcculloch We probably should clarify this is coming in protocol 26.

dmkozh: Protocol 26 does not get rid of the trustlines though, and there is still going to be friction in using them. Trustlines also have nothing to do with the admin, do they? I.e. surely a contract can mint token, but that doesn't allow it the requirement for a G-account receiver to have a trustline.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+- Stellar Assets (with built-in Stellar Asset Contract),
+- [SEP-41][sep-41] Contract Tokens, and
+- [SEP-57][sep-57] T-REX Tokens for Regulated EXchanges
+
+to help issuers select the most appropriate model for their use case.
+
+## Tokenization Model Comparison
+
+| Dimension | Stellar Asset (with SAC) | SEP-41 Contract Token | ERC-3643 (T-REX) |
```
dmkozh, line 23: General note: I know that we're trying to avoid 'Soroban' in the docs for the most part and use just 'smart contracts'/'Stellar smart contracts' etc. I don't have a strong preference here, but just wanted to bring this up for consideration.

docs/tokens/anatomy-of-an-asset.mdx
janewang: Okay deleting this.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- Native operations on Stellar Assets emit standardized events through the Stellar Asset Contract, allowing off-chain systems and indexers to observe asset activity in a uniform way. Event semantics are defined at the protocol level according to [CAP-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md).
+- Operations on Stellar Assets emit the same standardized events defined by SEP-41, because the Stellar Asset Contract implements the SEP-41 token interface, Payment operations on Stellar Assets surface identical transfer events, allowing off-chain systems and indexers to observe asset activity through a uniform way.
+- By default, the admin of a Stellar Asset Contract (SAC) is the asset issuer (a `G...` address). However, the admin can be changed to a smart contract. Assigning a contract as the SAC admin enables custom, onchain policy for administrative actions for privileged operations.
+
+### Tradeoffs
+
+- Accounts must have an active trustline to receive, hold, and transact.
+- Trustlines cannot be added programmatically through contract invocations, limiting support for assets that rely on bridging or chain abstraction patterns.
```
dmkozh: * until protocol 26

Though the most important thing about trustline is really that they require explicit user authorization.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+This allows Stellar assets to interoperate seamlessly with Soroban smart contracts.
+
+- [Read more about the Stellar Asset Contract](./stellar-asset-contract.mdx)
+- [How to deploy a Stellar Asset Contract](../tools/cli/cookbook/deploy-stellar-asset-contract.mdx)
+
+### Benefits of the SAC
+
+**Compatibility**  
+The SAC preserves full compatibility with the existing Stellar asset model, including trustlines, authorization flags, and the Stellar DEX.
```
dmkozh: The Stellar DEX part is kind of misleading, it might be worth specifying that DEX compatibility does not exist for SAC , but exists for the underlying assets via classic operation. That's pretty important, because e.g. C-accounts holding a SAC balance won't have access to DEX.

docs/tokens/anatomy-of-an-asset.mdx
janewang: Good call, okay will globally replace

docs/tokens/anatomy-of-an-asset.mdx
janewang: @dmkozh do you have a write up on what's coming from protocol 26 that we could catch up on?

docs/tokens/anatomy-of-an-asset.mdx
janewang: Should we just not mention DEX?

docs/tokens/anatomy-of-an-asset.mdx
janewang: @dmkozh We have two viable models of managing this asset correct? Using a `C…` address as the admin over SAC balances, or a `G… `address to manage classic asset balances through trustlines. Managing with `C...` address gives you that composability where the admin can be another contract.  Anything else I could add?

docs/tokens/anatomy-of-an-asset.mdx
dmkozh: I think DEX and classic liquidity pools are a valid benefit of classic assets and thus SAC because of liquidity they provide. I just think we could separate these points (i.e. one point is what you have here but without DEX, another point is just additional liquidity thanks to the classic DEX/LPs). But the second point is up to your preference really.

docs/tokens/anatomy-of-an-asset.mdx
dmkozh: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0073.md is a good reference for now.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- [SEP-41][sep-41] Contract Tokens, and
+- [SEP-57][sep-57] T-REX Tokens for Regulated EXchanges
+
+to help issuers select the most appropriate model for their use case.
+
+## Tokenization Model Comparison
+
+| Dimension | Stellar Asset (with SAC) | SEP-41 Contract Token | ERC-3643 (T-REX) |
+| --- | --- | --- | --- |
+| **Platform** | Stellar Operations + Soroban | Soroban on Stellar | Soroban on Stellar |
```
JakeUrban: I'm not sure if this row is worth having. What is "Platform"? In my opinion the distinction being made here is better described in the other rows below.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+This brief compares the three primary tokenization models available on the Stellar network:
+
+- Stellar Assets (with built-in Stellar Asset Contract),
+- [SEP-41][sep-41] Contract Tokens, and
+- [SEP-57][sep-57] T-REX Tokens for Regulated EXchanges
+
+to help issuers select the most appropriate model for their use case.
+
+## Tokenization Model Comparison
```
JakeUrban, line 21: This table implies that SEP-41 and TREX are mutually exclusive, when in reality one is a subset of the other. For example there is no reason why a plain SEP-41 token couldn't have compliance rules embedded into its `transfer` implementation.

Can I suggest instead that the comparison in this document only be between Stellar Assets and SEP-41 Contract tokens? If we do this, we can make it clear that SEP-41 tokens are extensible, whereas Stellar assets are not. Then on a separate page, we can document the advantages of extending a SEP-41 token to support TREX.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+| Dimension | Stellar Asset (with SAC) | SEP-41 Contract Token | ERC-3643 (T-REX) |
+| --- | --- | --- | --- |
+| **Platform** | Stellar Operations + Soroban | Soroban on Stellar | Soroban on Stellar |
+| **Token Model** | Trustlines + Operations + SEP-41 via SAC | SEP-41 interface implemented by a smart contract, fully extensible | SEP-41 with extensible compliance logic |
```
JakeUrban: Similar feedback here, its not clear what "token model" is.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Relevant SEPs** | [SEP-0001 (Stellar Info File)][sep-1] <br /><br /> [SEP-0014 (Dynamic Asset Metadata)][sep-14] <br /><br />[SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] <br /><br /> [SEP-0057 (T-REX / Token for Regulated EXchanges)][sep-57] |
+
+_\* By assigning a smart contract (`C...` address) as the owner of a Stellar Asset Contract, balances and transfer rules can be fully managed within the contract. This removes the need for trustlines, as balances are hold in the contract storage._
+
+## 1. Stellar Assets (with Built-in [Stellar Asset Contract][sac])
+
+### What this category encompasses
+
+All assets issued by Stellar accounts (G…` addresses) are **Stellar Assets**. These assets can be represented in two forms:
```
JakeUrban: Maybe "interacted with" rather than "represented in"? Then the two forms listed are "payment operations" and "the builtin contract"

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+- **Stellar Asset** Held in trustlines and transferred via payment operations.
+
+- **Built-in Smart Contract (Stellar Asset Contract)** A [Stellar Asset Contract (SAC)][sac] can be deployed to a contract address for the asset, implementing the SEP-41 token interface so the asset can be used in smart contracts. Anyone can deploy this contract to enable interacting with the asset in smart contracts.
+
+### How they work
+
+- For protocol-level behavior and trustline semantics, see [Asset Design Considerations](./control-asset-access.mdx).
+- Implemented at the protocol level, the **Stellar Asset Contract (SAC)** provides a smart contract interface that enables Stellar assets to interoperate seamlessly with Stellar smart contracts.
+- Transfers between accounts and contracts using SAC resolve to the same trustline balance updates the Stellar protocol has always used.
```
JakeUrban: There is some nuance here we should probably expand on:

- Transfers to or from a G account results in the same modifications to trustline ledger entries
- Transfers to or from contracts result in updates to the token contract's data entries.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+| **Source Code** | [Built-in Stellar Asset Contract](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host/src/builtin_contracts) | [OpenZeppelin Fungible Token reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/fungible) | [OpenZeppelin T-REX (RWA token) reference](https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/src/rwa) |
+| **Relevant SEPs** | [SEP-0001 (Stellar Info File)][sep-1] <br /><br /> [SEP-0014 (Dynamic Asset Metadata)][sep-14] <br /><br />[SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] | [SEP-41 (Soroban Token Interface)][sep-41] <br /><br /> [SEP-0057 (T-REX / Token for Regulated EXchanges)][sep-57] |
+
+_\* By assigning a smart contract (`C...` address) as the owner of a Stellar Asset Contract, balances and transfer rules can be fully managed within the contract. This removes the need for trustlines, as balances are hold in the contract storage._
+
+## 1. Stellar Assets (with Built-in [Stellar Asset Contract][sac])
```
JakeUrban:
```suggestion
## 1. Stellar Assets (with Built-in [Stellar Asset Contracts][sac])
```

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+### How they work
+
+- For protocol-level behavior and trustline semantics, see [Asset Design Considerations](./control-asset-access.mdx).
+- Implemented at the protocol level, the **Stellar Asset Contract (SAC)** provides a smart contract interface that enables Stellar assets to interoperate seamlessly with Stellar smart contracts.
+- Transfers between accounts and contracts using SAC resolve to the same trustline balance updates the Stellar protocol has always used.
+
+### Strengths
+
+- Has the benefits of Stellar assets (fast, low cost), making these great for remittances and micropayments.
```
JakeUrban: Transfers of stellar assets aren't "faster" than contract tokens, if by fast you mean they settle between the source and destination faster. Them being less computationally expensive is still enough for them to be preferable for the use cases you mentioned though

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- Can be used in Stellar smart contracts via SAC with a [SEP-41][sep-41] interface.
+- Maintains issuer controls and trustline semantics even when used from contracts.
+- Wide ecosystem wallet and indexer support. Assets issued by `G...` accounts are compatible with Stellar ecosystem products (such as Stellar wallets) and other ecosystem products (such as exchanges).
+- Benefit from [anchors](../learn/fundamentals/anchors.mdx), the bridges between the Stellar network and traditional financial systems. Explore the global [Stellar anchor directory](https://anchors.stellar.org/) for further details.
```
JakeUrban: There is nothing that stops anchors / ramps from supporting contract tokens, so this point is more a reiteration of the previous point about ecosystem support.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+- Give the issuer granular control over asset management with features that allow the issuer to [name the asset](./control-asset-access.mdx#naming-an-asset), [determine access control](./control-asset-access.mdx#controlling-access-to-an-asset-with-flags), [limit asset supply](./control-asset-access.mdx#limiting-the-supply-of-an-asset), [publish asset information](./publishing-asset-info.mdx), and [ensure compliance](./anatomy-of-an-asset.mdx#compliance).
```
JakeUrban: Contract tokens can have even more granular control to issuers than stellar assets, so maybe we should rephrase to say that there are "built-in" control features, compared to contract tokens where nothing is built-in, everything must be custom-built.

docs/tokens/anatomy-of-an-asset.mdx
```diff
@@ -1,8 +1,246 @@
...
+
+### Tradeoffs
+
+- Accounts must have an active trustline to receive, hold, and transact.
+- Trustlines cannot be added programmatically through contract invocations, limiting support for assets that rely on bridging or chain abstraction patterns.
+
+### When to use
+
+- Simple payment rails with optional smart contract integration.
+- Trustline-governed asset flows that also need Soroban composability
```
JakeUrban: Agreed that using stellar assets is "simpler", but we should probably mention "cheaper" too. In use cases where transaction fees need to be minimized, stellar assets are definitely preferable.

docs/tokens/anatomy-of-an-asset.mdx
janewang: Updated

docs/tokens/anatomy-of-an-asset.mdx
janewang: SEP-41 and TREX are not subsets. TREX is an extension of SEP-41 with a lot more features added. I think I'd like to keep T-REX here as it is an important standard for tokenization.

docs/tokens/anatomy-of-an-asset.mdx
janewang: @JakeUrban What wording do you suggest? Thanks in advance

docs/tokens/anatomy-of-an-asset.mdx
janewang: I updated to `interacted with`. I think that the first category needs to be an asset, not payment operations.

docs/tokens/anatomy-of-an-asset.mdx
janewang: Okay let's remove

docs/tokens/anatomy-of-an-asset.mdx
janewang: You are correct. Let's change to (low cost, less computationally expensive).

docs/tokens/anatomy-of-an-asset.mdx
janewang: Updated. The only instances where `Soroban` is mentioned is for the original name of SEP-41.

docs/tokens/anatomy-of-an-asset.mdx
janewang: Let's combine into a single point

docs/tokens/anatomy-of-an-asset.mdx
janewang: Thanks, great addition for better clarity!

docs/tokens/anatomy-of-an-asset.mdx
janewang: Great addition

docs/tokens/anatomy-of-an-asset.mdx
janewang: `Token Implementation`

JakeUrban: Maybe `Storage Model`? Stellar assets are stored in trustlines and contract data entries, SEP41 tokens are stored only in contract data entries. Same for TREX.

PR #2231, opened 2026-02-06
🎨 Clean up images, imports

JFWooten4:
does #2230


ElliotFriend:
Good news: #2410 has landed, so the `i18n/` directory is gone from main entirely. That should shrink this PR considerably — could you rebase onto current main (or regenerate the branch) so it contains only the non-i18n image/import cleanup? Happy to review once it's trimmed down.

JFWooten4:
I've adjusted it to remove these references and merged it into main - thanks for the attention here! I'd still rather release it as a draft before duplicative smaller PRs like #2400 and #2380 merge in. Those were some of the stickier adjustments, which I thought would be easier to review independently.

PR #2234, opened 2026-02-07
🖌️ Clean up meta tags

JFWooten4:
this should merge cleanly after #2231

it just polishes up some of the old SEO


JFWooten4:
It'd also be super if this could clean up some of the tangled API endpoints. There are areas with inconsistent and unreadable blobs for filenames. Some paths to start with:

- `docs\data\indexers\build-your-own\ingest-sdk\developer_guide\ledgerbackends\bufferedstoragebackend.mdx`
  - Has underscores in folder, too.
  - I'm thinking `kebab-case`.
    - It's done proper in `galexie\admin_guide\full-history-exporting.mdx`.
- `docs\data\apis\rpc\api-reference\methods`
  - Still uses capitalized filenames which are...
  - It works but is very standalone.
  - Produces a capitalized URL.
- `docs\data\apis\horizon\api-reference\resources\claimablebalances`
  - The worst thing is that `aggregations\order-books` splits instead.
- `data\analytics\analytics-providers\analytics-providers.mdx`
  - Lastly, I think this should just be a README.
  - ✅ #2450

ElliotFriend:
Thanks for the effort here — 1,584 files is no small thing, and that's honestly part of why I'm closing it.

Having sampled the changeset, the bulk of it is: (a) reordering frontmatter keys within pages (no rendering or SEO effect, and since no linter enforces key order, it would drift back immediately), (b) converting reference-style markdown links to inline (a style preference we don't consider an upgrade — reference-style is a deliberate choice in several sections), and (c) ~780 files under `i18n/`, which is about to be removed entirely (see #2410 — the maintained-translations experiment is winding down), so half this PR evaporates regardless.

Set against that, a 1,584-file review isn't a good trade for either of us. If there are pages with genuinely duplicated or redundant meta/head blocks, I'd gladly take that as a small, targeted PR after #2410 lands — a focused diff of real removals will merge fast.

One more thing: the API-reference file-renaming idea you floated in the comments would break URLs and needs redirects, so please raise that as an issue for discussion before any code — it's not a decision to make inside a cleanup PR.

PR #2240, opened 2026-02-11
Update examples in the send/receive from C accounts guide

brunomlr:
With the testnet reset, links to Horizon were broken. Removed the direct link and added a more detailed example of responses in the page content itself.

### DIFF
--- a/docs/build/guides/transactions/send-and-receive-c-accounts.mdx
+++ b/docs/build/guides/transactions/send-and-receive-c-accounts.mdx
@@ -29,24 +29,54 @@ That's the main difference between receiving from customers using G or Contract 
 
 **Receiving from a G address**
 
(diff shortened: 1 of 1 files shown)


docs/build/guides/transactions/send-and-receive-c-accounts.mdx
```diff
@@ -29,24 +29,54 @@ That's the main difference between receiving from customers using G or Contract
 
 **Receiving from a G address**
 
-Using Horizon's `accounts/{acc_id}/transactions` endpoint ([example](https://horizon-testnet.stellar.org/accounts/GBLVHX33XGOBDOXK7ERDL34NVH6WW7VTT2OBAHPJ7G3D423HBG5NOMY7/transactions))
+Using Horizon's `accounts/{acc_id}/transactions` endpoint
```
ElliotFriend: i think it's good to remove the testnet transaction links.  maybe we can replace it with links to horizon's API docs page for this endpoint?

ElliotFriend (commented):
seems like some really useful changes. we could replace the horizon links with links to the horizon api reference in the docs.

what do you think?

ElliotFriend (approved):

PR #2242, opened 2026-02-11
Update contract authorization how-to-guide cross contract auth example

mootz12:
* Fix bug that states non-root auth will error
* Improve context as to why `user.require_auth()` is recommended at the top level of a contract call


docs/build/guides/auth/contract-authorization.mdx
```diff
@@ -252,7 +252,7 @@ To illustrate how to add authorization to a simple cross-contract invocation, we
 
 The invoking contract in the mentioned example contract creates a client for invoking the `add()` function in the addition contract. Let’s say we want the `add()` contract to be authorized by a user. Then, we need to pass the user to the `add()` contract function from the invoking function.
 
-Even though the invoking function may not require the user’s authorization, the `add()` function will panic if it encounters authorization not tied to the root/top contract invocation. Therefore, we need to use `require_auth()` on the user in the invoking function.
+Even though the invoking function may not require the user’s authorization, it's recommended to authorize the user at the entry point. This way, the authorization for `add()` is correctly recorded as a sub-invocation of `add_with()`, binding the authorization of `add()` to the top level `add_with()` call. Therefore, we need to use `require_auth()` on the user in the invoking function.
```
dmkozh: I think we can be more specific here, something along the lines of

"...it's recommended to authorize the user at the entry point. Without that, the authorized inner `add` call can be front-run by anyone without being wrapped in `add_with`. Adding `require_auth` at the entry point ensures that all the inner contract calls that are authorized on behalf of `user` will be executed atomically together with the entry point call."

mootz12: Agreed, this is clearer.

docs/build/guides/auth/contract-authorization.mdx
```diff
@@ -252,7 +252,7 @@ To illustrate how to add authorization to a simple cross-contract invocation, we
 
 The invoking contract in the mentioned example contract creates a client for invoking the `add()` function in the addition contract. Let’s say we want the `add()` contract to be authorized by a user. Then, we need to pass the user to the `add()` contract function from the invoking function.
 
-Even though the invoking function may not require the user’s authorization, the `add()` function will panic if it encounters authorization not tied to the root/top contract invocation. Therefore, we need to use `require_auth()` on the user in the invoking function.
+Even though the invoking function may not require the user’s authorization, it's recommended to authorize the user at the entry point. Without that, the authorized inner `add()` call can be front-run by anyone without being wrapped in `add_with()`. Adding `require_auth()` at the entry point ensures that all the inner contract calls that are authorized on behalf of user will be executed atomically together with the entry point call. Therefore, we need to use `require_auth()` on the user in the invoking function.
```
fnando:
```suggestion
Even though the invoking function may not require the user’s authorization, it's recommended to authorize the user at the entry point. Without that, the authorized inner `add()` call can be front-run by anyone without being wrapped in `add_with()`. Adding `require_auth()` at the entry point ensures that all the inner contract calls that are authorized on behalf of the user will be executed atomically together with the entry point call. Therefore, we need to use `require_auth()` on the user in the invoking function.
```

dmkozh: nit: I think we should use `user` instead of just 'user' here to make it clear that we're talking about the concrete variable and not some abstract contract user.

briwylde08 (approved):
Thank you!!

Issue #2210, opened 2026-01-26
Create a dropdown menu for sub guides in Transactions category


jeesunikim:
Where: [Build > How-To Guides](https://developers.stellar.org/docs/build/guides/transactions)

`Transactions` page has 15 guides within this category. That should be reflected in the menu on the left like Build Applications page does ([Build a Wallet with the Wallet Sdk](https://developers.stellar.org/docs/build/apps/wallet) as an example):

<img width="1200" height="859" alt="Image" src="https://github.com/user-attachments/assets/eff5e644-abb1-4ef9-b4e9-d910fabd4c7f" />

PR #2279, opened 2026-02-24
How-To Guide Dropdowns

briwylde08:
Addresses: https://github.com/stellar/stellar-docs/issues/2210


docs/build/guides/archival/_category_.json
```diff
@@ -1,3 +1,4 @@
 {
-    "collapsible": false
+    "collapsible": true,
+    "collapsed": true
```
ElliotFriend: these fields both default to `true`. in this pr, we're better off removing the `_category_.json` files entirely

PR #2290, opened 2026-03-03
fix: correct pagination limits for getLedgers and getTransactions in OpenRPC spec

miguelnietoa:
The pagination limit descriptions for `getLedgers` and `getTransactions` reference a maximum of `10,000`, but the actual defaults in https://github.com/stellar/stellar-rpc/blob/1a9a97b3f533f91c509661094e71b269d33b735a/cmd/stellar-rpc/internal/config/options.go are:                                                                      
  - `getEvents`: max 10,000, default 100 (correct in docs)                                                                                                                
  - `getLedgers`: max 200, default 50                                                                                                                                     
  - `getTransactions`: max 200, default 50

Confirmed empirically against both testnet (`soroban-testnet.stellar.org`) and mainnet (`soroban-rpc.mainnet.stellar.gateway.fm`) - requesting limit=201 returns "limit must not exceed 200".


miguelnietoa:
Hi @ElliotFriend 👋🏻 
Thanks for pointing me the right direction. I updated the source files instead of editing the generated output directly. However, I replaced the shared `Pagination` schema/contentDescriptor with method-specific ones (`PaginationEvents`, `PaginationLedgers`, `PaginationTransactions`) since each method has different limits.

Lmk if it needs further changes!

miguelnietoa:
Hey @ElliotFriend, good call! I just pushed the changes

Issue #2228, opened 2026-02-05
Add how-to guide for storage data migration


leighmcculloch:
### What problem does your feature solve?

When contracts upgrade and their data structures change (e.g., adding new fields), developers need clear guidance on how to handle existing stored data. The intuitive approaches don't work as expected, and developers often discover this the hard way.

**Intuitive approach 1** - Just read with the new type:
```rust
#[contracttype]
pub struct DataV1 { a: i64, b: i64 }

#[contracttype]
pub struct DataV2 { a: i64, b: i64, c: Option<i64> }

// Developer expects c = None for old entries
let data: DataV2 = env.storage().persistent().get(&key).unwrap();
// TRAPS with Error(Object, UnexpectedSize)
```

**Intuitive approach 2** - Use try_from_val fallback:
```rust
let raw: Val = env.storage().persistent().get(&key).unwrap();
if let Ok(v2) = DataV2::try_from_val(&env, &raw) {
    v2
} else {
    // Never reached - host traps before returning Err
    let v1 = DataV1::try_from_val(&env, &raw).unwrap();
    DataV2 { a: v1.a, b: v1.b, c: None }
}
```

Both approaches trap at the host level before the SDK can handle the mismatch. There is no documentation explaining this behavior or the correct pattern to use.

### What would you like to see?

A new how-to guide added to [https://developers.stellar.org/docs/build/guides/storage](https://developers.stellar.org/docs/build/guides/storage) explaining storage data migration using the version marker pattern:

```rust
#[contracttype]
pub struct DataV1 { a: i64, b: i64 }

#[contracttype]
pub struct DataV2 { a: i64, b: i64, c: Option<i64> }

#[contracttype]
pub enum DataKey {
    DataVersion(u32),  // version marker keyed by id
    Data(u32),         // data keyed by id
}

fn read_data(env: &Env, id: u32) -> DataV2 {
    let version: u32 = env.storage().persistent()
        .get(&DataKey::DataVersion(id))
        .unwrap_or(1);  // default to v1 for entries without version marker
    
    match version {
        1 => {
            let v1: DataV1 = env.storage().persistent().get(&DataKey::Data(id)).unwrap();
            DataV2 { a: v1.a, b: v1.b, c: None }
        }
        _ => env.storage().persistent().get(&DataKey::Data(id)).unwrap(),
    }
}

fn write_data(env: &Env, id: u32, data: &DataV2) {
[...]

PR #2299, opened 2026-03-06
Add how-to guide for storage data migration (#2228)

carstenjacobsen:
PR for issue #2228


docs/build/guides/storage/migrate-contract-storage.mdx
```diff
@@ -0,0 +1,330 @@
...
+fn test_reads_v1_entry_as_v2() {
+    let env = Env::default();
+    let id: u32 = 42;
+
+    // Simulate what the old contract wrote: a DataV1 payload,
+    // no DataVersion entry (old contracts did not write one).
+    let v1_data = DataV1 { a: 10, b: 20 };
+    env.storage().persistent().set(&DataKey::Data(id), &v1_data);
+
+    let result = read_data(&env, id);
```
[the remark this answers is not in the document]
carstenjacobsen: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/stellar/stellar-docs/pull/2299#discussion_r2895656503)

docs/build/guides/storage/migrate-contract-storage.mdx
```diff
@@ -0,0 +1,342 @@
...
+} else {
+    // This branch is never reached — the host traps before returning Err.
+    let v1 = DataV1::try_from_val(&env, &raw).unwrap();
+    DataV2 { a: v1.a, b: v1.b, c: None }
+}
+```
+
+This also traps at the host level. The field count validation happens in the host environment during deserialization — it does not produce a Rust `Err` that the SDK can intercept. There is no way to catch or recover from the mismatch at the contract level.
+
+The root issue is that a contract cannot determine which type an existing storage entry was written as just by reading it. That information must be stored explicitly.
```
leighmcculloch: On a first read, I missed that the first two approaches are what not to do and I got confused. So I think it might be helpful if this how-to guide leads with what someone should do, and then we could have some notes in those little boxes saying here's an approach that someone might think to use but don't do it, it won't work. I would put them further down in the document rather than leading with them.

docs/build/guides/storage/migrate-contract-storage.mdx
```diff
@@ -0,0 +1,342 @@
...
+}
+```
+
+The three test cases cover the three states a record can be in after an upgrade:
+
+- A `DataV1` entry with no version marker (pre-versioning era records)
+- A `DataV2` entry written by the new contract
+- A `DataV1` entry that is read and then written back (the lazy migration round-trip)
+
+## Versioned Enum Pattern
```
leighmcculloch: I'd lead with this approach, and then list the other approaches as alternatives, and then finally have a section of approaches you think will work but won't.

leighmcculloch (commented):
Content is great, but I think we should reorder the document so it leads with the preferred approach, then lists alternative approaches, then finally lists those approaches that you think will work but won't which are currently the first thing a reader sees.

ElliotFriend:
@leighmcculloch i've tried to incorporate the feedback you left. how does everything look now?

briwylde08 (approved):
🥇

PR #2321, opened 2026-03-23
Add Alchemy as a new RPC provider

sydneynotthecity:
Alchemy is live with their RPC node service. Adding to the provider list

### DIFF
--- a/docs/data/apis/rpc/providers.mdx
+++ b/docs/data/apis/rpc/providers.mdx
@@ -22,7 +22,7 @@ These providers allow access to the Futurenet, Testnet and Mainnet network.
 | [Lightsail Network - Quasar\*](https://quasar.lightsail.network/) | ❌ | ❌ | ✅ | ❌ | ✅ |
 | [Uniblock](https://www.uniblock.dev/) | ❌ | ✅ | ✅ | ❌ | ❌ |
 | [Exaion](https://crypto.exaion.com) | ❌ | ❌ | ✅ | ✅ | ✅ |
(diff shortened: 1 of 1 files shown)


docs/data/apis/rpc/providers.mdx
```diff
@@ -21,7 +21,8 @@ These providers allow access to the Futurenet, Testnet and Mainnet network.
 | [OnFinality\*](https://onfinality.io/networks/stellar) | ❌ | ❌ | ✅ | ✅ | ✅ |
 | [Lightsail Network - Quasar\*](https://quasar.lightsail.network/) | ❌ | ❌ | ✅ | ❌ | ✅ |
 | [Uniblock](https://www.uniblock.dev/) | ❌ | ✅ | ✅ | ❌ | ❌ |
-| [Exaion](https://crypto.exaion.com) | ❌ | ❌ | ✅ | ✅ |
+| [Exaion](https://crypto.exaion.com) | ❌ | ❌ | ✅ | ✅ | ✅ |
+| [Alchemy](https://dashboard.alchemy.com/?utm_source=chain_partner&utm_medium=referral&utm_campaign=stellar) | ❌ | ✅ | ✅ | ✅ | ❌ |
```
janewang, line 25: @sydneynotthecity should we have added the utm codes?

sydneynotthecity: @janewang this url was built for us by alchemy so that we can track sign ups from our website (context [here](https://stellarfoundation.slack.com/archives/C0A6UM476BX/p1774299007540319?thread_ts=1774297811.605339&cid=C0A6UM476BX)). my preference is to leave them in, but lmk if you think there is a better spot in our docs for this UTM

PR #2358, opened 2026-04-13
Update the description for 'create account'

jeesunikim:
I found the original text `"New account address that is created and funded"` ambiguous. It sounded like 'an address that is already created and funded.' 

Updated the description to `"The account address to be created and funded by this operation"`


briwylde08 (approved):
👌

PR #2363, opened 2026-04-16
Use custom audio embed screen in existing meeting files

JFWooten4:
In a similar spirit to #2361 and #2362, this new script drastically simplifies the page imports for old audio archives. Thankfully, we've migrated to YouTube for new content, which has the added benefit of automatic, free transcription. But these old ones still carry significant protocol developments, captions or no.

I was quite uninspired by the original default audio element. It works fine even with the messy imports, but I decided to throw in a little style with my rendition. It colors and rounds the element in accordance with the style unification principles apparent in the [brand kit](https://stellar.org/brand-resources).

I only used `maxWidth` so that we can set a reasonable size while maintaining mobile compatibility. If we try setting height too, it can render strangely on small devices.


briwylde08:
@JFWooten4 Can we adjust so this isn't touching the Spanish pages?
Same for: https://github.com/stellar/stellar-docs/pull/2361
Same for: https://github.com/stellar/stellar-docs/pull/2362

JFWooten4:
Sure all done

ElliotFriend:
nice one!! the component is a nice addition to the existing embeds, and the swaps look good.

One change before merge: the component hardcodes blue rgba values (`#3B82F6`) and an emoji label. Could you switch the colors to the theme's `--ifm-*` CSS variables so it respects dark mode and stays consistent if the brand palette changes? Happy to merge once that's in.

JFWooten4:
Interesting thought, “if the brand palette changes”—didn’t even consider that.

ElliotFriend (approved):
LGTM!

PR #2364, opened 2026-04-16
Move AI docs to prominent position in Build

briwylde08:


ElliotFriend (commented):
great change!! only thing is it probably needs a redirect. something like this should do it

```
rewrite ^/docs/tools/developer-tools/building-with-ai(.*)$ "/docs/build/building-with-ai$1" permanent;
```

oceans404:
excited that these AI docs will be more visible! And yep, I agree with Elliot. Please make sure to update `nginx/includes/redirects.conf` with your redirect

briwylde08:
I always forget about redirects 😅

ElliotFriend (approved):

PR #2374, opened 2026-04-22
🛠️ Stylize developer meetings sidebar

JFWooten4:
Right now, the sidebar listing past meetings uses plaintext ISO-formatted `YYYY-MM-DD` format:

[
](https://developers.stellar.org/meetings)

There are a few problems with this approach, and I will admit that the main note inspiring these changes is that I do not reliably remember the number for each month of the year. Thus, the first thing I did was standardize the dates in `MMMM D` format, maintaining the ISO globalization while keeping it a little easier to parse.

This alone would inadequately separate the discussions by year, however. That's why I placed greater emphasis on the existing year sections by expanding them into foldouts. This adds weight with a parallel icon for each year.

That said, one of the biggest challenges I had was the readability. Mobile aside, a plain list like this will grow to be a nightmare to scroll as we continue having meetings in the coming years. So I'll implement a hover-style foldout, where the page reacts to the cursor and expands meetings by year.[^1]

[^1]: On mobile you can just click on the section header like a hamburger.

Lastly, I added colors to stylize this along with the site brand. They are inherited from global values, so there's nothing to maintain specifically in these themes. The rest of the CSS is just text styling to keep it displaying properly on the page.


ElliotFriend:
Thanks for the work here — the sidebar polish is thoughtful, and I can see the care that went into matching the existing look.

I'm closing this one, and I want to be transparent about why, because it's policy rather than code quality: this swizzles the full `BlogSidebar` (Content/Desktop/Mobile) and imports from `@docusaurus/theme-common/internal`, which copies Docusaurus internals into our tree. We already maintain five swizzled components, and each one has to be re-verified against upstream on every Docusaurus upgrade — I'm actively trying to shrink that surface, not grow it. Ejected internals are the most expensive kind.

If you think the sidebar UX itself (e.g., collapsing older years by default) is worth changing, I'd genuinely welcome an issue describing the behavior first — if we agree on the design, we can then figure out the cheapest implementation together. Appreciate you, and sorry to close work you put real time into.

PR #36, opened 2022-09-16
First draft of README file

ElliotFriend:
Took some inspiration from the `stellar/new-docs` repo, and some other README files around github. I think this is a pretty decent start. Any other thoughts?

PR #2375, opened 2026-04-22
🖼️ Some simple little image org

JFWooten4:
This is a setup to get everything in one place for #2231 as I piece it out.


briwylde08:
Where is that Testnet reset notice in the docs? Testnet is now only resetting twice per year.

JFWooten4:
Okay, first of all, thank you for the comment there. I don't know why Copilot didn't notice, but in looking over the files again, I realized I hadn't deleted the original ones from #36. That PR was back in 2022, so definitely will update the notice in a subsequent branch.

That particular alert image was removed in https://github.com/stellar/stellar-docs/commit/c731e708c562b80911045ebfff29ac729daddc35 with the deprecation of the alert element. I've kept it in the src based on a precedent of maintained [unused images](https://github.com/stellar/stellar-docs/tree/4918d8231e31846ef44122d90f3b9b55aff332be/static/assets/sdp/unused) in #2380.

briwylde08 (approved):

PR #2379, opened 2026-04-23
🖼️ Reference images as files, not variables

JFWooten4:
In a similar spirit to normalizaiotn from #2378, this PR removes some wierd parts that imported images from the asset folder like a package rather than just referencing them.


briwylde08:
@JFWooten4 same for this one. We shouldn't be editing the Spanish pages.

JFWooten4:
Okay, but I'm adding another [Eng page](http://localhost:3000/docs/data/analytics/hubble/analyst-guide/creating-visualizations) in exchange.

ElliotFriend:
Good cleanup, but one bug: the new src points to `/img/hubble/hubble_visualization_demo.gif`, but the actual file in the repo is `hubble_visualizatoin_demo.gif` (long-standing typo in the filename). Since JSX `` strings aren't validated at build time, this ships a broken image.

since we're here, let's just go ahead and rename the file to the corrected spelling and keep your `src` as-is (fixes the typo for good).

ElliotFriend (approved):

PR #2384, opened 2026-04-23
🔗 Add Horizon streaming mode cross-references

JFWooten4:
basis extoled in #2382 and #2383

These descriptions already reference streaming behavior, but they were not consistently linked to the actual streaming docs. Adding direct links makes the API pages easier to navigate and gives readers immediate context on how Horizon streaming works.


ElliotFriend:
hey, @JFWooten4!! i think this is could be a useful addition. before i review it, can you merge the existing changes from `main` and re-do the API pages rebuild. we've moved to `pnpm` since this PR and we've had an upgrade of that particular plugin. so, just want to make sure that we're building against the current packages

JFWooten4:
You bet

ElliotFriend:
Some thoughts about this:

1. these endpoints already make use of an `x-supports-streaming: true` extension, which we call out and render in our swizzled component [`/src/theme/ApiExplorer/index.tsx`](https://github.com/stellar/stellar-docs/blob/main/src/theme/ApiExplorer/index.tsx)
2. since these links are using the full `https://developers.stellar.org...` URL (which they SHOULD in an API spec), the link opens in a new tab. that's just kindof "part of doing business" with these spec-generated pages.
3. these pages would now have two links to the same page about streaming on them. see the screenshot below. the one in the box doesn't open in a new tab/window, if that's important.

given all that, do we want this change still? i could go either way, but just wanted to ask the question. for LLM readers, it probably makes more sense and adds better context to have it more "in line" like that.

Screenshot 2026-07-15 at 12 42 14 PM

JFWooten4:
I agree that it probably makes more sense for LLM readers to have it inline, but I certainly see the point you're making with the duplication. I never like repeating myself, and I can't stand it when authors keep referencing the same link like the reader is too stupid to explore it if they'd like on the first mention. That said, I think it's a little outside the scope of this PR.

Perhaps the best course of action is to merge this, which just adds specificity to the text on existing pages. Then see what projects import the API docs, and whether or not their specs use `x-supports-streaming`. I'd also want to affirm the long-term intent of swizzling `ApiExplorer`.

If downstreams don't have that note box, then it makes a lot of sense. If they do (or otherwise implicate streaming), then I think it's a large decision to reword. The whole page could honestly use a better start in cases like this, whereby streaming gets moved closer toward the end as additive rather than explanatory content.

Nice drawing, and yes I'm still for the changes. They could even be expanded to swizzle-fix the new tab:


```ts
export default function MDXLink(props: Props) {
  const { href } = props;

  const isCanonicalDocsLink =
    href === "https://developers.stellar.org" ||
    href?.startsWith("https://developers.stellar.org/");

  return (
    <OriginalMDXLink
      {...props}
      target={isCanonicalDocsLink ? "_self" : props.target}
    />
  );
}
```

docs/platforms/stellar-disbursement-platform/api-reference/request-challenge-transaction.api.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 id: request-challenge-transaction
 title: "Request Challenge Transaction"
-description: "Allows the wallet to get the SEP-10 challenge transaction to be signed. More information [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#challenge). The endpoint url can change but is provided in the Get Stellar toml endpoint response under the `WEB_AUTH_ENDPOINT` variable."
+description: "Allows the wallet to get the SEP-10 challenge transaction to be signed. More information here. The endpoint url can change but is provided in the Get Stellar toml endpoint response under the `WEB_AUTH_ENDPOINT` variable."
```
leighmcculloch: Regenerating the API docs stripped the markdown hyperlink out of the frontmatter `description` field on this and two other unrelated Stellar Disbursement Platform pages, turning `[here](https://...)` into plain unlinked text "here".

On `main`, this line reads `...More information [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#challenge)...`; after this PR the URL is gone entirely. The same regression hits `provide-signed-challenge-transaction.api.mdx` (`[here](...)` → `here`) and `upload-disbursement-instructions.api.mdx` (`[public/resources/disbursementTemplates](...)` → `public/resources/disbursementTemplates`).

These pages have nothing to do with Horizon streaming, and the PR's own `openapi/stellar-disbursement-platform` source is untouched — this looks like a side effect of running the full `pnpm api:gen` (docusaurus gen-api-docs all) generator rather than scoping it to Horizon, and it ships a real content/SEO regression on unrelated pages.

---

docs/platforms/stellar-disbursement-platform/api-reference/_category_.json
```diff
@@ -1,3 +0,0 @@
-{
```
leighmcculloch: This PR deletes the SDP api-reference sidebar's `_category_.json`, which `CONTRIBUTING.md` explicitly says exists only to override the sidebar's `collapsed` state (something `README.mdx` frontmatter can't express).

The file contained `{"collapsed": false}` on `main` and is fully removed here with no corresponding source change requested by the PR (title: "Add Horizon streaming mode cross-references"). Once merged, the Stellar Disbursement Platform API reference category loses its explicit `collapsed: false` override and reverts to whatever Docusaurus' default folder behavior is — changing sidebar UX for a section unrelated to this PR's stated purpose. `CONTRIBUTING.md` states `_category_.json` should be added "only when something genuinely can't be expressed in README.mdx — e.g., overriding collapsible/collapsed."

This looks like unintended fallout from running the full API doc generator rather than an intentional change.

---

docs/data/apis/horizon/api-reference/get-offer-by-offer-id.api.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 id: get-offer-by-offer-id
 title: "Retrieve an Offer"
-description: "The single offer endpoint provides information on a specific [offer](https://developers.stellar.org/docs/learn/fundamentals/liquidity-on-stellar-sdex-liquidity-pools#orders)."
+description: "The single offer endpoint provides information on a specific offer."
```
leighmcculloch: This is one of ~8 generated `.api.mdx` files (also list-all-assets, list-all-claimable-balances, list-strict-receive/send-payment-paths, list-trade-aggregations, retrieve-a-ledger, submit-async-transaction) that get re-emitted with content changes unrelated to the PR's stated goal of adding Horizon streaming cross-references — here the markdown link in the description was stripped out entirely (`[offer](https://developers.stellar.org/...)` → plain `offer`), which is the same generator-driven regression as flagged on the SDP pages, just within Horizon docs this time.

Combined with several other files that only get a re-encoded (but content-equivalent) base64 `api:` blob, this bloats the diff (45 files changed) well beyond what a small copy addition needs. Scoping regeneration to just the Horizon spec (or hand-editing only the touched `.api.mdx` files) would avoid this noise and the link-stripping regressions.

---

kaankacar:

Reviewing because a review was requested by @ElliotFriend.

Decision needed: do we want a second link to the streaming page in the prose, when `src/theme/ApiExplorer/index.tsx` already renders a same-tab "Supports Streaming" tip linking that page on all 25 flagged endpoints? @kaankacar @ElliotFriend

The link format itself checks out. The target returns 200, and full `https://developers.stellar.org` URLs in spec files match your guidance in #2386. `build / build` and `mdx-format` are green at `ffbeef9`.

Three unrelated changes should come out either way:

- `9394aa3` strips markdown links from 11 meta descriptions (8 Horizon, 3 SDP) that main still has. The generator writes those links, so the next `pnpm api` puts them back.
- `c370976` deletes `docs/platforms/stellar-disbursement-platform/api-reference/_category_.json`. That file is not generated. #2791 regenerated the whole SDP reference on main and kept it, and anchor-platform has one too. Deleting it collapses the SDP sidebar category.
- The 3 SDP page edits are now stale. #2791 rewrote that directory on main, and this branch is 43 commits behind.

PR #2389, opened 2026-04-24
Add CAP-77 note for Soroban settings upgrades.

dmkozh:
CAP-77 setting upgrade process is mostly the same as for any other setting, the only caveat is hex-encoding, for which we provide a helper script.

### DIFF
--- a/docs/validators/admin-guide/soroban-settings.mdx
+++ b/docs/validators/admin-guide/soroban-settings.mdx
@@ -15,6 +15,32 @@ If you are being asked to vote for an upgrade, please move on to the [Examine a 
 
 :::
 
+### Frozen ledger keys
+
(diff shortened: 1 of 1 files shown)


anupsdf (approved):

anupsdf:
One of the first use case would be to start with `Strkey` (G-addresses) and get the `LedgerKey` base64 XDR for corresponding account and then use the script to convert that to hex. Should we also add instructions on how to get base64 `LedgerKey` XDR from  `StrKey`?

dmkozh:
> One of the first use case would be to start with Strkey (G-addresses) and get the LedgerKey base64 XDR for corresponding account and then use the script to convert that to hex. Should we also add instructions on how to get base64 LedgerKey XDR from StrKey?

Not sure if we need to get into the weeds of it TBH. I'd imagine the precise methodology depends on the circumstances and different tools may be used. E.g. a simple manual process is to use https://lab.stellar.org/endpoints/rpc/get-ledger-entries (there is no need to even execute a request to get a key).

PR #2390, opened 2026-04-24
📂 Change base dir from `meeting-notes` to just `meetings`

JFWooten4:
After #2375 for the contributior page guide, very other root folder and the vast majority of subfolders use a single-word name. It looks like the meetings held onto a legacy version, and this just updates that with a fix.

It moves the files and configures the build, as Copilot will surely parrot.

I think this makes a whole lot more sense, given the notes are really a tiny proportion of these pages, which focus on the actual meetings. And of course the production slug itself also just uses `meetings`. Just makes diffs a whole lot cleaner to align the file names like that too.

> [!WARNING]
> Affects 2388


kaankacar (commented):
Thanks for the cleanup! Agree on the convention.

One thing on the [nginx rewrites](https://github.com/stellar/stellar-docs/pull/2390/files#diff-19b1bea8e44075edb3c2dba76b95c3e28c463ba2dec0e1ea4dc3f8a4d4d0e3a4):

`routeBasePath: 'meetings'` has been set since the blog was first introduced — see [`8abb6525`](https://github.com/stellar/stellar-docs/commit/8abb6525)([#449](https://github.com/stellar/stellar-docs/pull/449), April 2024). The [`path`](https://github.com/stellar/stellar-docs/blob/main/docusaurus.config.ts#L103) and [`routeBasePath`](https://github.com/stellar/stellar-docs/blob/main/docusaurus.config.ts#L109) have just been mismatched on disk all along; the URL has always been `/meetings/...` publicly. So `/meeting-notes/...` was never a live URL. There's nothing for those redirects to catch.

Unless I'm missing a case where one of those paths actually resolved, I'd drop both rules rather than land permanent rewrites for paths that never existed.

Otherwise LGTM!

motheramyisback-a11y (approved):

JFWooten4:
Thank you for the comment. It reminded me that I missed an edge case from the original branch. Fixed in https://github.com/stellar/stellar-docs/pull/2390/changes/c6828e517fd9b4d2faf4e031a182a55bd6d9e354

As for the two nginx fixes mentioned:
1. The first fix redirects slugs of the form `^/meeting-notes/2020-10-30` to `^/meetings/2020/10/30`
   - I added this forward because there are multiple pages on the official Stellar blog which use this form.
     - OK, well, "blog" is debatable, but the series of community event posts for the old protocol discussions.
     - Notwithstanding if they show up in the post list or RSS feed, those pages still exist on Google searches.
     - Some of them use this old reference structure, although I can only state that from memory when I write this back in the very beginning of February, as I've lost the links to time.
   - You can see this [referenced in the Discord](https://discord.com/channels/897514728459468821/1336645145453531156/1353856643112239105) by old Stella training data.
     - [Another example](https://discord.com/channels/897514728459468821/1349853938366808084/1349853996281630833)
     - [Another example](https://discord.com/channels/897514728459468821/1336022723683422281/1336061223724253245)
     - [Another example](https://discord.com/channels/897514728459468821/1331684414932910151/1333151445985529866)
     - [Another example](https://discord.com/channels/897514728459468821/1328549991597346826/1328573690903920640)
     - [Another example](https://discord.com/channels/897514728459468821/1320449705167290378/1320456802076201103)
     - [Another example](https://discord.com/channels/897514728459468821/1315963117326438492/1316049637034954794)
     - ...there are about 100 [of these](https://discord.com/channels/897514728459468821/1265827786111586344/1266153878588620891)

2. Given the existence of the above mislinks, which does not even begin to explore the web inconsistencies across places like Reddit, I thought the second redirect would be a good safety measure to save a little reader confusion, since the meetings list page is so hard to find from its sole location in the footer.

I do agree that the URL itself in this repo has always been meetings, as far back as the official [JS configuration](https://github.com/stellar/stellar-docs/blob/c4182fbbe120c43b47b7174977f45ef5bbd5347d/docusaurus.config.js) ends. However, it is my understanding of the default behavior before this definition and in the `old-docs` import that before the defined method it just inherited the folder-structure name. And thus that's why I think we have these historical references in official and community media which use the wrong base path.

motheramyisback-a11y (approved):

motheramyisback-a11y (approved):

ElliotFriend:
I want to land this. thanks for the cleanup! the production slug is already `/meetings`, so the rename makes sense. i think there was a previous version of docusaurus that did put in links like `meeting-notes/...`, so the redirect does seem to fit. But sequencing is gonna matter here:

1. It's conflicting with `main` now, and the rename list stops at 2026-05-07 — main has meeting files past that date.
2. i'm trying merge the embed-component PRs (#2361–#2363) first, and rebasing a 75-file rename across those would be painful in the wrong order.

once those three are in, you might be better-off recrating this PR fresh from current main (probably easier than rebasing).

you can ignore the crowdin file, too. that's another holdover from the old way of translating things.

JFWooten4:
It looks like 2361 and 2363 were successfully merged. I've synced this to `main` and found no lingering `iframe` embeds to replace. And I've removed the changes to `crowdin.yaml`.

nginx/includes/redirects.conf
```diff
@@ -117,6 +117,10 @@ rewrite ^/docs/learn/interactive/dapps "/docs/learn/interactive" permanent;
 rewrite ^/docs/learn/networks$ "/docs/networks" permanent;
 rewrite ^/docs/learn/fundamentals/networks$ "/docs/networks" permanent;
 
+# Meetings: old slug formats
+rewrite ^/meeting-notes/([0-9]{4})-([0-9]{2})-([0-9]{2})/?$ "/meetings/$1/$2/$3" permanent;
```
ElliotFriend, line 121: this one's kinda tricky... if there's a link out there in the wild that includes the specific title hash on the page (like `/meeting-notes/1999-01-01#some-anchor-link-thing`), this might not get caught because of the optional trailing slash followed directly by the end of the string. we might want to adjust this regex accordingly.

src/pages/docs/notes.tsx
ElliotFriend, line 1: i'm not sure i understand what this page is meant to be doing. are there links you've seen that would land at `/docs/notes#1999-01-01`?

nginx/includes/redirects.conf
ElliotFriend: scratch that! today i learned the `#anchor` parts of URLs never actually reach the server... 🤷🏻‍♂️ calls into question some of the other redirect things i've tried to handle here in this file lol 🤦🏻‍♂️ 

in other news, the regex looks good to me!

ElliotFriend (commented):
this is looking pretty good to me! i'm not really sure about the `src/pages/docs/notes.tsx` addition, though. can you provide some more details on that?

src/pages/docs/notes.tsx
JFWooten4: These paths are holdovers from the old `soroban.stellar.org` site. In my initial approach, I mistakenly interpreted that they both used the same host, and so I made changes based on the old `soroban.stellar.org/docs/notes#2024-02-09` structure referenced in external materials. However, this was incorrect.

The old site's independent host automatically redirects to `https://developers.stellar.org/meetings#2024-02-09` (sticking with this as _one_ example). However, `meetings` does not 404, and hence it does not hit the Nginx redirect. Thus, we need a way to catch `#1999-01-01` on the index page for legacy Soroban links like in https://github.com/stellar/stellar-docs/pull/2390/commits/c6828e517fd9b4d2faf4e031a182a55bd6d9e354.

A naive approach swizzles `theme/BlogListPage/index.tsx` into something like:

```tsx
...

const SOROBAN_MEETING_DATE = /^#\d{4}-\d{2}-\d{2}$/;

function findSorobanMeetingPermalink(
  hash: string,
  sidebar: Props["sidebar"],
): string | null {
  if (!SOROBAN_MEETING_DATE.test(hash)) {
    return null;
  }

  const meetingDate = hash.slice(1);

  return (
    sidebar?.items.find((item) => item.title === meetingDate)?.permalink ?? null
  );
}

export default function BlogListPageWrapper(props: Props): ReactNode {
  
  ...

    const permalink = findSorobanMeetingPermalink(window.location.hash, sidebar);

    if (permalink) {
      history.replace(permalink);
    }
  }
  
  ...

}
```


I've taken a more elegant approach in 9b008b15, which introduces our first ClientModule. It runs a check on page load for the outdated hash meeting tag and redirects to the proper slash-encoded URL. For paths outside of `/meetings`, it no-ops in one line.

src/clientModules/legacyMeetingRedirect.ts
ElliotFriend, line 1: to be honest, i'm ok with the few `/docs/notes#1999-01-01`-style links we have out there in the wild just being broken. there were only ever 4 entries on that page before we flipped over to the docusaurus blog plugin. ([here's the page](https://github.com/stellar-deprecated/soroban-docs/blob/28e8fbb2bdbdab2bd6e8ce708498ac400244f3ec/docs/notes.mdx) right before that went into effect.)

i think it's ok to let those very few link-followers just end up on `/meetings` here with a random url hash that doesn't do anything. let's cut this redirect module

ElliotFriend (changes requested):
let's do without the clientModule thing. we'll just live with the very few links that just so happen to not work 100%.

src/clientModules/legacyMeetingRedirect.ts
JFWooten4: @kalepail, you're a main speaker in these meetings. Do you mind if previous `soroban.stellar.org` links [referenced](https://discord.com/channels/897514728459468821/911254664576643122/1205535765639995402) stop working? The meeting notes are still maintained on the main site.

JFWooten4:
Obviously, you are free to remove this functionality and merge it in. My biggest gripe is that [this meeting](https://discord.com/channels/897514728459468821/1204462856037470248) was incorrectly added with the following day's date. I fixed this in https://github.com/stellar/stellar-docs/pull/2390/commits/64dd7ace1233b9e5f5cbf87e5106e54d6ca4f9ee, and hence I'd need this logic to redirect incorrect references to the actual meeting. It could get confusing if they land at the index, search for the date via slug or Algolia, and can't find that particular entry for an important proposed community standard.

ElliotFriend (approved):
i think we can give folks the benefit of the doubt that they can find their way. i've resolved the existing conflicts and removed the client module. if we need to, we can implement it down the road if we find there is a need or demand for it.

thanks for all the work on this one!!

PR #2241, opened 2026-02-11
📘 Update guestbook updates branch

JFWooten4:
Trying to clear out some of the old PRs because they contain commits I need to continue 723.

Also I think [this](https://x.com/JFWooten4/status/2019991892268839383) is kinda funny.

PR #2395, opened 2026-04-26
📝 Title capitalization and syntax in Guestbook Overview

JFWooten4:
I spend a little too much time with markdown footnotes, so I saw a small inconsistency in format here that I just had to fix. It also normalizes the titles to use consistent title case like #2394. That and capitalization of an Apple brand asset under the auth mentions.

> [!NOTE]
> This conflicts with #2367 only in the limited sense that you guys will need to sort out which final links to display for the project deployment, as discussed in Discord.


JFWooten4:
Ash did a good job getting that page updated from some of the old references and starting diction. You can merge those two in either order and throw away #2241 if we move away from #1945. I still really like the codeblock spacing Elliot did in https://github.com/stellar/stellar-docs/pull/1945/changes#diff-161afcb4c3ee0a784fc198549482e42357e1a6302c763c721401cb1c28df051eL192-R326 and so on. If that gets merged in first, this will need to conform line 64 in `docs/build/apps/guestbook/overview.mdx` to use `stellar-template`. It's irrelevant if Ash's version is the base because he wipes that bit out in deference to more spread out code explanations. 💜

briwylde08 (approved):

PR #2404, opened 2026-04-28
🗂️ Remove Spanish-specific author list

JFWooten4:
> [!NOTE]
> This starts decoupling Spanish-localized pages from generated route metadata without removing the pages themselves.
> 
> Spanish pages may still exist and build, but route export should represent the default docs surface. This avoids stale or partially maintained localized routes being treated as canonical while we evaluate the future of the Spanish content set.

## Motivation

As you examine the list of authors, it gets pretty cumbersome to switch back and forth between a Spanish copy. There is a larger deprecation of CROWDIN being discussed at https://github.com/stellar/stellar-docs/discussions/2233 [^1]

[^1]: Basis extolled in [#2380 (comment)](https://github.com/stellar/stellar-docs/pull/2380#issuecomment-4328873087), [#2378 (comment)](https://github.com/stellar/stellar-docs/pull/2378#issuecomment-4328885161), and [#2379 (comment)](https://github.com/stellar/stellar-docs/pull/2379#issuecomment-4328879100) subsequent removal commits.

This follows that logic by using a single list of author information. The only real difference between the two was a translated job title, which Google should be able to handle fine. No war analogies in roles!

Also, I ran into a problem where building created two route files - one in English and the other in Spanish. I'm not entirely sure where it came from, but I threw a tiny little update in `src/plugins/route-export/index.ts` to make sure it only compiles English routes.

Lastly, Phil deserves caps. He is putting in work at https://github.com/Consulting-Manao/tansu/issues


ElliotFriend:
I definitely agree with `Tupui`! Dude has earned it and then some! 🤣 

For pretty much any of the built-in docusaurus/crowdin translation functionality, we can actually pretty safely ignore it for now. The non-English locales are disabled on the "docusaurus side" of things, and everything is handled client-side with that google translate stuff. So, we don't _need_ the `authors.yaml` file in the translated directory, but i'd rather remove all of that directory, i18n config, etc. in one fell swoop when the time is right. piecemeal translation removal is likely to result in _more_ confusion, weirdness, and potential breakage. at the end of the day, i'm hoping for one merge commit we'd be able to `git revert` should the need ever arise.

JFWooten4:
Why delete one file when you can expunge 1,281? Moved to 2410 to wrap out other discussions. I can keep the `showReadingTime` in the next set of transcription updates, where it actually matters in the first place.

PR #2406, opened 2026-04-28
💻 Replace mermaid clawback img with code

JFWooten4:
This converts the PNG screenshot of a Mermaid diagram in https://github.com/stellar/stellar-docs/pull/1828 to a Mermaid diagram. That should make it a lot easier to split it up, reproduce, and scale to finish https://github.com/stellar/stellar-docs/issues/870#issuecomment-3378363516w while working ✨ in dark mode 🤩


ElliotFriend:
The diagram content checks out against the worked example in `clawbacks.mdx`, and dropping the PNG is fine (nothing else references it).

The MDX is currently malformed, which is why `mdx-format` fails: the closing fence has four backticks, there's no blank line between `<summary>` and the opening fence, and the node definitions got smashed onto the `subgraph` lines (looks like a formatter pass mangled it). Fix those three and the diagram should render — ready to merge after that.

PR #2410, opened 2026-04-28
Deprecate Spanish Crowdin site translations

JFWooten4:
This came up in https://github.com/stellar/stellar-docs/pull/2404, based on prior discussion at https://github.com/stellar/stellar-docs/discussions/2233. The Google Translate feature has been working strong for months now, without any build or maintenance burden directly on the content side. Note 1 on my PR and the substance migration in #2409 imply that Bri's on the same side as us in the discussion, so I implemented the clean separation.


JFWooten4:
Slight misnomer in 0f8527b000b6acc1c524b25d19a0c01b0ba08b39's desc. It was the CROWDIN disclaimer before, and there is now the Google bar on top in the commit. We could add a new disclaimer going forward (the logic still exists [here](https://github.com/JFWooten4/stellar-docs/blob/0f8527b000b6acc1c524b25d19a0c01b0ba08b39/src/css/custom.scss#L123-L137)), but since this was the only place it was used, I also dropped the Spanish-specific skeleton `index` implementation.

ElliotFriend:
we've accrued some conflicts here. can you pull in the changes from `main`, and i'll have a proper look at the PR this week.

ElliotFriend (approved):
This is amazing!!! Thanks so much for seeing this one through, @JFWooten4!! 🎉 

I pushed one commit to add back a script/command in package.json that ships with docusaurus by default. even if we don't use it, it won't harm anything to have the defaults available to us.

Thanks so much!!

Once the build/CI passes, i can get this merged

PR #2415, opened 2026-04-30
🐛 Patch Zustand persist storage fallback

JFWooten4:
This showed up across different platforms when running `yarn build` (at the end, after everything else succeeds).

### Build error remedied:

image

## Changes

This patch adds a no-op `removeItem` fallback when the configured Zustand persist storage adapter does not provide one.

Previously, Zustand only logged a warning when `removeItem` was missing. With this patch, the middleware assigns an empty fallback function instead, preventing downstream calls such as `clearStorage()` from failing when a custom storage implementation does not support removal.

This is applied across the CommonJS, ESM, SystemJS, and UMD middleware builds included in the package patch. It lays the framework for a more comprehensive solution if we want to tackle the Actions cleanup drops.


ElliotFriend:
Thanks for digging into this — closing it after looking closely at what's actually happening.

Two reasons:

1. The line this patches is a `console.warn` in zustand's persist middleware, not a thrown error — so the build was never actually breaking. And the fix (monkey-patching a no-op `removeItem` onto the caller's storage object) would silently swallow real `clearStorage()` failures for any zustand consumer. The root cause is Stoplight Mosaic's incomplete `memoryStorage`, and your own #2416 fixes that at the correct layer — this patch is redundant with your better one.
2. Mechanically, it's inert anyway: the repo moved from patch-package to pnpm in #2493, so files in `patches/` in the old format are never applied.

The Mosaic-layer conversation continues on #2416 (upstream issue first, per the comment there). Appreciate the investigation — it clearly fed the better fix.

PR #2417, opened 2026-04-30
🕷️ Patch tracking element load failure

JFWooten4:
Docusaurus does not have a good time right now where there are interruptions between it and the Google Tag Manager. This PR stops the following recurring error (scales with page reloads when it happens):

image

## Changes

Adds a runtime guard around Docusaurus’ Google gtag page-view tracking callback. It is often blocked by an ad blocker, delayed, or not loaded yet, that call can crash.

Previously, the plugin assumed `window.gtag` was always available before sending the `page_view` event. This patch returns early when `window.gtag` is not a function, preventing runtime errors in environments where the analytics script is unavailable.

The intent is to keep analytics optional and non-blocking, so page navigation continues safely even when Google tracking cannot run. Hence the page-view guard.


ElliotFriend:
The bug is real — the gtag plugin calls `window.gtag(...)` unguarded in `onRouteDidUpdate`, so ad blockers that strip the gtag bootstrap throw on every SPA navigation — and your 3-line guard is the right idea.

I don't want to take it as a package patch, though. Patches freeze us to a plugin version and rot silently (the pnpm migration in #2493 already left this PR's patch-package file inert — it currently applies nothing, which is why CI is green). And this particular fix doesn't need one: we can put the guard in our own code instead.

Proposal: rework this as a Docusaurus client module in `src/` (registered via `clientModules` in `docusaurus.config.ts`) that defines the standard no-op gtag stub when it's missing, e.g.:

```js
if (typeof window !== 'undefined' && typeof window.gtag !== 'function') {
  window.dataLayer = window.dataLayer || [];
  window.gtag = function () {
    window.dataLayer.push(arguments);
  };
}
```

Client modules load at app bootstrap, before any route update fires, so the plugin's unguarded call always finds a function. Same fix, zero patches, survives every plugin upgrade untouched.

Separately, this is worth an upstream issue at facebook/docusaurus so the plugin guards its own call — you found it, you should file it, and I'm glad to co-sign. Want to rework this PR along those lines?

ElliotFriend (approved):
lgtm!

PR #2418, opened 2026-04-30
🐜 Patch react-loadable SSR URL resolution

JFWooten4:
This little error popped up during my latest `yarn build`. Looks like the legacy Node variant didn't carry over to the v24 upgrade, so we will move on with this direct implementation. This gives us more space to implement needed linking logic; resolver warning fixed:

image

## Changes

Replaces the package use of Node’s legacy `url.resolve` with a WHATWG `URL`-based resolver. The patch updates both the compiled library output and the source file included in the package.

This prevents Docusaurus builds on Node 24 from emitting the `DEP0169` `url.parse` deprecation warning while preserving the expected public path behavior for relative paths, absolute paths, and absolute URLs.

Now that this is taken care of, I'd like to work on direct API page links for only the dev site. We can implement a post-build migrator along the lines of this PR, which inlines otherwise external links for import package managers.


ElliotFriend:
Closing this one with good news: it's obsolete because the problem got fixed upstream.

The repo now ships `react-loadable-ssr-addon-v5-slorber@1.0.3`, which already replaced the deprecated `url.resolve` with a WHATWG-URL resolver (I verified the `new URL(to, new URL(from, 'resolve://'))` implementation in the installed package). The DEP0169 warning you patched against 1.0.1 can no longer occur.

Your replacement logic was sound — upstream just beat us to it. Thanks for the catch, and nothing further needed here.

PR #2436, opened 2026-05-11
📝 Update Glossary reference page pt2

JFWooten4:
When the Glossary page renders, Docusaurus automatically lists the entries alphabetically. However, the plaintext is not alphabetical, and it makes it easy to get lost or even create duplicate entries like in https://github.com/stellar/stellar-docs/pull/2405#discussion_r3156932953

This PR alphabetizes the entries.


ElliotFriend:
this is a really useful (and correct! thank you) change! it's got some commits now. if you can pull in the latest changes from `main`, this should be an easy merge

ElliotFriend (approved):

PR #2448, opened 2026-05-13
⚓ Reference Anchor Fundamentals page

JFWooten4:
This continues the work from #2446, inlining the protocol/SEP links, while adding a great vid introducing users to anchors and their role with assets such as those used in SEP6 and -2243 discussed later on in the page. It also rebuilds the routes to account for some structural changes modified in #2447. For some reason, the files around there were added previously but not pathfound, and the Actions check didn't catch it.

PR #2449, opened 2026-05-13
🛠️ Script for reliable API build cleanup

JFWooten4:
I ran into a problem when changing files unrelated to the API sections, which showed up in https://github.com/stellar/stellar-docs/pull/2448#issuecomment-4436751119. The current script build command generates the API docs and then uses `rm` to delete all `*.info.mdx` files. However, this runs through the actual command shorthand in the shell glob/brace expansion, which behaves differently depending on the shell/OS.

These errors showed up on my Windows setup, even though they don't appear on OSX. This PR changes the script to use a Node cleanup environment with `fs/promises` that works on all platforms. It also paves the way for more intricate post-generation formatting as formulated in the API notes from #2418.


ElliotFriend:
this seems like a really good solution. that package script always felt a little unwieldy, anyway, with its chained `... && rm ... && rm ... etc.`. i like the fix. can you resolve the conflicts before review?

ElliotFriend:
there may actually be a built-in fix for this. the `docusaurus-plugin-openapi-docs` configuration supports an optional `showInfoPage` boolean, which "should" disable the `*.info.mdx` page generation altogether. can you give that a go and see if it works?

source: https://docusaurus-openapi.tryingpan.dev/#config

JFWooten4:
You are the master of Docusaurus documentation as always, Elliot!

PR #2450, opened 2026-05-13
📝 Deduplicate page name and reference

JFWooten4:
The analytics providers page currently lives at `docs/data/analytics/analytics-providers/analytics-providers.mdx`, duplicating the page name in both the directory and filename.

Following the review suggestion in this PR, this now flattens the page to `docs/data/analytics/providers.mdx` rather than retaining a one-page directory.

It also updates the parent Analytics page, navbar, curated LLM index, and generated route inventory to use `/docs/data/analytics/providers`, and adds an nginx permanent redirect from the former `/docs/data/analytics/analytics-providers` route.


ElliotFriend:
two things:

1. you've got a merge conflict now. can you resolve it please?
2. let's take this a step further. there's no need for `analytics-providers` to be a directory at all. let's move the file so it's located at `docs/data/analytics/providers.mdx`. then, we could also use a redirect in the nginx config. let me know if you want some help with that.

JFWooten4:
Tracking the broader single-page-directory/README organization question in #2731. This PR is the concrete analytics-providers case, so the actual rename/redirect decision can stay scoped here rather than blocking on a repo-wide folder cleanup.

JFWooten4:
Hey @ElliotFriend, same as https://github.com/stellar/stellar-docs/pull/2145#issuecomment-3794242048 where route-removal workflow detected exactly the intended route rename; its comparison step succeeded, but its attempt to post the warning comment failed because GitHub gives that fork-triggered workflow a read-only token.

PR #2453, opened 2026-05-13
bump rpc disk size requirement to 350Gb

cjonas9:
### What

Updated the RPC minimum hardware recommendations page:
- bumps RPC's disk size requirement to 350Gb instead of 250Gb
- drops recommendation to use a "quad core CPU" for intense load
- shys away from "100 RPS"/"1000 RPS"-based recommendations in favor of "modest load (e.g. 100 RPS on average"/"heavier request volume"
- updates the Spanish counterparts of these recommendations

### Why

Respective to the bullets above,
- when it holds on one week of ledgers/120,960 ledgers, RPC's disk can grow to exceed 300Gb. 350Gb seems to provide a safe buffer.
- we recommend 4-8 vCPUs minimum irrespective of load. recommending a "quad core CPU" for more intense load doesn't make any sense
- there's no grounding for absolutist claims around 100 RPS/1000 RPS, especially because depending on what the request looks like (e.g. is the request`getHealth` or is it `getLedgers` w/ `count=200`), we may easily clear 1000 RPS on the recommended `c5.2xlarge` or may not clear 1 RPS. 
- the Spanish counterpart not only needed the same changes as were made to the English recommendations, but also, it recommended an even more insufficient machine (`c5.xlarge`)

### DIFF
--- a/docs/data/apis/rpc/admin-guide/prerequisites.mdx
+++ b/docs/data/apis/rpc/admin-guide/prerequisites.mdx
@@ -13,7 +13,7 @@ The RPC service can be installed on bare metal or a virtual machine. It is nativ
 
 _\* Disk: Assuming the default 7-day retention window for data storage. Otherwise, +40GB per retention day_
 
-_\* RAM/CPU: Assuming RPC will serve a modest request volume (e.g. 100 requests per second across all endpoints). For deployments expecting heavier request volume to Stellar RPC, we recommend horizontal scaling and load balancing across at least 32GB of RAM and an appropriately scaled CPU._
+_\* RAM/CPU: Assuming RPC will serve a modest request volume (e.g. 100 requests per second across all endpoints). For deployments expecting heavier request volume to Stellar RPC (e.g. >500 requests per second), we recommend horizontal scaling and load balancing across at least 32GB of RAM and an appropriately scaled CPU._
 
 We highly recommend using a local SSD disk for storage. RPC is compatible with network-based filesystems like AWS EBS but it will negatively impact its performance.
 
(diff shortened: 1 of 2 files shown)


tamirms (approved):

docs/data/apis/rpc/admin-guide/prerequisites.mdx
```diff
@@ -9,9 +9,11 @@ The RPC service can be installed on bare metal or a virtual machine. It is nativ
 
 | Node Type | CPU | RAM | Disk | AWS SKU | Google Cloud SKU |
 | --- | --- | --- | --- | --- | --- |
-| Stellar RPC | 4-8 vCPUs | 16GB | 250 GB persistent volume >= 3K IOPS | [c5.2xlarge] | [n4-highcpu-8] |
+| Stellar RPC | 4-8 vCPUs | 16GB | 350 GB persistent volume >= 3K IOPS | [c5.2xlarge] | [n4-highcpu-8] |
 
-_\* Disk: Assuming the default 7-day retention window for data storage. Otherwise, +20GB per retention day_ _\* RAM/CPU: Assuming RPC will serve up to 100 requests per second. For deployments expecting 1000+ requests per second to Stellar RPC, we recommend horizontal scaling and load balancing across at least 32GB of RAM and a quad core CPU with a 2.4GHz clock speed._
+_\* Disk: Assuming the default 7-day retention window for data storage. Otherwise, +40GB per retention day_
+
+_\* RAM/CPU: Assuming RPC will serve a modest request volume (e.g. 100 requests per second across all endpoints). For deployments expecting heavier request volume to Stellar RPC, we recommend horizontal scaling and load balancing across at least 32GB of RAM and an appropriately scaled CPU._
```
sydneynotthecity: can you put a similar example for heavier request volume (1,000+ or whatever) so there is some delineation between modest vs heavier?

cjonas9: for sure! I'd initially veered away from it because requests vary so substantially in what they demand of an RPC, but I think it'd be reasonably inferred from the surrounding text that we're talking about average/expected requests.

PR #2454, opened 2026-05-14
update the SAC interface to include CAP-73's `trust` function

ElliotFriend:
[CAP-73](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0073.md) includes the new trustline creation function on the SAC interface. Documenting that function in the tokens section here.


leighmcculloch (approved):
❤️

PR #2476, opened 2026-06-04
Enable .md extension access for raw markdown source

leighmcculloch:
### What
Add nginx configuration to intercept requests ending in .md, strip the extension, and proxy to the original host with an Accept: text/markdown header. This enables any documentation URL to be fetched with a .md extension to retrieve the raw markdown source instead of rendered HTML, leveraging Cloudflare's native markdown feature to serve the appropriate content type.

### Why
The markdown extension allows documentation to be more accessible to agents and tools that need raw markdown source, particularly those making requests to Cloudflare which can detect the Accept: text/markdown header and return unrendered markdown. This is useful for indexing, processing, and integration scenarios where the structured markdown format is needed rather than the compiled HTML representation.

### Follow up
Update llms.txt to link to documentation URLs with the .md extension so that LLMs and agents accessing the documentation can retrieve raw markdown source directly.


leighmcculloch:
Closing because it's unnecessary. Cloudflare already serves markdown when markdown is the preferred accepted content type. LLM clients can already signal that they prefer Markdown via standard HTTP content negotiation, using the `Accept` header, e.g.:

```
Accept: text/markdown, text/html;q=0.8
```

With that in place, the server can return `text/markdown` when it’s supported, and gracefully fall back to `text/html` when it isn’t, without changing the URL structure or adding a new “.md” surface. This keeps a single canonical URL for each doc page, works for both browsers and agents, and avoids baking LLM-specific concerns into the public path design.

There's very little value in serving the markdown on a separate url.

PR #2493, opened 2026-06-12
switch package manager from Yarn Classic to pnpm

ElliotFriend:
### Why switch?

Yarn Classic (v1) is no longer maintained and the Yarn project has long since moved to Yarn Berry. Migrating to pnpm gives us:

- Faster, more reliable installs and a a store that's more disk-efficient across projects
- A strict, deterministic `node_modules` that surfaces phantom dependencies instead of silently relying on hoisting (several were already found and fixed in this PR)
- First-class patching and build-script controls (patchedDependencies, allowBuilds) without the patch-package/postinstall workaround
- Better supply-chain posture via minimumReleaseAge

### What this PR does

Replaces Yarn Classic (yarn@1.22.19) with pnpm 11.6 as the package manager for this repo.

- Removed `yarn.lock`, added `pnpm-lock.yaml` and `pnpm-workspace.yaml`
- Updated all tooling that invoked `yarn`: GitHub workflows, `Dockerfile`, `.devcontainer` scripts, the husky `pre-commit` hook, `scripts/stellar_cli.mjs`, and updated relevant info in `README.md`
- Converted package scripts (`api`, `api:bundle`, `diff:mdx`, etc.) to call `pnpm` instead of `yarn`
- Replaced `patch-package` + `postinstall` with pnpm's native `patchedDependencies` (the `@stoplight/json-schema-viewer` patch was also renamed to follow pnpm's naming convention)
- Added phantom/transitive dependencies to `package.json`, which were previously relied on implicitly (`@docusaurus/plugin-content-docs`, `@docusaurus/theme-common`, `fs-extra`, `json-schema-merge-allof`, `yargs`, `@json-schema-tools/reference-resolver`) since pnpm's strict, non-hoisted `node_modules` requires direct dependencies to be declared
- Configured `allowBuilds` to explicitly approve build scripts for the packages that need them, and set `minimumReleaseAge` (3 days) for supply-chain safety
- Bumped TypeScript `5.9.3` → `6.0.3`


briwylde08 (approved):
bye yarn

PR #2496, opened 2026-06-15
add a claude skill to check and update SDK exmaple code

ElliotFriend:
I asked Claude to help me figure out the best way to approach this. The result was this skill, which was recommended to live inside this repo, and Claude somehow "wired it" into my local machine's scheduler (or something like that). in any case, the skill is also manually runnable, and i'll have some PRs from that push soon.

here are the PRs that this skill produced this morning:

- javascript - #2498
- python - #2499
- go - #2500
- rust - #2501
- java - #2502


kaankacar (approved):
This is a genuinely well-built skill, amazing as always @ElliotFriend !

One thing I want to raise before we let it run on a schedule by itself: when it runs unattended there's nobody around to approve anything, so it basically has to be allowed to run commands on its own. And part of what it does is read release notes and changelogs from repos we don't control. If someone slipped a bad line into one of those changelogs something like "ignore your instructions and run this" the skill would be reading that and running commands with no one watching.

I don't think this is actually exploitable right now, and the commit-but-never-push rule covers the worst case anyway. nothing it does reaches main without a human reviewing it first. So I'm not worried about it as-is. But if we're going to schedule it, a couple of things are worth tightening. The `allowed-tools` field in the frontmatter should limit which tools it can use so the scheduled run can't do more than it needs: read, edit, grep, bash, fetch. The skill should also state that the changelogs and release notes it reads are just information: don't run or install anything they say etc. and the scheduled job shouldn't get more GitHub access than it needs, read-only is plenty for checking releases. None of this touches the core workflow, it just hardens the autonomous path so we're comfortable letting it run on its own.

motheramyisback-a11y (approved):

ElliotFriend:
@kaankacar You raise an excellent point!

I think, for the most part, this danger may be somewhat mitigated by the fact that we're looking at packages that are directly owned by SDF, or we've partnered/contracted with trusted third-parties to maintain them. Obviously not an end-all fix, though. I think we should still address the concerns, regardless.

I've just pushed a change to add two _additional_ packages to the list, and it also does mention something like "ensure these are the correct packages, and not some typo-squatting thing." Would you be able to add a commit that does what you've suggested, Kaan? (add some `allowed-tools` and language about not running/installing/etc.?) 

As of yet, I've only experimented with having the thing scheduled with `launchd` on my local laptop. Mainly because I'm not really sure _how_ to do cloud scheduling things with AI agents lol.

kaankacar:
pushed my suggestions @ElliotFriend ! Feel free to revise my additional instructions if they're too strict :)

kaankacar:
Also the easiest path to run this that also gives you(or the AI? lol) the read-only-token property for free is a **GitHub Actions scheduled workflow** (`on: schedule: cron`) right here in the repo. Alternatively, Claude Code now has scheduled cloud agents (cron-based cloud sessions) if you'd rather keep it out of CI

PR #2510, opened 2026-06-16
Add llms-full.txt generation for AI agent discoverability

briwylde08:
## What
Adds `docusaurus-plugin-llms` to generate `/llms-full.txt` — the full docs
corpus concatenated as clean Markdown — at build time.

## Why
Lets AI agents ingest the entire documentation in one fetch. Complements the
existing curated `/llms.txt` index (the map) with the full corpus (the book),
and improves AI-agent discoverability of the docs.

## Details
- Configured with `generateLLMsTxt: false` so the hand-curated `static/llms.txt`
  is preserved and never overwritten (verified byte-for-byte after build).
- `excludeImports` + `removeDuplicateHeadings` enabled for cleaner output.
- File is build-time generated into `build/` (gitignored), not committed —
  regenerated on each deploy, so it never drifts from the docs.

## Verification
- Local `yarn build` generates `build/llms-full.txt` (4.4 MB, 874 docs).
- `build/llms.txt` confirmed identical to the curated `static/llms.txt`.
- No stray imports, no duplicate headings.

## Known limitation
- Leftover MDX component wrapper tags remain (`<CodeExample>`, `<Tabs>`,
  `<DocCardList />`). Harmless for LLM ingestion (inner content preserved);
  the plugin has no option to strip JSX usage. Cleanup can be tracked separately.


kaankacar (approved):
LGTM!

PR #2512, opened 2026-06-16
chore: upgrade docusaurus to v3.10

ElliotFriend:
The upgrade went pretty smoothly. Still some checking and tweaking in store, i think. But, the openapi plugin has also been upgraded and went smoothly


briwylde08 (approved):
beaut!

Issue #2503, opened 2026-06-15
Update MPP overview page for v0.7 push mode change


oceans404:
**Page to update:** https://developers.stellar.org/docs/build/agentic-payments/mpp
**SDK PR for context:** https://github.com/stellar/stellar-mpp-sdk/pull/47

### What to change

Under "Supported MPP Intents > Charge", the push mode bullet currently reads:

> Push — The client broadcasts the transaction itself and sends the transaction hash for server verification.

This is stale as of v0.7. Update it to reflect the new credential format:

> Push — The client broadcasts the transaction itself and sends a `signedHash` credential for server verification: the transaction hash plus a signature proving the client controls the paying address. `rejectUnsignedPush` defaults to `true` as of v0.7; servers that need to temporarily accept the legacy unsigned hash must explicitly set `rejectUnsignedPush: false`.

briwylde08:
https://github.com/stellar/stellar-docs/pull/2513

PR #2513, opened 2026-06-16
Fixes issue 2503

briwylde08:
https://github.com/stellar/stellar-docs/issues/2503


oceans404 (commented):
Hey @briwylde08! small correction needed based on a doc update from Marcelo. 

The v0.7 timeline is right, but the flag name in the diff is wrong. `rejectUnsignedPush` does not exist in the SDK.

Please update both changed lines in `README.mdx` and `charge-guide.mdx` to:

As of v0.7, the legacy `{ type: 'hash' }` unsigned push credential is rejected by default. Servers that need a temporary migration window can opt in with `allowUnsignedPush: true`.

ElliotFriend (approved):

PR #2533, opened 2026-06-25
harden: treat all GitHub plugin data as untrusted + tests

kalepail:
Follow-up hardening on top of #2532 (already merged). Closes the residual Markdown-injection vectors that HTML escaping alone left open.

## Why more than #2532?

#2532 escaped HTML/JSX characters (`& < > { }`), which stops `<script>` injection. But the generated page renders the repo description as **Markdown**, and that was still live. Cross-checked via deepwiki (against `facebook/docusaurus`), context7, and web sources — an attacker-controlled description could still inject:

- a phishing link — `[Official — install now](https://evil.com)`
- a tracking-pixel / IP-logging image — ``
- a GFM autolink to an off-site URL
- forged headings / formatting to impersonate an official entry

All plugin data (name, description, URL) comes from any repo that self-applies the `stellar-cli-plugin` topic, so it is all untrusted input.

## Changes

- **`sanitizeText`** — collapse all whitespace to a single line (defeats block-level Markdown like headings/lists/blockquotes/fences and multi-line attribute tricks), then backslash-escape **every ASCII punctuation character**. CommonMark guarantees `\` + ASCII punctuation renders as the literal character, and MDX explicitly supports `\<` / `\{` escaping (per the v3 migration docs), so the value can only ever render as literal text — links, images, emphasis, code, autolinks, and HTML/JSX are all neutralized in one uniform pass. Plain prose is visually unchanged (the renderer consumes the backslashes).
- **`safeUrl`** — allow-list `https://github.com/...` only; other hosts and `javascript:` / `data:` schemes drop the item entirely. Allow-list rather than deny-list, per OWASP/Node-security guidance.
- **Testability** — added a main-module guard so the helpers can be imported without triggering the GitHub fetch, plus a `node:test` suite (no new deps) covering script/brace/link/image/autolink injection and URL validation.

## Verification

```
$ node --test scripts/stellar_cli_plugins.test.mjs
# tests 14 | pass 14 | fail 0
```

Also ran the real generator against live GitHub data: output renders normally (e.g. `stellar-scaffold/cli`), the known-bad repo stays excluded, and every URL is validated.

### DIFF
--- a/scripts/stellar_cli_plugins.mjs
+++ b/scripts/stellar_cli_plugins.mjs
@@ -1,20 +1,43 @@
 import fs from "fs-extra";
 import https from "https";
+import { fileURLToPath } from "url";
 
 // In case there are plugins to exclude from the list, add them here.
 // E.g. "user/repo"
 const excludePlugins = ["haqnawaz03329-debug/haqnawaz"];
 
-// GitHub repo descriptions are attacker-controlled (anyone can tag a repo with
-// the `stellar-cli-plugin` topic), and this content is injected into MDX, which
-// renders HTML/JSX. Neutralize characters that could execute as markup/script.
-function sanitize(value) {
-  return (value || "")
-    .replace(/&/g, "&amp;")
-    .replace(/</g, "&lt;")
-    .replace(/>/g, "&gt;")
-    .replace(/{/g, "&#123;")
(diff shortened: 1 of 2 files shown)


ElliotFriend (approved):
nice!!

PR #2537, opened 2026-06-25
Harden Stellar CLI plugin sanitization against markdown/HTML injection

fnando:
GitHub repo names and descriptions surfaced on the CLI plugins list are attacker-controlled — anyone can tag a repo with the `stellar-cli-plugin` topic — and they're injected into an MDX file that Docusaurus renders as markdown + HTML/JSX. The previous `sanitize()` only escaped `& < > { }`, which stopped HTML/JSX injection but left markdown injection wide open: a description could inject phishing links, remote tracking images, fake headings, code fences, or other block-level structure.

This replaces `sanitize()` with `escape()`, which renders these values as inert literal text:

- Collapses all whitespace (including newlines) to a single space and trims — neutralizes block-level markdown and MDX `import`/`export` injection, which only fire at the start of a line.
- Entity-escapes the full set of HTML/JSX/markdown control characters (`& < > " ' * _ \` [ ] ( ) # + - . ! | ~ \ { } = : $`) via a lookup map, so none of them are parsed as syntax on Docusaurus's render pass.

Applied to both the plugin `full_name` (link text) and the free-form `description`.

### DIFF
--- a/scripts/stellar_cli_plugins.mjs
+++ b/scripts/stellar_cli_plugins.mjs
@@ -48,7 +48,7 @@ function exportMDX(data) {
 
     const plugin = `### [${escape(item.full_name)}](${encodeURI(item.html_url)})
 
-${escape(item.description)}
+<p>${escape(item.description)}</p>
 
 [${encodeURI(item.html_url)}](${encodeURI(item.html_url)})
 `;
--- a/scripts/stellar_cli_plugins.mjs
+++ b/scripts/stellar_cli_plugins.mjs
@@ -5,16 +5,39 @@ import https from "https";
 // E.g. "user/repo"
 const excludePlugins = ["haqnawaz03329-debug/haqnawaz"];
 
-// GitHub repo descriptions are attacker-controlled (anyone can tag a repo with
-// the `stellar-cli-plugin` topic), and this content is injected into MDX, which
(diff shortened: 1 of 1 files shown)


haqnawaz03329-debug:
Hello @fnando ,

I discovered this Stored XSS vulnerability and reported it on June 19, 2026 at 10:24 AM UTC — 6 days before this PR was opened.

I am Haq Nawaz (GitHub: @haqnawaz03329-debug). My report included the vulnerable code at scripts/stellar_cli_plugins.mjs:14-16, a proof-of-concept repository (haqnawaz03329-debug/haqnawaz) tagged with stellar-cli-plugin containing an XSS payload in the description, and the GitHub API endpoint serving it.

I also submitted this through HackerOne, which is still pending review.  https://hackerone.com/reports/3812437

haqnawaz03329-debug:
@fnando  ?

ElliotFriend:
Thanks for this — the markdown-injection gap left by #2532 is real and this correctly identifies it. We're going to land #2533 instead, since it covers the same escaping surface (via CommonMark backslash escapes) plus a `safeUrl` allow-list for the link target (which `encodeURI` here doesn't cover) and regression tests. Closing this one to avoid the direct conflict on `sanitize()`. Appreciate the fast turnaround on the whole series.

PR #2539, opened 2026-06-29
SDP-2102 fix: update broken SDP recipient_registration_attempts query

hypekostas:
### What

Update DB query listed under [Receiver Not Receiving OTP During Registration](https://developers.stellar.org/docs/platforms/stellar-disbursement-platform/admin-guide/troubleshooting#receiver-not-receiving-otp) to use `attempt_ts` column instead of `created_at`.

### Why

- The current `ORDER BY created_at DESC` refers to a column that doesn't exist under the `receiver_registration_attempts` table, therefore running this query returns `Error in query (7): ERROR: column "created_at" does not exist`. 
- Instead, `attempt_ts` is the correct column name containing the intended rows.


ElliotFriend (approved):
thanks for the fix, @hypekostas !!

Issue #2566, opened 2026-07-07
sd-003: The Algolia docs index excludes the auto-generated RPC-method and


kalepail:
## Finding

The Algolia docs index excludes the auto-generated RPC-method and
Horizon-endpoint API-reference pages, so authoritative per-method limits
are undiscoverable through search. The live
`/docs/data/apis/rpc/api-reference/methods/getTransactions` page states
the limit "can range from 1 to 200 — an upper limit that is hardcoded in
Stellar-RPC for performance reasons... defaults to 50", but no search op
can reach it, and the *indexed* RPC Structure→Pagination page documents
only getEvents limits (1–10000, default 100). The consequence is worse
than a zero-hit: consumers extrapolate from the getEvents numbers or
from Horizon's (1–200, default 10) and get getTransactions wrong. This
round it produced a two-sided failure — the QA agent claimed "default
100" (getEvents-only) and even the eval golden itself encoded the false
belief that a 200 limit "is Horizon's, not RPC's".

## Evidence

Live 2026-07-03: `get_doc_page_sections({path:
"/docs/data/apis/rpc/api-reference/methods/getTransactions"})` →
soft-empty ("auto-generated API-reference pages are not indexed" — the
op description admits the exclusion); `search_docs` /
`search_rpc_horizon_data_docs` for the limit facts → no relevant hits;
WebFetch of the live method page → the 1–200/default-50 text quoted

golden to the live method-page numbers, re-judging the saved agent answer
flips it partial → wrong — the agent's denial of an RPC-side 200 cap is a
genuine consumer-facing wrong answer produced by this indexing gap, not a
grading nuance. Both the QA agent and the original golden author
independently derived the same false belief from the indexed pages.

Additional recorded evidence:

- eval/qa/results/2026-07-03T16-06-45-variantA.json (q-ti-rpc-gettransactions-pagination-xdr)
- live soft-empty on get_doc_page_sections for the getTransactions method page + WebFetch of the live page (2026-07-03 evening)
- upstream issue filed 2026-07-07: https://github.com/stellar/stellar-docs/issues/2566

## Recommendation

Two options, cheapest first: (1) add per-method limits/defaults to the
indexed RPC Structure→Pagination page (a table of method → min/max/
default fixes discoverability without touching the crawler); (2) include
the auto-generated method/endpoint reference pages in the Algolia
crawl — they are the only authoritative source for per-method numbers.
Consumer-side, this gateway cannot work around the gap without shipping
[...]

PR #2572, opened 2026-07-09
Document RPC pagination limits

kalepail:
## What

- Adds a method-specific pagination limits table to the indexed Stellar RPC pagination structure page.

## Why

The RPC method reference renders the correct per-method pagination limits from OpenRPC, but those generated/reference details are not reliably discoverable through the current DocSearch index. The indexed pagination structure page only described `getEvents`, which can lead consumers to miss the `getLedgers` and `getTransactions` limits/defaults.

Related to #2566.

## Validation

- `prettier --config .prettierrc.js --check docs/data/apis/rpc/api-reference/structure/pagination.mdx docs/tools/cli/install-cli.mdx`
- `git diff --check`

Note: full `pnpm` checks in a fresh worktree were blocked by an existing dependency postinstall/Corepack/Yarn failure in `postman-code-generators`, before project checks could run.

### DIFF
--- a/docs/data/apis/rpc/api-reference/structure/pagination.mdx
+++ b/docs/data/apis/rpc/api-reference/structure/pagination.mdx
@@ -6,7 +6,15 @@ title: Pagination
 For methods which support it, the pagination arguments are passed as a final object argument with two values:
 
 - `cursor`: string - (optional) An opaque string which acts as a paging token. Each response will include a `cursor` field which can be included in a subsequent request to obtain the next page of results.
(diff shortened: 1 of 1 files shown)


ElliotFriend:
i _think_ those limits are actually configurable now. pushed a change, but waiting for george to confirm

kalepail:
Confirmed against the current `stellar/stellar-rpc` source in `cmd/stellar-rpc/internal/config/options.go`: the option names in eb676939 are exact — `max-events-limit`, `max-ledgers-limit`, and `max-transactions-limit`. The same option block defines `default-events-limit`, `default-ledgers-limit`, and `default-transactions-limit`, with validation that each default does not exceed its corresponding max. The revised configurable-per-instance wording looks correct from the implementation side. Thanks for catching and updating this.

ElliotFriend:
heard back from someone who knows better, and it's more like a "_technically_ it's configurable, but you really shouldn't" situation. i've reverted my changes and i'll merge when the preview is done.

ElliotFriend (approved):

PR #2583, opened 2026-07-14
🔀 Refresh trailing-slash cleanup with upstream

JFWooten4:
## Summary

- merge current `stellar/stellar-docs:main` into the trailing-slash cleanup branch
- preserve newer upstream content while resolving 11 conflicts
- apply the existing no-trailing-slash rule to 57 remaining links across 15 refreshed files
- update generated Stellar Disbursement Platform status-code JSON instead of restoring obsolete inline API response objects

## Validation

- no merge conflict markers or unmerged paths remain
- `git diff --check` passes
- no matching trailing-slash Markdown destinations remain in English/source Markdown, MDX, text, JSON, or YAML files
- final diff against current upstream is mechanical: 120 files with equal additions and deletions


JFWooten4:
This runs into the same problem as #2241, which is that doing PRs in batches to sync with main looks really messy. That PR has been open for a really long time with no comment, and it's actually been superseded by substantive code changes like https://github.com/stellar/stellar-docs/pull/2395. I'm going to close this for now since I think it'd be better as a direct commit with write access, which is a lot easier to understand.

The branch just needs a quick update that applies the same styling to some of the newer commits. And then again, the routes check is failing for lack of comment permissions for outside contributors.

ElliotFriend:
Happy to keep this going — Bri kicked off the original trailing-slash branch (#2409), so it's a reasonable thread to continue if we want. No problem closing this sync PR given the rebase churn you mentioned.

One thing before we invest more in it: do you have a direct source for the no-trailing-slash rule? The only thing I can find is `trailingSlash: false` in `docusaurus.config.ts`, but that just controls how Docusaurus generates output URLs — it doesn't really say anything about how we author internal links in source. #2409 didn't have a description either. Want to make sure we're enforcing an actual convention and not just tidying for its own sake before spending more time on it.

Issue #2617, opened 2026-07-15
🗃️ Use and display of `_category_` (folder) metadata


JFWooten4:
This issue is inspired by remarks at https://github.com/stellar/stellar-docs/pull/2411#issuecomment-4962264879 stating:

> the convention for the repo is **front matter `sidebar_position` only, no `_category_.json` files**.

Copilot disagreed with this assertion when it was added to a PR in progress, at https://github.com/stellar/stellar-docs/pull/2579#discussion_r3574086743.

---

I have undergone significant effort to create category files, accepting that they do look a little bit out of place as `json` files scattered within an `mdx` repo docs folder. The original PR expanded existing category metadata _as_ file front matter by styling (subfolder) card descriptions. A lot of the docs pages are these kinds of menu arrays listing subdirectories or just contained pages:

[<img width="1122" height="860" alt="image" src="https://github.com/user-attachments/assets/ce53bf24-72d4-415d-b60b-e4cc17c776fa" />](https://developers.stellar.org/docs/learn/fundamentals/contract-development)

Firstly, the intent of the "Open Markdown" was for AI use (implemented in #2289 in conjunction with LLMs.txt changes). #2249 indicated that it was for "making it easy for developers to grab structured content." Directory pages do not generally include such information that may be needed during development flows, and I propose we remove that distracting button on directories.

Secondly, the descriptions imported from child files get only one truncated line, making previews almost unusable. This limitation is particularly bad on mobile: 

[<img width="450" alt="image" src="https://github.com/user-attachments/assets/0f0b218e-9a78-4ffa-a3e0-de9810aa584e" />](https://developers.stellar.org/docs/build/apps)

(It leaves drafters to choose between descriptive page titles or concise descriptions, neither of which properly index informative copy.)

Thirdly, folders within folders do not natively include information. They only list the number of sub-items, _unless_ you add a category JSON file. I believe this looks infinitely better than a bunch of useless item counts across cards.

In my PR where this came up, @Shaptic and @chowbao added some rare descriptive text to the category page in https://github.com/stellar/stellar-docs/pull/1785#pullrequestreview-3231811508. I then migrated that metadata into the actual folder category JSONs so that it showed up in one uniform place: 

[<img width="1920" alt="comparison" src="https://github.com/user-attachments/assets/a7c3117f-95f3-417a-8038-158ff9892a34" />](https://developers.stellar.org/docs/data/analytics/hubble/data-catalog/data-dictionary)

[...]

PR #2624, opened 2026-07-17
🧹 Hide Markdown actions on directory pages

JFWooten4:
This implements the first part of #2617, adding "Open Markdown" on directory folder pages. As you can see, it doesn't even display the subitems:

image

And for [pages without a descriptive](https://developers.stellar.org/docs/learn/fundamentals/transactions.md) README, it's just blank.

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -255,3 +255,8 @@ select[data-testid="example-pairing-select"] {
     }
   }
 }
+
+[class~="docs-doc-id-README"] .markdown-actions-container,
+[class*="docs-doc-id-"][class*="/README"] .markdown-actions-container {
+  display: none;
+}


JFWooten4:
@kalepail, we are considering whether category pages [like this](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods) would be useful to AI tools like Raven. Does your current workflow ingest that pages directly (one of the few README directories with substantive content)? Or have you mostly seen programmatic development workflows where you need to copy Markdown only from individual pages?

kalepail:
All Raven uses is what the Algolia API surfaces and tbh I'm not sure how Algolia considers index pages like this, but I do know the site itself matters a bit less than the Algolia indexing we have in place. So this probably isn't really a Raven concern

PR #2627, opened 2026-07-17
Add Copilot code review instructions for docs-aware PR reviews

### LINKED ISSUE #2626: Add Copilot code review instructions for docs-aware PR reviews
GitHub Copilot code review already runs on our PRs. Adding a `.github/copilot-instructions.md` makes those reviews **docs-aware**: it applies a consistent Stellar-docs rubric (mechanics → technical-accuracy → completeness) with reference facts (current protocol/versions, exact network passphrases, RPC-over-Horizon, SDK naming) and ends each review with a clear **MERGE-READY / NEEDS-CHANGES / NEEDS-HUMAN-REVIEW** recommendation a maintainer can act on.

No secrets or new permissions needed — it only guides the Copilot review we already run. (Applying labels / gating merges is out of scope: Copilot code review can't do that; that would need a separate review bot.)

kaankacar:
Adds `.github/copilot-instructions.md` so the Copilot code review that already runs on our PRs reviews with a consistent Stellar-docs rubric and ends with a clear recommendation.

**What it does**
- Three-pass review: mechanics (relative links, anchors, MDX, sidebar collisions) → technical accuracy (protocol/version/date correctness, exact network passphrases, RPC-over-Horizon, SDK naming, SEP/CAP correctness) → completeness (does the diff fully do what it claims).
- Ends each review with **MERGE-READY / NEEDS-CHANGES / NEEDS-HUMAN-REVIEW** so it's easy to act on.

**Scope / notes**
- No secrets, no new permissions, no new workflow — it only shapes the existing Copilot review's comments.
- Copilot code review can't apply labels or block merges, so this is advisory (comment-only). Label/merge automation would need a separate review bot (separate discussion).

Closes #2626

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -11,8 +11,12 @@ ecosystem fact you cannot confirm from these sources, say so and recommend human
 than guessing.
 
 ## 1. Mechanics
-- Internal links must be **relative** (no hard-coded `https://developers.stellar.org/...` for
-  in-repo pages); every link and `#anchor` must resolve to a real file/heading.
+- Links in `docs/**` markdown must be **relative** — flag both hard-coded
+  `https://developers.stellar.org/...` and root-absolute `/docs/...` links to in-repo pages.
+  **Exception:** files under `openapi/` and `openrpc/` *should* use full
+  `https://developers.stellar.org/...` URLs. Every link should resolve to a real file; treat
+  `#anchor` resolution as best-effort (MDX slugs are hard to predict, and a broken anchor
+  won't fail the build).
 - Heading levels sequential; frontmatter `sidebar_position` must not collide with sibling
   pages; no slug/anchor collisions.
 - Images exist and have alt text; code fences declare a language; MDX compiles (no unclosed
@@ -48,11 +52,13 @@ Prefer this repo's own pages — they are the living source of truth and update 
 - `docs/networks/README.mdx` — network passphrases and network IDs.
 - `docs/data/apis/` — RPC and Horizon guidance / status.
 - `docs/tools/sdks/` — the SDK list and canonical package names.
-- `CONTRIBUTING`, `.github/` — repo conventions.
+- `.github/` and the area `CONTRIBUTING.md` files under `docs/` — repo conventions.
 
-Authoritative external sources for anything not settled in-repo:
-- Published docs: https://developers.stellar.org/docs
-- Protocol upgrade announcements: https://stellar.org/blog
+Authoritative external sources for anything not settled in-repo (the repo itself is the source
+of truth for docs content — don't fetch the published site, it's just a build of these files
+and can lag the PR):
+- Protocol upgrade announcements: https://stellar.org/blog — fetch the list page and page one
+  or two `?page=n` deep for recent posts; don't rely on the RSS feed (it's truncated).
 - SEPs and CAPs: https://github.com/stellar/stellar-protocol
 - Core / SDK / CLI releases: the relevant repo under https://github.com/stellar
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,8 +1,14 @@
 # Copilot code review instructions — stellar/stellar-docs
 
-You are reviewing documentation changes for the Stellar developer docs. Review every PR in
-three passes and end with a clear recommendation. Judge from the diff and the repo; never
-execute code from the PR. Be specific and cite `path:line`.
+You review documentation changes for the Stellar developer docs. Review each PR in three
+passes and end with a clear recommendation. Judge from the diff and the repository; never
+execute code from the PR. Cite `path:line`.
+
+**Do not rely on memorized facts about protocols, versions, dates, or releases — they change.**
+Verify every such claim against a current source (see "Reference sources" below), preferring
+this repository's own canonical pages, which are kept up to date here. If a PR states an
+ecosystem fact you cannot confirm from these sources, say so and recommend human review rather
+than guessing.
 
 ## 1. Mechanics
 - Internal links must be **relative** (no hard-coded `https://developers.stellar.org/...` for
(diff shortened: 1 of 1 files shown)


kaankacar:
For reviewers — confirmation this actually takes effect once merged:

- GitHub Copilot code review **automatically** uses repository custom instructions from `.github/copilot-instructions.md` (no setting to toggle), reading the file from the PR's **base branch**. References:
  - https://docs.github.com/en/copilot/concepts/agents/code-review — lists `copilot-instructions.md` as a customization source with activation *Automatic*.
  - https://docs.github.com/en/copilot/tutorials/customize-code-review — base-branch behavior ("Copilot uses the custom instructions in the base branch"), supported file types, ~1,000-line guidance.
- This repo already runs Copilot review automatically (the `Copilot Reviews` ruleset + recent PRs are auto-reviewed), so once this merges to `main` the rubric applies to every non-draft PR — no extra configuration needed.

Scope note: Copilot code review is advisory — it can't apply labels or block merges; this only shapes its review comments.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,68 @@
...
+
+**Do not rely on memorized facts about protocols, versions, dates, or releases — they change.**
+Verify every such claim against a current source (see "Reference sources" below), preferring
+this repository's own canonical pages, which are kept up to date here. If a PR states an
+ecosystem fact you cannot confirm from these sources, say so and recommend human review rather
+than guessing.
+
+## 1. Mechanics
+- Internal links must be **relative** (no hard-coded `https://developers.stellar.org/...` for
+  in-repo pages); every link and `#anchor` must resolve to a real file/heading.
```
ElliotFriend: two things might be worth softening here.

1. yes, links from markdown pages in `docs/` should always be relative. but, any links that are in the `openapi/` or `openrpc/` spec files should actually use the full `https://developers.stellar.org/...` format. also, it's worth identifying that we want "absolute" URLs excluded: no `[link to](/docs/some/absolute/path)` with the `/docs` part being the dead giveaway.
2. the `#anchors` might be a little tough to enforce/handle. not sure how well copilot will be able to identify/parse/whatever how the anchors get rendered by MDX. for example, `### Anchors (with Parenthesis) Might be Unpredictable` could hit a false-positive or something. plus, any `#anchor` isn't going to break the build, and it won't be _that_ detrimental to the user experience. though, maybe copilot is better at that than i'm assuming?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,68 @@
...
+- If it references or "fixes" an issue, does it FULLY resolve it?
+- Note sequencing constraints (a rename, or a companion PR in another repo) and related PRs.
+
+## Reference sources (consult these instead of memorized facts)
+Prefer this repo's own pages — they are the living source of truth and update over time:
+- `docs/networks/software-versions.mdx` — current protocols, activation dates, component versions.
+- `docs/networks/README.mdx` — network passphrases and network IDs.
+- `docs/data/apis/` — RPC and Horizon guidance / status.
+- `docs/tools/sdks/` — the SDK list and canonical package names.
+- `CONTRIBUTING`, `.github/` — repo conventions.
```
ElliotFriend:
```suggestion
- `CONTRIBUTING.md`, `.github/` — repo conventions.
```

would that make a difference, even? or would copilot be able to figure it out? also, the `CONTRIBUTING.md` file hasn't been merged yet, if that's important

.github/copilot-instructions.md
```diff
@@ -0,0 +1,68 @@
...
+
+Authoritative external sources for anything not settled in-repo:
+- Published docs: https://developers.stellar.org/docs
```
ElliotFriend: i mean, this is just a build of what is already in-repo. wouldn't this just result in extra tool calls and stuff that ends up returning almost exactly the same content?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,68 @@
...
+- Protocol upgrade announcements: https://stellar.org/blog
```
ElliotFriend: was trying to find a more "lean" version of getting this information for the agent. was thinking `https://stellar.org/blog/rss.xml` (but that's a short, fixed length of posts) or with `Accept: text/markdown` header (but that comes up with the same short list, with no real links to speak of) or using the developers blog, in particular (but not all posts get properly tagged).

i've come up short, and i think it might be worth specifying that the agent probably wants to get the actual text/html of the blog list page, and then page through at least one or two more `?page=n` params to get a sense of the latest posts that might be relevant. is it worth it, or (again) will copilot be able to just figure it out?

ElliotFriend (approved):
looks great! i left some notes/questions throughout, but nothing i'd really block on, if it's not necessary.

do we also want to include instructions about the "voice" or "tone" that writing should adhere to (e.g., second person, avoid editorializing, etc.)?

.github/copilot-instructions.md
kaankacar: scoped the relative-link rule to docs/** (now flagging /docs/... absolute paths too) with an explicit openapi//openrpc/ full-URL exception, and softened #anchor checks to best-effort since MDX slugs are unpredictable and a broken anchor won't fail the build.

.github/copilot-instructions.md
kaankacar: You're right that it 404s on main, so I dropped the root CONTRIBUTING reference for now and pointed to the area CONTRIBUTING.md files under docs/ instead

.github/copilot-instructions.md
kaankacar: Agreed, removed the published-docs source since it also can lag the PR, so the reviewer now reads the repo directly.

.github/copilot-instructions.md
kaankacar: Matches what I found too. I added a hint to fetch the blog list page and page a ?page=n or two deep rather than trusting the truncated RSS feed, while keeping the in-repo software-versions.mdx as the primary source.

PR #2639, opened 2026-07-18
🐛 Render the CodeExamples together

JFWooten4:
Correct formatting for code example in README.


ElliotFriend (approved):
good catch!

JFWooten4:
Not as good as parsing out the bug this weekend!

PR #2650, opened 2026-07-20
[chore] fix anchors page pagination

### LINKED ISSUE #2181: The product overview page in the anchor section is self-referential
When you're on the anchor docs under the ['product overview'](https://developers.stellar.org/docs/platforms/anchor-platform/admin-guide) page, the navigation at the bottom should ideally take you to the next page in series that is 'admin guide' but instead it links to itself

ElliotFriend:
fixes the strange pagination on the "frontpage" of the anchor platform section. applies `pagination_next` to the file's frontmatter, and points it at the admin guide section.

Closes #2181


kaankacar (approved):
Verified this on the preview build: Product Overview now advances to Admin Guide, and the Admin Guide page correctly points back to Product Overview and forward to Architecture. Confirmed the self-link still reproduces on prod, so this fully resolves #2181. Also checked that dropping the label from _category_.json changes nothing visible, the sidebar and breadcrumbs fall back to the README title. Same pattern we already use on the SDP front page. LGTM.

PR #2652, opened 2026-07-21
Rename token approve arg to live_until_ledger

leighmcculloch:
### What

Rename the `approve` function's allowance-expiration argument from `expiration_ledger` to `live_until_ledger` in the `TokenInterface` and `StellarAssetInterface` trait listings.

### Why

`rs-soroban-sdk` PR [#1932](https://github.com/stellar/rs-soroban-sdk/pull/1932), shipping in `soroban-sdk` v27.0.1, renamed this argument in both traits to align with the SEP-41 spec, since the argument name is part of the generated contract spec XDR. These docs mirrored the old SDK source verbatim, so they were left showing the outdated argument name.

### DIFF
--- a/docs/tokens/stellar-asset-contract.mdx
+++ b/docs/tokens/stellar-asset-contract.mdx
@@ -128,7 +128,7 @@ The [`token` module] of the Rust SDK contains two traits, and corresponding clie
 pub trait StellarAssetInterface {
     // All this is available in any SEP-41-compliant contract
     fn allowance(env: Env, from: Address, spender: Address) -> i128;
-    fn approve(env: Env, from: Address, spender: Address, amount: i128, expiration_ledger: u32);
+    fn approve(env: Env, from: Address, spender: Address, amount: i128, live_until_ledger: u32);
     fn balance(env: Env, id: Address) -> i128;
     fn transfer(env: Env, from: Address, to: MuxedAddress, amount: i128);
     fn transfer_from(env: Env, spender: Address, from: Address, to: Address, amount: i128);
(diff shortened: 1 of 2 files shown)


fnando (approved):

ElliotFriend:
@leighmcculloch thanks for this fix!!

looks like we're missing the note about when the SDK was changed, though. also, what's the ground truth about the events label? copilot seems to think events are still _intentionally_ labeled as `expiration_ledger`.

leighmcculloch:
> note about when the SDK was changed,

I don't think we  need to note in the docs when the SDK changed for this because it's not a breaking change. Variable names in traits and functions are essentially comments about the value and have no compile affect. If we do, it is soroban-sdk v27.0.1 when the change occurred and it was a bug fix, SEP-41 didn't change.

> what's the ground truth about the events label? copilot seems to think events are still _intentionally_ labeled as `expiration_ledger`.

The event in the soroban-token-sdk still uses the expiration ledger term, and we are discussing that here:
- https://github.com/stellar/rs-soroban-sdk/issues/1944

ElliotFriend:
@leighmcculloch i think you're generally right about not needing to call out all the bugfix versions throughout the docs. what i meant was your PR description says it includes:

> and add a note explaining the rename and which SDK version introduced it.

i don't see a note like that in the changes. does it still seem like something we want to have?

ElliotFriend:
also, do we want to ship the event field rename now? or, should we wait on this PR until the event has been updated in the SDK?

leighmcculloch:
> > and add a note explaining the rename and which SDK version introduced it.
> 
> i don't see a note like that in the changes. does it still seem like something we want to have?

Oh sorry I didn't update the PR description. No I don't think we need it.

The event rename is more complex and has to be timed so it won't happen immediately, or rather we are discussing over at:
- https://github.com/stellar/rs-soroban-sdk/issues/1944

Either way this doesn't need to wait for it.

PR #2664, opened 2026-07-21
📝 Use SDP acronym in API README

JFWooten4:


ElliotFriend:
i like the description you've added.

for the use of "SDP" over "Stellar Disbursement Platform" throughout the rest of the page, can you help me understand the why behind it? it seems a bit like it's an unnecessary abbreviation in the introductory material.

JFWooten4:
The abbreviation is introduced in the first sentence of the page. Repeating it three more times makes no sense to me. It adds verbosity, which makes the whole list harder to parse, a list that already repeats the full platform name twice in the repo names.

ElliotFriend (commented):
Thanks, John. The `description` frontmatter is a nice add — the page was missing a meta description, so that's a real improvement. The rest I'd reconsider: the frontmatter reordering is a no-op, and the "Stellar Disbursement Platform → SDP" swaps only hit 3 of the mentions (the intro/outro stay spelled out, and the "kubernetes" line is untouched), so it ends up less consistent than before rather than more.

If you trim it to just the `description` addition, I'm happy to take it. Also note this overlaps #2665 on `README.mdx` — they'll need coordinating.

JFWooten4:
I'm trying to clean up the frontmatter ordering. I think it looks a lot better alphabetically, plus I want to group categorically similar items like title and description together for easier future revisions.

Emir just reviewed 20 of my overlapping PRs in the Handbook for SCF, and I will admit I have a tendency to revise some of the same lines across PRs (there were seven conflicts during the merge order). But I've had an easy time syncing to main and keeping the diffs in order, so I'm not too worried about it.

Feel free to merge 2665 first, and then it's a simple sync of this branch. I am not really sure what you mean by the SDP acronym being less consistent when you use the defined term, as that was the point of this branch.

PR #2665, opened 2026-07-21
📝 Capitalize Kubernetes in documentation

JFWooten4:


ElliotFriend (changes requested):
Thanks, John. The capitalization itself is right — Kubernetes is a proper noun — but this needs to be scoped down a lot before it can go in, because most of the 130 files are generated output.

- **The 3 prose files are good as-is:** `running.mdx`, `upgrading.mdx`, and the Galexie `installing.mdx`. Those are the real fixes.
- **The SDP `*.api.mdx` and `*.StatusCodes.json` are generated** from the openapi spec (where "kubernetes" appears twice). Editing the generated output means it gets overwritten on the next `pnpm api`. Fix the two occurrences in `main.yaml`, run `pnpm api`, and commit the source + regenerated output together.
- **The Horizon `*.api.mdx` changes should come out entirely.** Horizon's OpenAPI source doesn't contain "kubernetes" at all — the only thing changing in those files is the encoded `api:` blob, i.e. regeneration noise unrelated to this PR.

That should take it from ~130 files down to a handful, all of which will actually persist. One more heads-up: this and #2664 both edit the SDP `README.mdx` and disagree on that same "kubernetes" line, so they'll conflict — worth deciding which one carries it.

JFWooten4:
API rebuilds limited to `platforms/stellar-disbursement-platform/api-reference` from the changed `yml`s, 83 files changed.

ElliotFriend (approved):
thanks for the fixes!!

PR #2666, opened 2026-07-21
Document constructor auth for register/register_at

### LINKED ISSUE #1738: `register_at` does not switch to recording auth for WASM constructors
### What version are you using?

Latest main branch (`soroban-sdk/src/env.rs`)

### What did you do?

Called `register_at(addr, WASM, constructor_args)` with a WASM contract whose constructor calls `require_auth`.

The `register_at` code path (`register_contract_with_contract_id_and_executable` at `env.rs:1502-1539`) calls the constructor without switching to recording auth:

```rust
// register_contract_with_contract_id_and_executable
// Used by register_at — NO auth switching
self.env_impl
    .call_constructor_for_stored_contract_unsafe(&contract_id, constructor_args.to_object())
    .unwrap();
```

Compare with the `register` code path (`register_contract_with_source` at `env.rs:1086-1089`) which does switch:

```rust
// register_contract_with_source
// Used by register — HAS auth switching
let prev_auth_manager = self.env_impl.snapshot_auth_manager().unwrap();
self.env_impl
    .switch_to_recording_auth_inherited_from_snapshot(&prev_auth_manager)
    .unwrap();
// ... constructor runs with recording auth ...
self.env_impl.set_auth_manager(prev_auth_manager).unwrap();
```

`register_at` should behave the same as `register` with respect to authorization during the constructor call. The two APIs are documented to differ only in whether the contract ID is auto-generated or user-supplied. The constructor should run under recording auth so that `require_auth` calls succeed.

### What did you see instead?

`register_at` calls the constructor under whatever auth mode is currently active. Any WASM constructor calling `require_auth` fails with an unexpected auth error when registered via `register_at`, while the same contract works when registered via `register`.

leighmcculloch:
### What

Add a new "Authorization in constructors" section to the Test Authorization guide explaining that both `env.register(...)` and `env.register_at(...)` switch the test environment to recording-auth mode for the duration of a Wasm contract's constructor invocation, so a constructor calling `require_auth()` succeeds without `mock_all_auths()`, and note the current caveat that this does not apply to native/mock contracts registered for testing.

### Why

As a result of the following issue and related changes its good to discuss this here:
- https://github.com/stellar/rs-soroban-sdk/issues/1738

### DIFF
--- a/docs/build/guides/testing/test-contract-auth.mdx
+++ b/docs/build/guides/testing/test-contract-auth.mdx
@@ -61,10 +61,10 @@ For the full example the above snippet is extracted from, see the [auth example 
 
 ## Authorization in constructors
 
-If a contract's constructor calls `require_auth()` (or `require_auth_for_args()`), both `env.register(...)` and `env.register_at(...)` switch the environment to recording-auth mode, inherited from the current auth manager, for the duration of the constructor invocation, then restore the previous auth manager afterward. This means constructor auth checks succeed for Wasm contracts registered with either function, without requiring `mock_all_auths()` to be called first.
+If a contract's constructor calls `require_auth()` (or `require_auth_for_args()`), both `env.register(...)` and `env.register_at(...)` switch the environment to recording-auth mode, inherited from the current auth manager, for the duration of the constructor invocation, then restore the previous auth manager afterward. This means constructor auth checks succeed for contracts registered with either function, without requiring `mock_all_auths()` to be called first.
 
 :::caution
 
-This only applies to contracts registered from Wasm bytes. If a contract is registered as a native/mock Rust value for testing (not from Wasm), its constructor invocation does not currently switch to recording auth. In that case, a `require_auth()` call in the constructor still requires `mock_all_auths()` (or explicit auth mocking) to be set up beforehand.
+This behaviour was not consistent in some versions of the `soroban-sdk` due to a issue, but is consistent as of v27.0.2.
 
 :::
--- a/docs/build/guides/testing/test-contract-auth.mdx
+++ b/docs/build/guides/testing/test-contract-auth.mdx
@@ -67,6 +67,6 @@ To have auth in a constructor execute as it will in production the contract must
 
 :::caution
 
-This behaviour was not consistent in some versions of the `soroban-sdk` due to a issue, but is consistent as of v27.0.2.
+This behaviour was not consistent in some versions of the `soroban-sdk` due to an issue, but is consistent as of v27.0.2.
 
 :::
--- a/docs/build/guides/testing/test-contract-auth.mdx
+++ b/docs/build/guides/testing/test-contract-auth.mdx
@@ -61,7 +61,7 @@ For the full example the above snippet is extracted from, see the [auth example 
 
 ## Authorization in constructors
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Heads up: rs-soroban-sdk#1943 merged today, fixing the remaining gap this section's caveat/version note doesn't yet account for — native (non-Wasm, directly-registered `#[contract]` struct) contracts now also get recording auth around the constructor for both `register` and `register_at`, matching the Wasm path fixed by #1933. Previously native constructors ran under enforcing auth regardless of which function registered them.

So as of the two merged fixes, the behavior is now uniform across register/register_at × Wasm/native — no native-specific caveat needed, and the "deploy in test using standard deployment functions after being built to Wasm" note in this PR's diff may no longer be accurate as a caveat for auth-in-constructors specifically.

Opened #2677 independently with wording that accounts for both fixes (and avoids guessing a specific unreleased version number, since v27.0.1 is still the latest published release). Feel free to take whichever of the two is preferred — happy to close #2677 if you'd rather fold this into #2666, or vice versa.

---

docs/build/guides/testing/test-contract-auth.mdx
```diff
@@ -58,3 +58,15 @@ fn test() {
 For the full example the above snippet is extracted from, see the [auth example contract](../../smart-contracts/example-contracts/auth.mdx).
 
 :::
+
+## Authorization in constructors
+
+If a contract's constructor calls `require_auth()` (or `require_auth_for_args()`), both `env.register(...)` and `env.register_at(...)` switch the environment to recording-auth mode, inherited from the current auth manager, for the duration of the constructor invocation, then restore the previous auth manager afterward. This means constructor auth checks succeed for contracts registered with either function, without requiring `mock_all_auths()` to be called first.
```
ElliotFriend: the first sentence here feels a little long and complex. something like this might offer a bit more clarity:


```suggestion
If a contract's constructor calls `require_auth()` (or `require_auth_for_args()`), both `env.register(...)` and `env.register_at(...)` switch the environment to recording-auth mode for the duration of the constructor invocation, inheriting the mode from the current auth manager. Once the constructor returns, the previous auth manager is restored. This means constructor auth checks succeed for contracts registered with either function, without requiring `mock_all_auths()` to be called first.
```

docs/build/guides/testing/test-contract-auth.mdx
```diff
@@ -58,3 +58,15 @@ fn test() {
...
+
+To have auth in a constructor execute as it will in production the contract must be deployed in the test using the standard deployment functions after being built to Wasm.
+
+:::caution
+
+This behaviour was not consistent in some versions of the `soroban-sdk` due to an issue, but is consistent as of v27.0.2.
```
ElliotFriend, line 70: since we're pointing out an issue that exists, should we make `an issue` a link to that github issue?

i'm not insisting on it, just asking if it might add some extra context for the curious.

ElliotFriend (commented):
this is a great addition!

i added a suggestion that might make the first part of this a little easier to read and understand. feel free to take it or adapt it.

thanks!!

docs/build/guides/testing/test-contract-auth.mdx
leighmcculloch: The issue has been resolved so I don't think it is useful to link to it.

ElliotFriend (approved):

PR #2679, opened 2026-07-23
🔧 Update vanity toolkit repo reference

JFWooten4:
This project has branches which expand into other languages for complex computations, hence it has dropped the py-prefix (inspired by Overcat)


ElliotFriend (approved):
Thanks for keeping this up to date!

PR #2687, opened 2026-07-26
Fix event-testing assertion in events example to use ContractEvents

leighmcculloch:
### What
Update the test's event assertions to use the SDK's current event-testing APIs (XDR comparison via `to_xdr`, `filter_by_contract`), and point the example link at the current version.

### Why
The old tuple-vec comparison no longer type-checks against the current SDK, so the example as shown would fail to compile.

### DIFF
--- a/docs/build/smart-contracts/example-contracts/events.mdx
+++ b/docs/build/smart-contracts/example-contracts/events.mdx
@@ -32,18 +32,18 @@ With the release of Whisk, Protocol 23, the syntax for publishing smart contract
 
 [open-in-github-codespaces]: https://github.com/codespaces/new?repo=stellar/soroban-examples&editor=web
 [open-in-code-anywhere]: https://app.codeanywhere.com/#https://github.com/stellar/soroban-examples
-[events example]: https://github.com/stellar/soroban-examples/tree/v23.0.0/events
+[events example]: https://github.com/stellar/soroban-examples/tree/main/events
 [storing data example]: ../getting-started/storing-data.mdx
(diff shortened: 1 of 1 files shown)


docs/build/smart-contracts/example-contracts/events.mdx
```diff
@@ -185,34 +185,33 @@ fn test() {
...
         env.events().all(),
-        vec![
-            &env,
-            (
-                contract_id.clone(),
-                (symbol_short!("COUNTER"), symbol_short!("increment")).into_val(&env),
-                1u32.into_val(&env)
-            ),
-        ]
+        std::vec![IncrementEvent { count: 1 }.to_xdr(&env, &contract_id)]
```
ElliotFriend, line 194: since we're using `std::vec![...]` here, shouldn't we include the `extern crate std;` somewhere at the top of this example? might trip up some users who are copy/pasting from the docs.

ElliotFriend: on second thought, we're also omitting the rest of the `use super::*;` kinds of stuff. since we're inviting users to "follow along" with an external source, this doesn't seem like a big deal to me.

ElliotFriend (approved):
LGTM! 🚀

docs/build/smart-contracts/example-contracts/events.mdx
leighmcculloch: We should be using just arrays here too, instead, as we don't actually need to use std::vec.

PR #2688, opened 2026-07-27
Fix test code in deploying-a-sac guide

leighmcculloch:
### What
Replace the SAC deployment test snippet's made-up`TestEnv`/`SacDeployer::default()`/string-literal `Bytes` calls with the actual test code required to register the contract, serializes a real XDR asset, and deploys it.

### Why
None of the original types or calls exist in the SDK, so the example couldn't be compiled as shown. Unclear where the examples came from, maybe this is from a misfired AI contribution.


ElliotFriend (approved):
thanks for the fix!

PR #2695, opened 2026-07-27
Fix type mismatch and stale tag in example-contract template

leighmcculloch:
### What
Make the template's example contract and test code use the same type consistently, and update its placeholder example-repo reference to the current version.

### Why
The docs page about the example is out of date with the actual example at:
https://github.com/stellar/soroban-examples/tree/main/hello_world


ElliotFriend (approved):
thanks for fixing this up!

PR #2672, opened 2026-07-22
🔗 Update Anchor Platform repository links

JFWooten4:
Replace the former java-stellar-anchor-sdk URLs with the current stellar/anchor-platform repository location. I think this would all look better with DocCards, but the page descriptions aren't set up for that yet.

PR #2701, opened 2026-07-27
🔗 Remove trailing slashes from Markdown links

JFWooten4:
Re: https://github.com/stellar/stellar-docs/pull/2675#pullrequestreview-4758321927 (point 1) and https://github.com/stellar/stellar-docs/pull/2672#pullrequestreview-4758404130

There were a few commits after #2409 since it was three months ago. That's why I tried to merge into main in #2583.

Styles are all aligned now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ These guidelines exist because they're the difference between a PR that merges i
 
 ## Development Quickstart
 
-We use [pnpm](https://pnpm.io/). See the [README](README.md#quick-start) for full setup.
+We use [pnpm](https://pnpm.io). See the [README](README.md#quick-start) for full setup.
 
 ```bash
 pnpm install        # install dependencies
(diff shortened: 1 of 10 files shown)


ElliotFriend:
Thanks for splitting these out, John, and sorry for the mixed signal.

My #2675 wording ("the standalone sweep, bundled into unrelated work") was ambiguous, you reasonably read the objection as the bundling, and Bri merging #2409 the week before muddied it further. That's on me.

But it sends us back to the question I asked on #2583 and never got an answer to: is there a source for the no-trailing-slash "rule," or is it more a personal preference? `trailingSlash: false` controls how Docusaurus generates output URLs, not how we author links in source, and #2409 had no description. I don't think we ever actually decided this.

So either:

- it's a convention, in which case it belongs in `CONTRIBUTING.md` with a lint rule so nobody has to hand-edit or review them again, or
- it isn't, and we leave existing links alone.

Both answers make this PR unnecessary, so I'm closing it as parked rather than reviewing it. Not a rejection.

Do you know where the rule originated? You're closer to this thread than I am, and if there's a real reason (SEO, redirects, something in the build) I'd rather hear it before deciding.

PR #2723, opened 2026-08-04
Distinguish OpenZeppelin Relayer from the managed Stellar Channels service

### LINKED ISSUE #2707: Tools pages present OpenZeppelin Relayer as an alias of the managed Stellar Channels service
Both `/docs/tools` and `/docs/tools/openzeppelin-relayer` describe OpenZeppelin Relayer as "also known as Stellar Channels Service" and as managed infrastructure. That wording presents two things as one product when their operational contracts differ.

The canonical [OpenZeppelin/openzeppelin-relayer](https://github.com/OpenZeppelin/openzeppelin-relayer) project is self-hostable and AGPL-3.0. Its operator runs it, holds signer authority, and funds the Stellar account that pays fees. A managed offering has a different answer for each of those.

Reproduce:

```sh
curl -Ls https://developers.stellar.org/docs/tools | grep -c -i 'Stellar Channels Service'
curl -Ls https://developers.stellar.org/docs/tools/openzeppelin-relayer | grep -c -i 'Stellar Channels Service'
```

Both return a non-zero count today. (Matching the full phrase can fail on the second page because an anchor tag splits it — count the product name instead.)

### Why it matters

A reader following these pages gets the wrong answer about who funds the relayer account and how it is topped up, while correctly citing official documentation. That is harder to catch than an obviously missing page, because the citation is genuine.

Your own [x402 page](https://developers.stellar.org/docs/build/agentic-payments/x402) already draws the distinction — it describes the plugin as using the Relayer framework while leveraging managed Channels underneath. The Tools pages have not adopted that separation.

### Prevalence

2 of 2 directly relevant Tools pages. We have not scanned the wider corpus, so this is a floor.

### Suggested fixes, cheapest first

1. Remove the "also known as" framing from both pages and name the managed offering and OpenZeppelin Relayer as separate concepts, each linking to its own source.
2. If the managed offering builds on Relayer, state that relationship precisely rather than as an alias.
3. Optionally add a short comparison covering who operates it, who holds signer authority, who funds fees, and what any status guarantee covers — those are the axes a reader is actually choosing between.

Version numbers, supported networks, and provider health are observation-time facts; dating them would prevent a stale reading from surviving as current guidance.

### What we do today

We verify the Tools description against the canonical repository before advising anyone on custody or fee funding — which rather defeats the purpose of consulting the official docs for a tools question.

ElliotFriend:
Both Tools pages called the Relayer "also known as Stellar Channels Service". They are not the same thing: the Relayer is an open source (AGPL-3.0) framework you can run yourself, where you hold the signing keys and fund the fee accounts, while Channels is the service OpenZeppelin operates on top of it, running the channel account pool and covering fees under a fair use policy.

Rewrites both blurbs and the page description, and notes that the examples on the Relayer page authenticate against managed Channels endpoints.

Closes #2707


ElliotFriend:
Good catch, addressed in 894ffb92. Renamed the managed-service references in prose to Channels: the status heading (it points at `status.channels.openzeppelin.com`), the hosted endpoint in both tutorials, and the Launchtube replacement sentence. "Relayer" now only refers to the framework, the `@openzeppelin/relayer-plugin-channels` package, and the upstream repo link. Code blocks are untouched, since they already use `ChannelsClient`.

ElliotFriend:
Both valid, fixed in 989b7c76. SDF was the subject of "provides", so it read as SDF operating Channels; now attributed to OpenZeppelin. Fair use caveat added to the index blurb.

ElliotFriend:
Fair, done in the latest commit. Renamed the comments, button and result strings, and the error text in the snippets; only `@openzeppelin/relayer-plugin-channels` still says relayer, since that is the package name.

ElliotFriend:
Right, fixed. Now reads "does not have the same maturity, scalability, or level of auditing as OpenZeppelin's Relayer". Pre-existing wording, but the sentence is in my diff so it may as well be correct.

xw-dd (approved):
Auto-approved: raven label

kaankacar:

Verified against OpenZeppelin's own docs: Channels is the managed service, OpenZeppelin runs the channel account pool, and the fees fall under a fair use policy. The Relayer repo is AGPL-3.0.

I pushed four corrections: parallel submission and fee bumping belong to the Channels plugin, Launchtube is archived so the transition is complete, the auditing comparison had no source, and the signer can be Google Cloud KMS or Turnkey. Ready to merge.

kaankacar (approved):

kalepail:
Raven independently verified this fix on 2026-09-09.
The live Tools pages and both serving indexes distinguish self-hosted Relayer from managed Channels.
The original production search returns corrected records for both affected pages.
Strict searches for the old alias return zero hits.

The September 9 crawl completed at `12:03:55.702Z`, after the source deployment.
Raven will retire the active finding after the final reference checks.

Immutable finding evidence: https://github.com/stellar-experimental/stellar-raven/blob/176513cc5e058fa12ad81f607ea4ca0d514a659a/improvements/stellar-docs/sd-039-openzeppelin-relayer-conflated-with-managed-channels.md

PR #2720, opened 2026-08-03
docs(tokens): distinguish SEP-41 from Stellar Asset Contract events

pedro-pelicioni:
Closes #2715.

The token interface page embeds the generic SEP-41 trait, whose event shapes are correct for a contract token and wrong for a Stellar Asset Contract. Nothing on the page said so, so a reader following it alone decodes a SAC event with the generic schema and mis-parses it. Every SAC topic vector carries a trailing `sep0011_asset` identifier.

## What this adds

A new `### Events: SEP-41 Tokens vs. Stellar Asset Contracts` section covering the seven SAC events, the issuer dispatch, and how to separate a classic operation from a contract invocation.

Three things it deliberately does not restate, linking instead:

- the `transfer` topic breakdown and the muxed data map, already in the payments guide;
- protocol-23 classic event emission, already in the events reference;
- `mint`, `clawback`, and `set_authorized`, already correct in the Stellar Asset Contract page's own trait.

The token interface page had no outbound link to the Stellar Asset Contract page at all, so the relationship was documented in one direction only.

## Two wording choices worth reviewing

**"Contract invocation", not "direct SAC".** The issue asks to distinguish direct SAC events from classic ones. The section says "contract invocation" instead, because a contract calling the asset contract on a user's behalf emits an identical event and is the same `InvokeHostFunction` operation. Operation type cannot separate those two, so the claim is scoped to what the metadata actually supports.

**`fee` gets a one-line caveat rather than a table row**, since it is transaction-level, carries no asset topic, and no contract function emits it.

## Second file

This also corrects `set_admin`'s event in `docs/tokens/stellar-asset-contract.mdx`, which omits the `sep0011_asset` topic and renders the data as a vec where the host emits a bare `Address`. Verified against `soroban-env-host` 27.0.1 `builtin_contracts/stellar_asset_contract/event.rs`, where all seven SAC topic constructors end their topic vec in `read_name(e)?`.

That is arguably a second concern. It is here because the new table states the correct shape and the neighbouring page would otherwise state a different one. Happy to split it out if you would rather keep this to one page.

## Verification

- `prettier --config .prettierrc.js -c` on both files: clean
- `docusaurus build`: succeeds, both pages processed, no warnings against either
- `scripts/check-relative-links.sh --range`: "All links check out"
- No routes added or removed

PR #2725, opened 2026-08-04
docs(tokens): cite CAP-0067 as the source for the SAC event shapes

pedro-pelicioni:
Follow-up to #2720, prompted by the deployment recheck in [stellar-experimental/stellar-raven#6](https://github.com/stellar-experimental/stellar-raven/issues/6).

Three of the four markers that finding tracked went from 0 to present on the page. `CAP-67` stayed at 0.

I had left it off on the reasoning that the section links [Tracking the movement of value](../learn/fundamentals/stellar-data-structures/events.mdx) rather than restating protocol-23 emission on a page about the generic interface. Checking it, that reasoning does not hold: **the linked page has zero `CAP-0067` occurrences of its own**, so a reader wanting to check the table against the spec has no path from either page.

Half the original reasoning does stand, which is why this is one sentence rather than a section. Restating classic-op emission mechanics belongs in the events reference, not here. Citing CAP-0067 as the normative source *for the table itself* is a different thing, and sd-018's recommendation names it that way — "CAP-0067 is the normative source for the SAC `transfer`/`burn` shapes and Classic unification."

## Verification

- `prettier -c`: clean
- `scripts/check-relative-links.sh --range`: "All links check out"
- `docusaurus build`: succeeds, page processed, no warnings against it
- Marker count on the page: `sep0011_asset` 7, `clawback` 3, `set_authorized` 3, `CAP-0067` 0 → 1
- No routes added or removed

### DIFF
--- a/docs/tokens/token-interface.mdx
+++ b/docs/tokens/token-interface.mdx
@@ -198,6 +198,8 @@ The event shapes in the interface above are the ones a SEP-41 contract token emi
 
 Code written against the shapes above will therefore mis-parse an event emitted by a Stellar Asset Contract, reading the asset identifier as a missing field or ignoring it entirely.
 
+[CAP-0067](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md) is the normative source for the Stellar Asset Contract shapes below, and for the unification of classic events described further down.
+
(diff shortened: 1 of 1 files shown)


ElliotFriend (approved):
good catch! thanks for the fixup!

PR #2726, opened 2026-08-04
Use the CAP-67 shorthand on the token interface page

ElliotFriend:
Follow-up to #2725, which wrote `CAP-0067`. The docs use the `CAP-NN` shorthand almost everywhere: `CAP-67` appears 22 times against 4 uses of the padded form, and the same pattern holds for CAP-46, CAP-66, CAP-71, and the rest. The link target keeps the padded `cap-0067.md` filename, since that is the real path in stellar-protocol.

It also matters for tooling. The Raven finding behind #2725 records its marker as `CAP-67`, so a literal recheck against the padded spelling reads as unresolved even though the citation is right there.

Scope is the one line. The other four padded references (`anatomy-of-an-asset.mdx`, `send-and-receive-payments.mdx`, `galexie/admin_guide/running.mdx`, `state-archival.mdx`, and a few more) are pre-existing and left alone.

### DIFF
--- a/docs/tokens/token-interface.mdx
+++ b/docs/tokens/token-interface.mdx
@@ -198,7 +198,7 @@ The event shapes in the interface above are the ones a SEP-41 contract token emi
 
 Code written against the shapes above will therefore mis-parse an event emitted by a Stellar Asset Contract, reading the asset identifier as a missing field or ignoring it entirely.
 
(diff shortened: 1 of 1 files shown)


pedro-pelicioni:
good catch, and the tooling half is the sharper argument. the recheck line in sd-018 is a literal grep for `CAP-67`, so `CAP-0067` closed the gap semantically while still reading as unresolved by the instrument's own measure, I picked the padded form because the spec filename is `cap-0067.md` and did not check it against the marker that defined the criterion.

keeping the padded form in the link target is the right split: the path is one thing, the prose convention is another.

ElliotFriend:
superseded by #2735

PR #2727, opened 2026-08-06
📝 Simplify README `MethodTable` introductions

JFWooten4:
From https://github.com/stellar/stellar-docs/pull/2672#discussion_r3633503673

There are a few (mostly API) pages that introduce MethodTables which contain functions. In introducing these lists of cards, they repeat the exact information and links in the actual table. This is redundant.

### DIFF
--- a/docs/data/apis/horizon/api-reference/errors/README.mdx
+++ b/docs/data/apis/horizon/api-reference/errors/README.mdx
@@ -7,7 +7,7 @@ import { MethodTable } from "@site/src/components/MethodTable";
 
 After processing a request, Horizon returns a success or error response to the client. A success response will return a Status Code of 200, and an error response will return a Status Code in the range of 4XX - 5XX along with additional information about why the request could not complete successfully.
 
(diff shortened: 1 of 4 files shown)


ElliotFriend:
this is a nice instinct, but there might be a problem. i don't think the `<MethodTable>` component (and its children) end up in the markdown that's served to AI agents. if you click the "view as markdown" button on one of those pages, with these changes, the file would end with `There are three types of error responses:` followed by nothing.

JFWooten4:
Good catch


That bug is similar to https://github.com/stellar/stellar-docs/pull/2624#issue-4911265152, and I think it warrants its own issue since these probably need a fix upstream with the embed button. I don't think it blocks this PR, as we should write the Developer Docs for human Developers first. Agents should be able to figure it out from the folder structure in the interim.

[Screenshot 2026-08-13 at 10 03 30 AM
](https://github.com/JFWooten4/stellar-docs/tree/6a419b4634039cf43e87740db3bfe583de25df69/docs/data/apis/horizon/api-reference/errors/http-status-codes)

PR #2728, opened 2026-08-06
📝 Simplify README `DocCardList` introductions

JFWooten4:
Extends https://github.com/stellar/stellar-docs/pull/2727

### DIFF
--- a/docs/tools/sdks/client-sdks.mdx
+++ b/docs/tools/sdks/client-sdks.mdx
@@ -2,7 +2,7 @@
 title: "Simplify Blockchain Development with SDKs for Java, Python, and More"
 sidebar_label: "Client & XDR SDKs"
 sidebar_position: 20
-description: "Explore the Stellar SDK library to simplify blockchain development. Leverage Rust, JavaScript, Python, and more to build on and integrate with the network."
+description: "Use Client and XDR SDKs in applications to interact with the Stellar network."
 ---
 
 # Client & XDR SDKs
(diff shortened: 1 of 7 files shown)


docs/tools/sdks/client-sdks.mdx
```diff
@@ -2,7 +2,7 @@
 title: "Simplify Blockchain Development with SDKs for Java, Python, and More"
 sidebar_label: "Client & XDR SDKs"
 sidebar_position: 20
-description: "Explore the Stellar SDK library to simplify blockchain development. Leverage Rust, JavaScript, Python, and more to build on and integrate with the network."
+description: "Use Client and XDR SDKs in applications to interact with the Stellar network."
```
leighmcculloch: This page's frontmatter `description` sets not just the parent `DocCardList` card blurb, but also the standalone page's own HTML meta/OG description (search-result snippet, social share preview).

Before: "Explore the Stellar SDK library to simplify blockchain development. Leverage Rust, JavaScript, Python, and more to build on and integrate with the network."
After: "Use Client and XDR SDKs in applications to interact with the Stellar network."

The new text is better for the SDKs README card, but it drops keyword-rich content ("Rust", "JavaScript", "Python") from this page's own SEO description — someone searching for those terms loses that match in the client-sdks.mdx search snippet.

Consider keeping the keyword-rich description on the page's own frontmatter and, if a shorter blurb is wanted for the card, using a different mechanism (e.g. `_category_.json`-style per-link description) rather than overloading the same field for both purposes.

---

PR #2735, opened 2026-08-06
📝 Simplify CAP reference formatting

JFWooten4:
https://github.com/stellar/stellar-docs/pull/2734

PR #2736, opened 2026-08-06
🖌 Document CAP & SEP reference style

JFWooten4:
https://github.com/stellar/stellar-docs/pull/2735


ElliotFriend (approved):

JFWooten4:
Glad to have seen Pedro's perspective on this. I had made this specific rule internally while running the #2145 transcripts (people don't say leading zeros). It is a good convention to inspire more ecosystem participation in standards and development, since choices feel less 'permanently formal' in enumeration (as, for example, happens with SEC filings, which always use leading zeros too).

PR #2739, opened 2026-08-10
🧶 Stub files still ref `yarn`

JFWooten4:
#2493

### DIFF
--- a/docs/tools/cli/cookbook/asset-management.mdx
+++ b/docs/tools/cli/cookbook/asset-management.mdx
@@ -4,4 +4,6 @@ hide_table_of_contents: false
 description: Issue a Stellar Asset, deploy its contract, and mint, burn, freeze, and clawback.
 ---
 
-Stub file: the real file is generated in build time by `pnpm stellar-cli:build`.
+## Stub File
+
+The real file is generated in build time by `pnpm stellar-cli:build`.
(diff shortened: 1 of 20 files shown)


ElliotFriend (commented):
this is a good fix. thanks for noticing and pushing it!

question, though: why change to a `## Stub File` heading instead of keeping the "inline" text like we had? doesn't really change anything, materially

JFWooten4:
I think it looks better like that, and I want to see how this displays in production before I propose a more [substantive change](https://github.com/JFWooten4/stellar-docs/blob/2beaae180e9a11ec85b70679ada742bb18757524/docs/tools/cli/cookbook/restore-contract-storage.mdx?plain=1#L7-L11). As happened when I started contributing in https://github.com/stellar/stellar-docs/pull/991 and https://github.com/stellar/freighter/pull/2784, new members may not understand the complete semantics of automatic workflows. We should make these technical features more apparent for a faster learning curve in https://github.com/JFWooten4/stellar-docs/issues/8

ElliotFriend:
yeah, i agree it should be obvious, but i don't think the `## Stub File` actually makes it any more obvious, tbh. the way a reader's eye takes in a file, it can be easy to skim over/past the headers. and, these headers will _never actually_ get rendered into production. the file contents are replaced by the build-time script. let's leave the `Stub file: the real ...` wording in-place.

ElliotFriend (approved):
lgtm! thanks

PR #2741, opened 2026-08-12
Update contract constructor documentation

leighmcculloch:
### What
Update Soroban documentation and examples to use `__constructor` deployment arguments and `deploy_v2` or `register` instead of post-deployment initialization.

### Why
The existing examples describe the pre-constructor initialization flow, which no longer matches current practices.

Related https://github.com/stellar/soroban-examples/pull/412

### DIFF
--- a/docs/build/guides/auth/check-auth-tutorials.mdx
+++ b/docs/build/guides/auth/check-auth-tutorials.mdx
@@ -78,11 +78,13 @@ const TRANSFER_FN: Symbol = symbol_short!("transfer");
 1. Initializes the contract with a list of signers' public keys
 2. Stores the count of signers in the contract's storage
 
+This is done in `__constructor`, which the host invokes once as part of the deployment itself. Setting the signers here rather than in a separate function called after deployment means the account can never exist in a signerless state that someone else could claim by front-running the call.
+
 ```rust
 #[contractimpl]
 impl AccountContract {
 
-    pub fn init(env: Env, signers: Vec<BytesN<32>>) {
+    pub fn __constructor(env: Env, signers: Vec<BytesN<32>>) {
         for signer in signers.iter() {
             env.storage().instance().set(&DataKey::Signer(signer), &());
         }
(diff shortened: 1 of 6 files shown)


docs/build/guides/auth/check-auth-tutorials.mdx
```diff
@@ -82,7 +82,7 @@ const TRANSFER_FN: Symbol = symbol_short!("transfer");
 #[contractimpl]
 impl AccountContract {
 
-    pub fn init(env: Env, signers: Vec<BytesN<32>>) {
+    pub fn __constructor(env: Env, signers: Vec<BytesN<32>>) {
```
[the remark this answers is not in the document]
leighmcculloch: 67e1f293a3a870478b0df5de389bc863c52a1fe3

docs/build/smart-contracts/example-contracts/bls-signature.mdx
```diff
@@ -320,8 +320,8 @@ fn test() {
     let pk = aggregate_pk_bytes(&env);
     env.mock_all_auths();
 
-    let client = create_client(&env);
-    client.init(&pk);
+    let contract_id = env.register(IncrementContract, IncrementContractArgs::__constructor(&pk));
```
ElliotFriend, line 323: question about this one: doesn't the "normal" register pattern go more like `env.register(IncrementContract, (&pk,));`?

I see that this change does indeed reflect the source code in the examples repo, but I'm curious why this pattern is used here, but not in other places?

ElliotFriend (approved):
this looks great! thank you for submitting the changes!

i left one question, but it's more a reflection of my own curiosity. certainly, it's not a blocker, and you are welcome to merge whenever you're ready.

docs/build/smart-contracts/example-contracts/bls-signature.mdx
leighmcculloch: The pattern was introduced or the feature was introduced at some time and people people can use either approach. The advantage of using the args is that it gives you a typed function to pass the values through so you know you're gonna get the types right and the number of parameters right whereas you're sort of winging it when you're   using a tuple but the tuple approach is very valid valid too

leighmcculloch: The pattern was introduced or the feature was introduced at some time and people people can use either approach. The advantage of using the args is that it gives you a typed function to pass the values through so you know you're gonna get the types right and the number of parameters right whereas you're sort of winging it when you're   using a tuple but the tuple approach is very valid valid too

leighmcculloch: I'm gonna leave this as is because this is what's used in the example that's being referenced. We should probably think about maybe normalizing the examples and the docs usage of it. Or you, know this, is also minor enough that it it maybe   doesn't matter.

PR #2743, opened 2026-08-13
Update software versions for p27 (Mainnet) and initial p28

ElliotFriend:
Protocol 27 has been needed on the [Software Versions](https://developers.stellar.org/docs/networks/software-versions) page for a while. Protocol 28 is upcoming.

This PR adds a mainnet entry for 27, and stubs in a protocol 28 entry for testnet, so it's ready when we are announcing the timeline.

### DIFF
--- a/docs/networks/software-versions.mdx
+++ b/docs/networks/software-versions.mdx
@@ -12,6 +12,41 @@ Release candidates are software releases that are also released to the [Testnet]
 
 [testnet]: ./README.mdx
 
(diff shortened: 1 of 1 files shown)


ElliotFriend:
There have also been some issues and PRs thinking about how to re-create and/or re-format this information into the most useful formats. This PR hasn't done _anything_ to address the different approaches on the table. It's only aim is to get the information out there, and ready.

sisuresh (approved):

PR #2746, opened 2026-08-14
Revise networks README with new link and friendbot notice

drQedwards:
Updated README to include new link and deprecation notice.

### DIFF
--- a/docs/networks/README.mdx
+++ b/docs/networks/README.mdx
@@ -7,6 +7,10 @@ description: "Explore the different networks, including Mainnet, Testnet, and Fu
 
 # Networks
 
+go to https://lab.stellar.org/account 
+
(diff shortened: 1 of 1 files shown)


ElliotFriend:
I'm not sure where you got the idea that friendbot is deprecated (unless i've missed some big announcement). it's a pretty foundational piece of the testnet and futurenet networks.

We definitely appreciate the contribution, but we can't land this one.

drQedwards:
This was a hallucination my tool went through without my knowledge so thank you for closing this!

I think some of my connectors locally werent fully "connected" and that's what caused the hallucination here.

I just responded to the current open PR relating to the languages in the docs and specifically on where React could be used while redacting the HTML5 and CSS3 bloat.

PR #2737, opened 2026-08-07
Add per-product audits pages

leighmcculloch:
### What

Add an Audits section under Networks, next to Software Versions, with a page per product — Stellar Core, Soroban Host Environment, Soroban Rust SDK — each holding a version-to-audit-report table.

Screenshot 2026-08-07 at 2 13 40 pm

### Why

Published audit reports aren't listed anywhere in the docs, so there's no way to tell which version of a product was audited or by whom. Tables are seeded with a placeholder row only, to be filled in with the real audit entries.

PR #2787, opened 2026-08-24
Add Stellar RPC audits page

sydneynotthecity:
Adds a per-product audits page for `stellar-rpc`, following the structure from #2737, carrying the OtterSec security assessment of stellar-rpc (assessment Nov 27 - Dec 11 2024, report dated Dec 20 2024; 1 medium, 2 low findings, all resolved via stellar-rpc#343, released in v22.1.2).

The report PDF is added under `static/assets/audits/` using the existing file naming pattern, and the route is registered in `routes.txt`.

Note: Ottersec listed a specific commit hash, `a60044e`, instead of a version tag for audit. This needs updating in the official report. The commit hash is not in the working tree of `stellar-rpc`, which means that the repo was snapshotted and given to Ottersec off the main, public repo. Through sleuthing, I was able to confirm that this commit aligns with `v22.1.0`, but we need to get official confirmation from Ottersec and identify the actual repo audited.


docs/networks/audits/stellar-rpc.mdx
```diff
@@ -0,0 +1,13 @@
+---
+sidebar_position: 20
+title: Stellar RPC
+description: Security audits of Stellar RPC.
+---
+
+# Stellar RPC
```
ElliotFriend: this won't really affect anything, but you can do away with this line altogether. most often, the `<h1>` markdown tag (e.g., `# Stellar RPC`) is used to override the frontmatter's `title` value. a page without a specified `h1`, will render with an `h1` containing the `title` value.

ElliotFriend (approved):
this looks great! thanks for putting it up here. i'll take care of the `h1` thing, but just wanted to share the context

PR #2807, opened 2026-09-01
Add USDT0 launch page at /launch/usdt0

mmazco:
Adds a standalone launch page for USDT0 on Stellar at `/launch/usdt0`, intended as a permanent URL.

The page ships as designed — its own layout rather than the docs theme — so it goes in `static/` rather than `src/pages/`. nginx's `try_files $uri $uri.html` resolves `static/launch/usdt0.html` at the extensionless path.

**No existing files change**, and nothing the Docusaurus build compiles is touched.

Two choices:

**Fonts are self-hosted** (`static/assets/launch/fonts/`). developers.stellar.org makes no third-party font requests today and this page shouldn't be the first. Inter is one variable file covering 400–600 rather than three copies of identical bytes; licences included per SIL OFL. 58KB on first load.

**GTM and GA4 are inlined** with the site's existing IDs (`GTM-M2JLH37`, `G-ZCT4GYX8KN`). Files in `static/` skip the build, so the page would otherwise report nothing. Same container as every other page, so consent handling comes with it.

Preview: https://mmazco.github.io/a-currency-in-motion/usdt0-atlas.html — same page, Google-hosted fonts, no analytics.

Content has had engineering review; contract addresses verified against usdt0 docs.


mmazco:
Flagging timing for whoever picks this up: the USDT0 announcement is **September 2nd**, so it would help a lot to have this reviewed before then.

Happy to make any changes asked for — the page is generated from a single source, so edits are quick and land as a new commit on this branch.

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+    { label:"One asset, unified liquidity",
+      copy:"USDT0 maintains a unified supply anchored to USDT on Ethereum, so liquidity isn&rsquo;t fragmented across separate wrapped versions of the asset." },
+    { label:"Built to move across networks",
+      copy:"Operated by <a href='https://everdawn.to/' target='_blank' rel='noopener noreferrer'>Everdawn Labs</a>, USDT0 connects Stellar to <a href='https://analytics.usdt0.to/' target='_blank' rel='noopener noreferrer'>26+ networks</a> through LayerZero&rsquo;s OFT standard." }
+  ]
+};
+
+// Reading order across the section: the stack, the columns and the notes all
+// follow this, so priority is changed in one place.
+const ORDER = ["Payments","Treasury","Collateral","Distribution"];
```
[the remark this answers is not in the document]
ElliotFriend: this is definitely worth fixing

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+  <div class="top">
+    <span>USDT0 on Stellar Network</span>
+    <nav class="secnav" aria-label="Sections">
+      <a class="tick" href="#arrival" aria-label="Arrival" title="Arrival" aria-current="true"></a>
+      <a class="tick" href="#access" aria-label="Access" title="Access"></a>
+      <a class="tick" href="#dayone" aria-label="Day one" title="Day one"></a>
+      <a class="tick" href="#deploy" aria-label="Deploy" title="Deploy"></a>
+      <a class="tick" href="#build" aria-label="Before you build" title="Before you build"></a>
+    </nav>
+  </div>
```
ElliotFriend, line 525: i'd love to see if we could include a link to the rest of the stellar documentation in this header? maybe right-aligned but to the left of the nav squares?

we could even have that link go directly to the new USDT0 page we published recently

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+    <path class="fo-edge" d="${FO_BODY}"/>
+  </svg>`;
+
+// the stack is now a picture, not a control — every file is written out below it
+// generated, so it cannot drift out of step with ORDER the way a hand-written
+// label did
+stack.style.setProperty('--ar', String(AR0 * KY / KX));
+stack.style.setProperty('--fw', (FW0 * KX) + '%');
+stack.style.setProperty('--fh', (FH0 * KY) + '%');
+document.getElementById('files').style.setProperty('--cols', String(N));
```
ElliotFriend: it's fine to have this as a picture, but currently the "file tabs" are still click-able and give the impression that something should happen. looks like the CSS around line 245 is to blame for that.

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+      <div class="notes" id="notes"></div>
+    </div>
+  </section>
+
+  <dl class="spec">
+    <div><dt>Asset</dt><dd>USDT0</dd></div>
+    <div><dt>Transfer</dt><dd><a href='https://usdt0.to/transfer' target='_blank' rel='noopener noreferrer'>usdt0.to</a></dd></div>
+    <div><dt>Networks</dt><dd><a href='https://analytics.usdt0.to/' target='_blank' rel='noopener noreferrer'>26+</a></dd></div>
+    <div><dt>USDT in circulation</dt><dd><a href='https://coinmarketcap.com/currencies/tether/' target='_blank' rel='noopener noreferrer'>$180B+</a></dd></div>
+    <div><dt>Stellar settlement</dt><dd><a href='https://developers.stellar.org/docs/learn/fundamentals/stellar-stack' target='_blank' rel='noopener noreferrer'>~5s</a></dd></div>
```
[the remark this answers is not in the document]
ElliotFriend: agreed. we're linking to a page here that does say 5-7 seconds, we should use the same language here

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+    <div><dt>Stellar fee</dt><dd><a href='https://developers.stellar.org/docs/learn/fundamentals/fees-resource-limits-metering' target='_blank' rel='noopener noreferrer'>&lt;$0.001</a></dd></div>
+    <div><dt>Cross-chain transfer</dt><dd><a href='https://docs.usdt0.to/technical-documentation/developer/' target='_blank' rel='noopener noreferrer'>30s&ndash;3min</a></dd></div>
+  </dl>
+
+  <div class="sources">
+    <b>Sources</b>
+    <a href="https://docs.usdt0.to/technical-documentation/developer/" target="_blank" rel="noopener noreferrer">USDT0 developer docs</a>
+    <a href="https://docs.usdt0.to/technical-documentation/deployments?q=stellar" target="_blank" rel="noopener noreferrer">USDT0 deployments</a>
+    <a href="https://developers.stellar.org/docs/tokens/usdt0-layerzero" target="_blank" rel="noopener noreferrer">Stellar Docs</a>
```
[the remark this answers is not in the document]
ElliotFriend: this is fixed now.

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+    caveat:"<b>A trustline has to exist before the first payment lands.</b> Sending USDT0 to an account that has no USDT0 trustline fails with <a href='https://developers.stellar.org/network/horizon/errors/result-codes/operation-specific/payment' target='_blank' rel='noopener noreferrer'><code>op_no_trust</code></a>. A wallet may create it for you when you swap into USDT0 in-app; a transfer arriving from an exchange or another wallet has no such fallback. If you custody for users, create it during onboarding &mdash; <a href='https://developers.stellar.org/docs/build/guides/transactions/sponsored-reserves' target='_blank' rel='noopener noreferrer'>sponsored reserves</a> can cover the reserve."
```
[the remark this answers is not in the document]
ElliotFriend: we do have a redirect that "should" catch the old-style URL, but let's switch to the new, working URL anyway.

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>USDT0 on Stellar</title>
+<meta name="description" content="USDT0 is available on the Stellar network. How it arrives, what it unlocks, the contract addresses, and what to check before you build.">
+<meta property="og:type" content="article">
+<meta property="og:title" content="USDT0 on Stellar">
+<meta property="og:description" content="USDT0 is available on the Stellar network. How it arrives, what it unlocks, the contract addresses, and what to check before you build.">
+<meta name="twitter:card" content="summary">
```
ElliotFriend: do we want one of those `og:image` meta tags?

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+        View the asset on stellar.expert &rarr;</a>
+
+      <div class="ids" id="ids"></div>
+
+      <a class="dep-link" target="_blank" rel="noopener noreferrer"
+         href="https://docs.usdt0.to/technical-documentation/deployments?q=stellar">Contract deployments on docs.usdt0.to &rarr;</a>
+
+    </div>
+    <div class="dep-note"><b><a href='https://stellar.expert/explorer/public/asset/USDT0-GATISXX6BZ6NC7IKQBY37CJD4SOZL3CYZJWXEDG6JVIY4WBS6KXJHN6Q' target='_blank' rel='noopener noreferrer'>This asset</a> sets <code>revocable</code> and <code>clawback_enabled</code></b> &mdash; a trustline can be revoked or a <a href='https://developers.stellar.org/docs/build/guides/transactions/clawbacks' target='_blank' rel='noopener noreferrer'>balance clawed back</a>. <code>auth_required</code> is not set, so anyone can hold it. &#128161; Soroswap maintains an <a href='https://github.com/soroswap/token-list/tree/main/assets' target='_blank' rel='noopener noreferrer'>ecosystem token list</a> covering both classic assets and contract tokens (community-maintained rather than an official Stellar registry, and widely used).</div>
```
ElliotFriend: these flag names are missing the `auth_` prefix.

```suggestion
    <div class="dep-note"><b><a href='https://stellar.expert/explorer/public/asset/USDT0-GATISXX6BZ6NC7IKQBY37CJD4SOZL3CYZJWXEDG6JVIY4WBS6KXJHN6Q' target='_blank' rel='noopener noreferrer'>This asset</a> sets <code>auth_revocable</code> and <code>auth_clawback_enabled</code></b> &mdash; a trustline can be revoked or a <a href='https://developers.stellar.org/docs/build/guides/transactions/clawbacks' target='_blank' rel='noopener noreferrer'>balance clawed back</a>. <code>auth_required</code> is not set, so anyone can hold it. &#128161; Soroswap maintains an <a href='https://github.com/soroswap/token-list/tree/main/assets' target='_blank' rel='noopener noreferrer'>ecosystem token list</a> covering both classic assets and contract tokens (community-maintained rather than an official Stellar registry, and widely used).</div>
```

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+  { label:"Token contract (SAC)",
+    note:"The Stellar Smart Contract (Soroban) interface. Call this from contracts — lending markets, vaults, AMMs.",
+    value:"CBSJZEIO5C7KC2SF3MKSNXXJSW5G3VTNBX4ATMKUI3B2MR4JKM4R26YF" },
+  { label:"OFT",
+    note:"The USDT0-specific OFT contract on Stellar &mdash; it burns and mints USDT0 here. Not part of a Stellar-side payment.",
+    value:"CBOWOLFSDM5PZXNFIVDMP5NZ7U2GSIHED6H6R446QOHF266XINKUMMF6" },
+  { label:"LayerZero endpoint ID",
+    note:"Stellar's EID. You need it to address Stellar from another chain's OFT.",
+    value:"30600" }
+];
```
ElliotFriend: having this information live _only_ in the javascript carries some risk with it. in a browser with javascript disabled, or on a search engine that doesn't execute JS, these contract addresses and other info wouldn't ever render. might be worth hard-coding that kind of really important info directly into the HTML

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+    </div>
+    <div class="fig-files">
+      <div class="files" id="files"></div>
+    </div>
+  </figure>
+
+  <!-- ============ DAY ONE ============ -->
+  <section class="dayone" id="dayone" aria-labelledby="d1-h">
+    <div class="dep-head">
+      <span id="d1-h">Day one &mdash; what is already integrated</span>
```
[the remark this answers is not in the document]
ElliotFriend: if you wanted to carry this out, you would just replace the `<span>` with `<h2>` leaving all its attributes alone. it's not as bad as copilot makes it seem, but the fix is relatively easy. take it or leave it

```suggestion
  <section class="dayone" id="dayone" aria-labelledby="d1-h">
    <div class="dep-head">
      <h2 id="d1-h">Day one &mdash; what is already integrated</h2>
```

static/launch/usdt0.html
```diff
@@ -0,0 +1,1013 @@
...
+// No stored choice means follow the OS. Clicking sets an explicit override,
+// which the :root[data-theme] blocks pick up over the media query.
+(function(){
+  const root  = document.documentElement;
+  const KEY   = 'usdt0-atlas-theme';
+  const btn   = document.getElementById('themetog');
+  const mq    = window.matchMedia('(prefers-color-scheme: dark)');
+
+  const store = {
+    get(){ try { return localStorage.getItem(KEY); } catch (e) { return null; } },
```
ElliotFriend: if you add this fallback, we'll inherit the docusaurus theme value (dark or light) if a user has already been visiting the site, and has set their preference. (note: in some instances, this _might_ result in a white "flash" for some users as the page renders, but not all that likely. avoid that by moving the `saved = store.get()` read into a small inline script in the head, above the stylesheets.) this is also a take-it-or-leave-it kind of thing

```suggestion
    get(){ try { return localStorage.getItem(KEY) || localStorage.getItem('theme'); } catch (e) { return null; } },
```

static/launch/usdt0.html
ElliotFriend: beside that: the details here (addresses, EID, etc.) all check out! 👍🏻

ElliotFriend (commented):
overall, this is looking really good. i left some comments here 'n there. not many are deal-breakers for me.

couple broader questions to ask/consider:

- do we know who will be updating this page as time goes on? there are some stats, partners, etc. that are likely to shift/expand over time, and we probably want to keep up with that.
- no problem with styling things different from the rest of the docs for these sorts of "one-off" pages. if this grows into more of a "collection" of pages, should we come up with some kind of common design library/style specifically for them? that's not a today question, but something to put on the radar for the future.
- since this is a static HTML file, why not embed the data bits into the markup? in a web fetch or a JS-disabled environment, fairly significant portions of the page render as empty boxes (contract addresses, the "file tabs," etc.) this can materially affect how well agents read the content, too. this may be an improvement we can make after the thing ships, but it's worth thinking about.

thanks again for getting this prepared!

mmazco:
Thanks Elliot — really useful review. Pushed fixes in 281b89a6.

**Corrections you caught:**
- `auth_revocable` / `auth_clawback_enabled` — you're right, the flags carry the `auth_` prefix. Applied your suggestion.
- `op_no_trust` link switched to the generated Horizon route; verified it returns 200.
- Settlement now says 5–7s to match the page we link to.
- `<$0.001` replaced with the 100 stroop minimum — agreed that a USD figure implies a ceiling the protocol doesn't promise and drifts with the XLM price.
- File tiles no longer carry `cursor:pointer`; they're a picture and shouldn't suggest otherwise.
- Section titles are now `h2` and venue names `h3`, so there are no skipped heading levels. They inherit the band typography, so nothing moved visually.
- Copy buttons now name their row rather than four identical "Copy"s.
- Added the Stellar Docs link in the header, pointing at the new USDT0 page.
- Took the theme fallback too — inherits the Docusaurus preference when one is already set.

**On `ORDER` and Liquidity** — this one's worth flagging, because both you and Copilot read it as an accidental omission and it wasn't. The Liquidity column was cut during content review, so `ORDER` excludes it on purpose. And the `LAYERS.Liquidity` entry isn't dead: `byTag("Liquidity")` is what supplies the Backing note in "Before you build". Adding Liquidity back to `ORDER` would have resurrected a column we deliberately removed.

What *was* genuinely dead was the `FILES.Liquidity` entry — unreachable, since `inOrder()` filters through `ORDER`. That's removed, and there's now a comment explaining the arrangement so it doesn't get re-reported.

**Your broader questions:**

1. **Ownership** — I'll own updating the stats and partner lists going forward.

2. **Shared style library** — agreed it's worth thinking about, though I'd frame it slightly differently than "one-off pages". The useful version would be for developer-facing release and preview pages we build as artifacts. Something to noodle on rather than solve now.

3. **Hard-coding the data into markup** — agree, and more strongly than "nice to have". With JS disabled, Access, the contract addresses and all of "Before you build" render as empty boxes, which is most of the page's value — and as you say, it materially affects how agents read it, which matters for a developer page. Considering doing it in this PR rather than a follow-up, since it's a mechanical transform and a second PR costs you another review round. Will confirm shortly either way.

`og:image` is coming — you're right that it's missing, and the page is about to be shared on social, so it needs one.

Thanks for checking the addresses and EID.

ElliotFriend (approved):
looks great!! thanks for getting this polished up and ready to go!! 🎉

PR #2808, opened 2026-09-01
allow pr checkout in action

ElliotFriend:


kaankacar:

Workflow change, so a human decides. It opts pr-preview into unsafe fork PR checkout.
Verified: checkout v6 blocks the fork case today (failed run 33930501945), and the gate already requires the author, the actor and the head-repo owner to all be org members.
Please confirm the opt-in. @kaankacar @ElliotFriend

PR #2809, opened 2026-09-01
Test PR to trigger PR previews

satyamz:


kaankacar:

The preview test passed: the `preview` check is green and the preview URL serves the site.
The only change left is a stray double space in the README heading, so nothing here should land on main. Close candidate, like the earlier #2256.

PR #2814, opened 2026-09-03
docs: add Soroscan to block explorers

fxjrin:
Adds Soroscan to the block explorers list. Soroscan is a Stellar block explorer covering Mainnet, Testnet, and Futurenet, with a built-in AI assistant for on-chain questions. Live at https://soroscan.io.

### DIFF
--- a/docs/tools/developer-tools/block-explorers.mdx
+++ b/docs/tools/developer-tools/block-explorers.mdx
@@ -27,7 +27,7 @@ Explore transactions, accounts, assets, contracts, and ledgers, with a built-in 
 
 :::note
 
-Soroscan is a newer explorer under active development. A few things we're still working on:
+Soroscan is a newer explorer under active development. A few things the Soroscan team is still working on:
 
 - Richer analytics and network insights
 - Deeper detail across account, contract, and asset views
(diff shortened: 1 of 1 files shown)


ElliotFriend:
hi, @fxjrin! thanks for contributing, and thanks for creating a block explorer!

i like the look and feel of what you've put together, and think it could fit well alongside our other explorers. from the git history, the project looks to be relatively young (about 2 weeks).

i'd love if the listing here would include a bit of a "note" that it's a newer project, and maybe outline the current limitations? not in a "here are things we can't do" way, but more like "here are the things we're still working on." does that make sense?

fxjrin:
> hi, @fxjrin! thanks for contributing, and thanks for creating a block explorer!
> 
> i like the look and feel of what you've put together, and think it could fit well alongside our other explorers. from the git history, the project looks to be relatively young (about 2 weeks).
> 
> i'd love if the listing here would include a bit of a "note" that it's a newer project, and maybe outline the current limitations? not in a "here are things we can't do" way, but more like "here are the things we're still working on." does that make sense?

Thanks, @ElliotFriend! that makes sense, I've added a short note to the Soroscan entry saying it's a newer project and outlining a few things we're still working on. Happy to adjust the wording if needed.

ElliotFriend (approved):
nice work! thanks 🎉

PR #2823, opened 2026-09-07
Document CRITICAL validator quality

### LINKED ISSUE #985: Incomplete description of automatic Quorum Set generation

kaankacar:
Automated triage bot, acting for the docs maintainers.

Closes #985

`QUALITY` accepts four values, but this page listed three. The diff adds `CRITICAL` to both field tables, gives it its own subsection, and corrects the automatic quorum set generation rules, where every `CRITICAL` entity is required. It also corrects the `HIGH` requirements, because stellar-core enforces a three-validator redundancy as well as the archive. Every statement here comes from `src/main/Config.cpp` and `docs/stellar-core_example.cfg` at stellar-core v28.0.1.

Two notes for the reviewer:

- The leader election section needs no change. It defines the top weight as the highest quality assigned to any organization, so it already covers `CRITICAL`.
- `validator_complete.png` still shows three levels. I cannot regenerate the image, so the caption now states that the diagram shows a quorum set with no `CRITICAL` entity.

### DIFF
--- a/docs/validators/admin-guide/configuring.mdx
+++ b/docs/validators/admin-guide/configuring.mdx
@@ -226,7 +226,7 @@ A critical quality validator has the same programmatically enforced requirements
 - publishes a history archive, and
 - belongs to a home domain with at least three validators.
 
-Reserve this rating for an entity you cannot afford to disagree with. Because a critical entity is always required, your node stops closing ledgers when that entity drops below a simple majority of its own validators. In other words, this rating trades liveness for safety, so `HIGH` is the right choice for most quorum sets.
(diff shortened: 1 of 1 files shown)


kaankacar:

Verified against stellar-core v28.0.1. `CRITICAL` is a real `QUALITY` value. The archive and three-validator rules are enforced for `HIGH` and `CRITICAL`.

I pushed two fixes. A `CRITICAL` entity also makes the nested lower quality group required. Copilot was right that "most quorum sets" was unverifiable, so that line is now conditional.

`build` and `mdx-format` pass. The preview host 404s after the last push, so the build check is the render evidence. Ready to merge.

ElliotFriend (approved):
lgtm!

PR #2825, opened 2026-09-07
Document which contract errors try_call can recover from

### LINKED ISSUE #1074: Document recoverable versus nonrecoverable contract errors
### What
Document recoverable versus nonrecoverable contract errors to assist contract developers. We can provide some guidance of which scenarios one would be beneficial than the other. 
This can possibly go in [Errors](https://developers.stellar.org/docs/build/smart-contracts/example-contracts/errors) page

kaankacar:
Triage bot, acting for @kaankacar.

Closes #1074

Two pages said `try_` catches every error from a contract call. It does not. `HostError::is_recoverable` in `rs-soroban-env` v28.0.2 makes two groups non-recoverable, and `try_call` re-raises those, so the guest traps and the transaction fails. This corrects both sentences and adds one section that states the rule, on the page that already defines `call` and `try_call`. @dmkozh named the same two groups in the issue.

Non-recoverable, from `soroban-env-host/src/host/error.rs`:

- code `ExceededLimit` with type `Budget` or `Storage`
- code `InternalError` with any type other than `Contract`

The `Storage` case is the footprint violation. The issue calls it the one non-obvious case, so it gets its own paragraph.

Two notes. This adds no page and no new URL, so it does not answer the "common gotchas page" idea in the issue. Open PR #2732 edits the `<head>` block of the same overview page, four lines from my hunk.

### DIFF
--- a/docs/build/guides/conventions/cross-contract.mdx
+++ b/docs/build/guides/conventions/cross-contract.mdx
@@ -87,7 +87,7 @@ client::ContractAEnum::SomeField
 
 In the examples above, we have used `env.invoke_contract` and `client.some_function`. In both cases, if there is an issue with the underlying contract call, the contract will panic. This might be a valid approach, but in some cases we want to catch errors and handle them depending on the outcome. This allows you to forward a custom error message, or even trigger an alternative code path.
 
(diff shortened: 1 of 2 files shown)


kaankacar:

Verified against rs-soroban-env v28.0.2: `HostError::is_recoverable` and `try_call` match both non-recoverable groups exactly. The preview renders and the new anchor resolves from the cross-contract caution. Ready to merge.

ElliotFriend (approved):

PR #2840, opened 2026-09-08
Make /launch/usdt0 discoverable to search and AI crawlers

mmazco:
`/launch/usdt0` has been live since 1 Sep and is drawing real organic traffic — Google is already its second-largest source, and ChatGPT and Claude have both begun referring to it. Three things were limiting how findable it is. Two are mechanical; one is a bug.

The fourth issue, that nothing on the site links to the page, is handled separately in #2841 so it can land independently of the config change here.

### 1. The page is not in the sitemap

The generated sitemap has 960 URLs and none of them is this page. Files under `static/` bypass the Docusaurus build entirely, so nothing adds them.

Fixed with the sitemap plugin's `createSitemapItems` hook, so the entry is generated alongside every other URL rather than maintained by hand in a second file.

> ⚠️ **Needs CI to verify.** I could not install dependencies locally, so this config change is unbuilt. It is deliberately minimal — one extra item, no `changefreq` or `priority` — but a maintainer should confirm the build passes and that `/launch/usdt0` appears in `sitemap.xml`.

### 2. `canonical` and `og:url` have never shipped — a silent bug

The generator that produces `static/launch/usdt0.html` was written to inject both tags, anchored on:

```
<meta name="twitter:card" content="summary">
```

That string became `content="summary_large_image"` when the OG image was added, and `str.replace` fails silently when it matches nothing. Neither tag has been on the live page since.

This matters more than usual here: `/launch/usdt0`, `/launch/usdt0.html` and `/docs/launch/usdt0` are all currently receiving traffic as separate URLs, with no canonical consolidating them.

### 3. No structured data

Added a `TechArticle` JSON-LD block — headline, description, canonical URL, publish date, publisher, subject. This gives search and AI systems something explicit to cite rather than inferring what the page is.

### Already fine, not changed

- `robots.txt` allows all crawlers and sets `Content-Signal: ai-train=yes, search=yes, ai-input=yes`
- Page content is fully hard-coded in markup — nothing requires JavaScript to read
- Clean heading hierarchy: one `h1`, five `h2`s

Content updates to the page are in progress separately.

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -154,6 +154,21 @@ const config: Config = {
         googleTagManager: {
           containerId: "GTM-M2JLH37",
         },
+        sitemap: {
+          // Files under static/ bypass the Docusaurus build, so pages served
+          // from there never appear in the generated sitemap. /launch/usdt0 is
+          // a permanent page and was invisible to crawlers without this.
+          createSitemapItems: async ({
+            defaultCreateSitemapItems,
+            ...rest
+          }) => {
+            const items = await defaultCreateSitemapItems(rest);
+            return [
+              ...items,
+              { url: "https://developers.stellar.org/launch/usdt0" },
+            ];
+          },
+        },
       } satisfies Preset.Options,
     ],
   ],
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -165,7 +165,11 @@ const config: Config = {
             const items = await defaultCreateSitemapItems(rest);
             return [
               ...items,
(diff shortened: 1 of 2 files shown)


ElliotFriend (approved):
looks great! i'm adding a commit that does two things:

- includes those `changefreq` and `priority` to match the sitemap plugin's defaults which get included in the other entries.
- also reading the already-written URL from the siteconfig instead of hard-coding it again. (very minor nit there)

i confirmed the PR behaves the way we want in the local build. once the build passes here in CI i'll get this merged. thanks!!

PR #2841, opened 2026-09-08
Link the USDT0 docs page to the USDT0 launch guide

mmazco:
Adds one link from `docs/tokens/usdt0-layerzero.mdx` to `/launch/usdt0`.

`/launch/usdt0` currently has no inbound links from anywhere on developers.stellar.org. This page is its most natural referrer: a reader here is already looking at USDT0 on Stellar, and the launch guide covers what it unlocks, which partners support it today, and a deployment checklist — none of which this page goes into.

It also helps crawlers, which reach orphaned pages less often and pass them no internal link equity.

The link uses descriptive anchor text rather than "read more", so both readers and crawlers can tell where it goes.

One line, no other changes. Split out from #2840 so it can land independently of the sitemap configuration change there, which needs a build to verify.


ElliotFriend (approved):
i pushed a commit that uses the `pathname://` protocol, which will avoid the docusaurus build process, and won't throw the error at build-time.

PR #2850, opened 2026-09-11
Document the --enable-core-manual-close flag in Quickstart advanced usage

### LINKED ISSUE #2772: Quickstart documentation omits the shipped --enable-core-manual-close flag
> [!NOTE]
> **Automated notice:** [Stellar Raven](https://github.com/stellar-experimental/stellar-raven) filed this issue from its automated evaluation pipeline. The issue includes evidence and a public source record. Please verify the live surface before you act.

## Finding

The Quickstart container supports a manual ledger-close mode. The
`stellar/quickstart` `start` script parses `--enable-core-manual-close` and
writes the value into the generated `etc/stellar-core.cfg`.

Stellar Docs do not document the flag. The site holds a full Quickstart
section at `/docs/tools/quickstart/`, including `advanced-usage/container`,
`advanced-usage/operation-modes`, and `advanced-usage/run-command-examples`.
None of those pages names the flag. A repository-wide search of
`stellar/stellar-docs` for the exact string returns zero results.

Docs search cannot recover the flag either. A live search for the flag name
returns only ledger-ingestion pages.

Deterministic ledger close is a common local-test requirement. A developer who
needs it today must read the Quickstart shell script. The Quickstart section
already undertakes to document the container's run options, so this is an
omission on an existing page family.

The related count behavior is a separate upstream request. Stellar Core issue
4040, "Support manual closing N ledgers", is open. A manual close advances one
ledger per call today. Documentation should state that limit rather than imply
a count parameter exists.

## Evidence

The code search, the live docs search, and the source reads all ran on
2026-08-14.

The `stellar/quickstart` `start` script carries these three lines:

```sh
: "${ENABLE_CORE_MANUAL_CLOSE:=false}"
    --enable-core-manual-close)
  run_silent "finalize-core-config-manual-close" perl -pi -e "s/__MANUAL_CLOSE__/$ENABLE_CORE_MANUAL_CLOSE/g" etc/stellar-core.cfg
```

Additional recorded evidence:

- 2026-08-14 code search over stellar/stellar-docs for "enable-core-manual-close" returned zero results
- 2026-08-14 read of stellar/quickstart start script shows the flag ships today; line 47 sets ENABLE_CORE_MANUAL_CLOSE=false, line 202 parses --enable-core-manual-close, and line 520 writes it into etc/stellar-core.cfg
- 2026-08-14 live stellarDocs.search_docs for "enable-core-manual-close quickstart manual close ledger" returned eight hits, all about LedgerCloseMeta ingestion, and none about the flag
- 2026-08-14 read of the stellar/quickstart README found no manual-close documentation either
- eval round 2026-08-14, main stamp 2026-08-14T03-56-23-variantA, row q-quickstart-manual-ledger-close

## Recommendation

Add `--enable-core-manual-close` to the Quickstart run options on
`/docs/tools/quickstart/advanced-usage/`. Give the flag name, the default value
`false`, and the Stellar Core setting it controls.

Add one run-command example that starts a local network with the flag enabled.

State that one manual close advances one ledger, and link Stellar Core issue
4040 for the pending count request.

## Source Record

Raven recorded this finding as sd-044 (stellar-docs, discovered 2026-08-14).

Public source record: [improvements/stellar-docs/sd-044-quickstart-manual-close-flag-undocumented.md](https://github.com/stellar-experimental/stellar-raven/blob/main/improvements/stellar-docs/sd-044-quickstart-manual-close-flag-undocumented.md)

Immutable source snapshot: [dbc466e55105](https://github.com/stellar-experimental/stellar-raven/blob/dbc466e551053d262ea4ee4267a09a6526124396/improvements/stellar-docs/sd-044-quickstart-manual-close-flag-undocumented.md)

## Resolution Handoff

When you deploy a fix, link the resolving issue or pull request to the source record.
Then notify Raven through:

https://github.com/stellar-experimental/stellar-raven/issues/new?template=upstream-improvement-ready.yml&title=%5Bupstream-ready%5D%20sd-044%3A%20

Include the finding ID and the resolving issue or pull request.
Include the deployed version or timestamp. Include the smallest live recheck.
Raven verifies the live surface before it sets the finding to `fixed-upstream`.
An issue closure or merged pull request does not prove the fix.
A separate reviewer repeats the live check before Raven retires the active finding.
Raven keeps a commit-pinned snapshot when one is available.

devtechedge:
Fixes #2772

The Quickstart container supports manual ledger close via `--enable-core-manual-close`, but no Quickstart page named the flag. This adds it to the advanced usage section:

- A "Manual close mode" section on the Operation Modes page covering the flag name, its default value of `false`, the `MANUAL_CLOSE` setting it writes into `etc/stellar-core.cfg`, and how a close is triggered via the `manualclose` command on stellar-core's HTTP endpoint (port `11626`).
- A run-command example on the Run Commands page that starts an ephemeral local network with the flag enabled, in the existing macOS/Linux and Windows tab format.
- A note that each `manualclose` invocation advances exactly one ledger, linking [stellar-core#4040](https://github.com/stellar/stellar-core/issues/4040) where support for closing N ledgers is requested, so the docs do not imply a count parameter exists.

Flag behavior was verified against the `start` script on the current `stellar/quickstart` main: the flag is boolean, defaults to `false` (`ENABLE_CORE_MANUAL_CLOSE`), and replaces `__MANUAL_CLOSE__` in the generated core config.

### DIFF
--- a/docs/tools/quickstart/advanced-usage/operation-modes.mdx
+++ b/docs/tools/quickstart/advanced-usage/operation-modes.mdx
@@ -53,4 +53,4 @@ With manual close enabled, ledgers only advance when you explicitly trigger a cl
 curl "http://localhost:11626/manualclose"
 ```
 
-Each `manualclose` invocation advances the ledger by exactly one. Support for closing multiple ledgers per invocation is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
+With no parameters, each `manualclose` invocation advances the ledger by one. On a standalone network, the command also accepts an optional `ledgerSeq` parameter to close up to a specific ledger sequence, and an optional `closeTime` parameter to set the close time of the next ledger. Support for closing multiple consecutive ledgers with a single `count` parameter is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
--- a/docs/tools/quickstart/advanced-usage/operation-modes.mdx
+++ b/docs/tools/quickstart/advanced-usage/operation-modes.mdx
@@ -45,7 +45,7 @@ Upon launching a persistent mode container for the first time, the launch script
 
 ## Manual close mode
 
-By default, ledgers close automatically every few seconds. For deterministic local testing, you can disable automatic closing by passing the `--enable-core-manual-close` flag when starting the container. The flag sets the `MANUAL_CLOSE` setting in the container's generated `etc/stellar-core.cfg`, and defaults to `false` when not specified.
+By default, ledgers close automatically every few seconds. For deterministic local testing, you can disable automatic closing by passing the `--enable-core-manual-close` flag when starting the container. The flag sets the `MANUAL_CLOSE` setting in the `etc/stellar-core.cfg` file that the container generates when it initializes, and defaults to `false` when not specified. The configuration is only generated on first initialization, so adding the flag to a container that already has one (for example, on a persistent volume) has no effect until the configuration is regenerated.
 
 With manual close enabled, ledgers only advance when you explicitly trigger a close by sending the `manualclose` command to stellar-core's HTTP endpoint on port `11626`:
 
@@ -53,4 +53,4 @@ With manual close enabled, ledgers only advance when you explicitly trigger a cl
 curl "http://localhost:11626/manualclose"
 ```
 
-With no parameters, each `manualclose` invocation advances the ledger by one. On a standalone network, the command also accepts an optional `ledgerSeq` parameter to close up to a specific ledger sequence, and an optional `closeTime` parameter to set the close time of the next ledger. Support for closing multiple consecutive ledgers with a single `count` parameter is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
+With no parameters, each `manualclose` invocation advances the ledger by one. On a standalone network, the command also accepts an optional `ledgerSeq` parameter that selects the sequence number of the ledger to close (intermediate ledgers are skipped), and an optional `closeTime` parameter to set the close time of the next ledger. Support for closing multiple consecutive ledgers with a single `count` parameter is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
--- a/docs/tools/quickstart/advanced-usage/operation-modes.mdx
+++ b/docs/tools/quickstart/advanced-usage/operation-modes.mdx
@@ -42,3 +42,15 @@ Upon launching a persistent mode container for the first time, the launch script
 1. Run an interactive session of the container at first, ensuring that all services start and run correctly.
 2. Shut down the interactive container (using Ctrl-C).
 3. Start a new container using the same host directory in the background.
+
+## Manual close mode
+
+By default, ledgers close automatically every few seconds. For deterministic local testing, you can disable automatic closing by passing the `--enable-core-manual-close` flag when starting the container. The flag sets the `MANUAL_CLOSE` setting in the container's generated `etc/stellar-core.cfg`, and defaults to `false` when not specified.
+
+With manual close enabled, ledgers only advance when you explicitly trigger a close by sending the `manualclose` command to stellar-core's HTTP endpoint on port `11626`:
+
+```sh
+curl "http://localhost:11626/manualclose"
+```
+
+Each `manualclose` invocation advances the ledger by exactly one. Support for closing multiple ledgers per invocation is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
--- a/docs/tools/quickstart/advanced-usage/operation-modes.mdx
+++ b/docs/tools/quickstart/advanced-usage/operation-modes.mdx
@@ -45,7 +45,7 @@ Upon launching a persistent mode container for the first time, the launch script
 
 ## Manual close mode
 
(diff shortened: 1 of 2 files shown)


docs/tools/quickstart/advanced-usage/operation-modes.mdx
```diff
@@ -42,3 +42,15 @@ Upon launching a persistent mode container for the first time, the launch script
 1. Run an interactive session of the container at first, ensuring that all services start and run correctly.
 2. Shut down the interactive container (using Ctrl-C).
 3. Start a new container using the same host directory in the background.
+
+## Manual close mode
+
+By default, ledgers close automatically every few seconds. For deterministic local testing, you can disable automatic closing by passing the `--enable-core-manual-close` flag when starting the container. The flag sets the `MANUAL_CLOSE` setting in the `etc/stellar-core.cfg` file that the container generates when it initializes, and defaults to `false` when not specified. It applies to the `--local` network: stellar-core refuses to start with `MANUAL_CLOSE` enabled unless `NODE_IS_VALIDATOR` is set, and the Quickstart local network configuration is the one that sets it. The configuration is only generated on first initialization, so adding the flag to a container that already has one (for example, on a persistent volume) has no effect until the configuration is regenerated.
```
ElliotFriend: Two corrections here:

**Core doesn't refuse to start.** The `NODE_IS_VALIDATOR` check is in `ApplicationImpl::manualClose`, not startup validation, so a `--testnet --enable-core-manual-close` container starts fine and only fails when you hit the endpoint.

**The flag isn't scoped to `--local`.** `MANUAL_CLOSE=__MANUAL_CLOSE__` is in all four config templates. Local is the only usable one because `local/core/etc/stellar-core.cfg:12` is the only config setting `NODE_IS_VALIDATOR=true`.

Suggestion below also splits the paragraph, moves the persistent-volume caveat into a note, and fixes the cadence (`network-modes.mdx` says local closes every one second, not "every few seconds").

```suggestion
By default, ledgers close automatically. (In local network mode, that's once every second.) For deterministic local testing, you can disable automatic closing by passing the `--enable-core-manual-close` flag when starting the container. The flag sets the `MANUAL_CLOSE` setting in the `etc/stellar-core.cfg` file that the container generates when it initializes, and defaults to `false` when not specified.

Manual close is only usable in local network mode. Stellar Core requires `NODE_IS_VALIDATOR=true` to service a manual close request, and the local network is the only Quickstart configuration that sets it. The flag is accepted on the other networks and the container will start normally, but the `manualclose` request itself will fail.

:::note

The core configuration is only generated the first time a container initializes. Adding this flag to a container that already has a configuration (on a persistent volume, for example) has no effect. To pick up the change, start a fresh ephemeral container or clear the mounted volume.

:::
```

docs/tools/quickstart/advanced-usage/operation-modes.mdx
```diff
@@ -42,3 +42,15 @@ Upon launching a persistent mode container for the first time, the launch script
...
+
+With manual close enabled, ledgers only advance when you explicitly trigger a close by sending the `manualclose` command to stellar-core's HTTP endpoint on port `11626`:
+
+```sh
+curl "http://localhost:11626/manualclose"
+```
+
+With no parameters, each `manualclose` invocation advances the ledger by one. On a standalone network, the command also accepts an optional `ledgerSeq` parameter that selects the sequence number of the ledger to close (intermediate ledgers are skipped), and an optional `closeTime` parameter to set the close time of the next ledger. Support for closing multiple consecutive ledgers with a single `count` parameter is requested in [stellar-core issue 4040](https://github.com/stellar/stellar-core/issues/4040). For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
```
ElliotFriend: These parameters aren't reachable from Quickstart at all. `CommandHandler::manualClose` rejects any parameter unless `RUN_STANDALONE=true`, and `RUN_STANDALONE` appears nowhere in the quickstart repo, so Core's default of `false` always applies. `--enable-core-manual-close` doesn't change that; the two settings are independent.

"On a standalone network" makes it worse, since `network-modes.mdx` tells readers the local passphrase is literally `Standalone Network ; February 2017`.

Also dropping the stellar-core#4040 link, since nothing's moving on it and a stalled issue reads as a promise. The one-ledger limit is the part readers need.

(Your "intermediate ledgers are skipped" wording is correct, just not reachable here.)

```suggestion
Each `manualclose` invocation advances the ledger by one. For a complete example of starting a local network with the flag enabled, see [Run Commands](./run-command-examples.mdx).
```

ElliotFriend (commented):
Thanks for picking this up, and nice work tracking the flag down in the `start` script.

Two factual corrections inline, both in the new manual close section. The short version: the `NODE_IS_VALIDATOR` check fires when you call `manualclose`, not at startup, and the `ledgerSeq`/`closeTime` parameters require `RUN_STANDALONE=true`, which Quickstart never sets on any network.

One of the suggestions also drops the stellar-core#4040 link. Nothing seems to be moving on that issue, and pointing at it reads as a promise. The one-ledger limit still gets stated plainly, which was the substantive part of the ask in #2772.

Everything else looks good. The loopback bind on 11626 is the right call, the tab structure matches the surrounding examples, and there's no existing `manualclose` coverage elsewhere in the repo to reconcile with.

One follow-up for a separate PR, not this one: the `## Local` section in `network-modes.mdx` lists the local-only parameters (`--protocol-version` and `--limits`), and this flag belongs there too with a link across.

devtechedge:
Both corrections applied, thanks for the detail on where the `NODE_IS_VALIDATOR` check actually fires and on `RUN_STANDALONE`.

The manual close section now says the flag is accepted on every network but only usable on local, the persistent-volume caveat is in a note, and the cadence is one second. The unreachable `ledgerSeq` / `closeTime` parameters and the stellar-core#4040 link are gone.

Happy to open the `network-modes.mdx` `## Local` addition as a separate PR.

ElliotFriend (approved):
lgtm! thanks for getting this fixed up!

devtechedge:
Your two corrections were the useful part of this review, @ElliotFriend: the `NODE_IS_VALIDATOR` check fires when `manualclose` is called, not at startup, which is why the flag is accepted on every network but only usable on local.

The same pass caught `ledgerSeq` and `closeTime`, which `manualclose` rejects unless `RUN_STANDALONE=true`, and Quickstart never sets that on any network.

Dropping the stellar-core#4040 link was the right call too, since a stalled issue reads as a promise rather than a limitation.

Thanks for the careful read.

kalepail:
Raven independently verified this fix on September 16, 2026.

The original search returns two Quickstart records with the flag. Operation Modes documents its default, setting, and limits. The rendered Run Commands example contains the flag and Core port mapping. That example page remains unindexed.

A separate reviewer repeated the source and index checks.
Raven will retire `sd-044` after recording the resolution receipt.

[Original finding](https://github.com/stellar-experimental/stellar-raven/blob/50bf5518860584ec1e5d352acbe11033515a0b7f/improvements/stellar-docs/sd-044-quickstart-manual-close-flag-undocumented.md)

PR #2859, opened 2026-09-15
List --enable-core-manual-close with the other local network parameters

devtechedge:
Follows up on #2850, where @ElliotFriend asked for this flag to be listed alongside the other local-only parameters.

The `## Local` section of the Network Modes page lists the parameters that only the local network accepts, `--protocol-version` and `--limits`, and `--enable-core-manual-close` belongs in that list too. The flag is accepted on every network, but only the local configuration sets `NODE_IS_VALIDATOR=true`, which stellar-core requires to service a `manualclose` request.

This adds one bullet with a cross-link to the Manual close mode section, so the parameter list no longer omits a flag the container ships.


ElliotFriend (approved):
great addition! thanks for the help with these!!

devtechedge:
Thanks for merging this one so quickly, @ElliotFriend.

Your note on #2850 that the flag belonged in the `## Local` list was the right call, since `--protocol-version` and `--limits` already set the expectation that the section is complete.

Checking the container config is what pinned the wording down: local is the only network that sets `NODE_IS_VALIDATOR`, which stellar-core needs to service `manualclose`, so the cross-link earns its place there.

Appreciate you getting to both PRs the same day.

kalepail:
Raven independently verified this fix on September 16, 2026.

The original search returns two Quickstart records with the flag. Operation Modes documents its default, setting, and limits. The rendered Run Commands example contains the flag and Core port mapping. That example page remains unindexed.

A separate reviewer repeated the source and index checks.
Raven will retire `sd-044` after recording the resolution receipt.

[Original finding](https://github.com/stellar-experimental/stellar-raven/blob/50bf5518860584ec1e5d352acbe11033515a0b7f/improvements/stellar-docs/sd-044-quickstart-manual-close-flag-undocumented.md)

